PR #24, opened 2026-01-15
New crsp format

miahdy:
Updated all CRSP-related code to use the new CIZ-format tables, including table names, column mappings, and logic where variable definitions changed.

### DIFF
--- a/code/aux_functions.py
+++ b/code/aux_functions.py
@@ -490,7 +490,7 @@ def gen_crsp_sf(freq):
     # ---------- CIZ name mapping by frequency ----------
     if freq == "m":
         date_expr   = sf.mthcaldt.cast("date")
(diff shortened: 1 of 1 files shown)


theisij:
These correlations look surprisingly low to me. I didn't think it would make any noticeable difference. 

Can you run one more test? Specifically, I'd like you to compare the daily returns from the old and new file. Specifically, take the data from processed/return_data/daily_rets_by_country/usa and filter on id<=99999 (will only include PERMNOs). Then show (1) the total number of observations in each during the 1925-2024 period, (2) the fraction of exact matches (i.e., cases where the return in the old and new firm is literally the same), and (3) the fraction of exact matches for firms that delist on the delisting date.

I'm particularly interested in (3), because I want to make sure that we handle delistings correctly. 

Thanks!

PR #57, opened 2026-04-01
[codex] Add ticker to gen_crsp_sf output

Sasha-Cui:
## Summary

- expose `ticker` in `gen_crsp_sf()` output after the CRSP security-info join
- add a focused unit test covering both monthly and daily paths

## Why

Issue #36 asked for a reliable way to carry ticker information through this CRSP helper output. Theis specifically suggested adding the `ticker` column from `crsp_stksecurityinfohist.parquet` and opening a PR once validated.

## Impact

This changes the schema of the generated CRSP security-file output by adding a `ticker` column. Existing columns and join behavior are unchanged.

## Validation

- Bouchet: `source .venv/bin/activate && pytest tests/unit/test_aux_functions.py -q`
- Local syntax check: `python3 -m py_compile code/aux_functions.py tests/unit/test_aux_functions.py`

Closes #36.

PR #61, opened 2026-04-02
Revert "[codex] Add ticker to gen_crsp_sf output"

FernandoRDLL:
Reverts bkelly-lab/jkp-data#57


FernandoRDLL:
I need to finish a few final checks on this

Issue #38, opened 2026-03-11
[Feature]: winsorized Compustat returns in output files


theisij:
### Feature Description

Compustat returns have a few crazy outliers for tiny stocks, and whenever I work with returns outside of the US, I always winsorize the returns at 0.1% and 99.9%, using the data from `data/other_output/return_cutoffs.parquet` (for monthly data) and `data/other_output/return_cutoffs_daily.parquet` (for daily data). 

I suggest we add a new column called 'ret_exc_wins" to the final version of `world_dsf`, 'world_msf' and 'world_data', such that people don't have to do this themselves 

To get some inspiration for how to do this, I urged whoever is implementing this to look at the `code/portfolios.py' scipt. 

For example, the below is the approach we use to winsorize daily returns

```python
# Get data
ret_cutoffs_daily = pl.read_parquet(f"{data_path}/other_output/return_cutoffs_daily.parquet")

# Extracting year and month from the date column
daily = daily.with_columns(
    [
        pl.col("date").dt.year().cast(pl.Int64).alias("year"),
        pl.col("date").dt.month().cast(pl.Int64).alias("month"),
    ]
)

# Joining with daily return cutoffs
daily = daily.join(
    ret_cutoffs_daily.with_columns(
        [pl.col("year").cast(pl.Int64), pl.col("month").cast(pl.Int64)]
    )
    .select(["year", "month", "ret_exc_0_1", "ret_exc_99_9"])
    .rename({"ret_exc_0_1": "p001", "ret_exc_99_9": "p999"}),
    on=["year", "month"],
    how="left",
)

# Applying winsorization to daily returns for Compustat data (id > 99999)
daily = daily.with_columns(
    pl.when((pl.col("id") > 99999) & (pl.col("ret_exc") > pl.col("p999")))
    .then(pl.col("p999"))
    .when((pl.col("id") > 99999) & (pl.col("ret_exc") < pl.col("p001")))
    .then(pl.col("p001"))
    .otherwise(pl.col("ret_exc"))
    .alias("ret_exc")
).drop(["p001", "p999", "year", "month"])
```

### Use Case / Motivation

**NOTE:** This may be unnecessary if we get the Bessembinder et al algorithm implemented from issue #19.  

### Willingness to Contribute

Yes, I can implement this myself and submit a PR

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this feature has not already been requested
- [x] I have read the documentation to confirm this feature does not already exist

PR #63, opened 2026-04-03
Issue 38

FernandoRDLL:
## Summary
Solution to #38 
Add ret_exc_wins column (winsorized excess return) to world_dsf, world_msf, and world_data. Compustat returns are clipped at the 0.1%/99.9% percentiles per month; CRSP returns are left unchanged. Also documents the new column in the version history table and refactors the table so it breaks cleanly across pages.

## Change Type

- [X] Data/output change (affects computed outputs users download)
- [ ] Documentation only

## Methodological Impact
None. This adds a convenience column alongside existing return columns. No existing factor definitions or calculations are modified.

## Data Impact

  Five output files gain a new ret_exc_wins column:
  1. return_data/world_dsf.parquet
  2. return_data/world_ret_monthly.parquet
  3. return_data/daily_rets_by_country/
  4. characteristics/world_data_unfiltered.parquet
  5. characteristics/world_data_filtered.parquet
  All existing columns are unchanged. This is purely additive.

## Breaking Changes

Same as previous section. I have updated the pdf release notes.

## Tests

  New test file tests/unit/test_ret_exc_wins.py covers the winsorization logic, including:
  - CRSP returns are passed through unchanged
  - Compustat returns are clipped at the per-month percentile bounds
  - Null handling

## Checklist

- [X] I have run the tests locally (`uv run pytest`)
- [X] I have run the linter locally (`uv run ruff check code/ tests/`)
- [X] I have verified the pipeline runs successfully after my changes
- [X] I have updated documentation if needed
- [X] I have considered whether this change affects factor calculations

## Additional Notes

  The version history table in documentation.tex was refactored from \begin{itemize} blocks inside cells to one-row-per-bullet with \textbullet prefixes. This fixes a long-standing issue where the table couldn't break across pages because LaTeX cannot split a row mid-cell.


FernandoRDLL:

theisij:
## Review Feedback

Hi Fernando,

Thanks for this PR — the overall approach and test coverage look good. A few items to address before we can merge:

### Required Changes

- [ ] **[code/aux_functions.py:3263–3280] Use precomputed cutoffs instead of recomputing percentiles.** The current implementation computes `QUANTILE_DISC` over all stocks per `eom`, but the percentile population should come from filtered stocks only (`common=1, obs_main=1, exch_main=1, primary_sec=1, excntry!='ZWE'`) — matching what `return_cutoffs()` already produces. Issue #38 explicitly says to use `return_cutoffs.parquet` / `return_cutoffs_daily.parquet`. Rewrite to join against the precomputed cutoffs and clip, following the pattern in `add_cutoffs_and_winsorize()` / `portfolio.py`. This also removes the need for SQL window functions and will be faster on daily data.

- [ ] **[code/aux_functions.py:3243] Add input validation for `lower`/`upper`.** The percentile parameters are interpolated into SQL without any guard. Add a check like:
  ```python
  if not (0 <= lower < upper <= 1):
      raise ValueError(f"Percentile bounds must satisfy 0 <= lower < upper <= 1, got {lower=}, {upper=}")
  ```

- [ ] **[documentation/documentation.tex:~128–136] Replace `enumerate` with `\textbullet` rows.** The new 04-03-2026 changelog entry uses an `enumerate` environment inside a table cell, which can still cause page-break issues. Use the same multi-row `\textbullet~` pattern you applied to all the older entries.

- [ ] **[documentation/documentation.tex:~1996, ~2044] Add Market Characteristics table entries.** `ret_exc_wins` should be documented in the Market Characteristics table (both CRSP and Compustat sections), consistent with how `ret_exc`, `ret_local`, and `ret_exc_lead1m` are documented there.

- [ ] **[code/aux_functions.py] Consider using `source_crsp` instead of `id > 99999`.** The canonical pattern in `aux_functions.py` for distinguishing CRSP vs Compustat is `source_crsp = 0`, not `id > 99999`. Both are functionally equivalent, but `source_crsp` is semantically clearer. If `source_crsp` is not available in the DataFrame at this pipeline stage, `id > 99999` is fine.

- [ ] **[tests/unit/test_ret_exc_wins.py:25, 30, 35] Add type annotation for `path` parameter.** The helper functions `_make_world_msf`, `_make_world_dsf`, and `_read_result` are missing a type annotation on `path` (should be `path: Path`).

FernandoRDLL:
@theisij 
Ready for review. I applied the changes

theisij (approved):
Looks good — approving and merging.

Issue #64, opened 2026-04-03
[Other]: Simplify winsorization logic


FernandoRDLL:
### Category

Methodology / academic question

### Description

The winsorization logic can be simplified for cleaner code.
Daily return cutoffs can be derived using the eom variable for grouping rather than year and month. Most files already contain it.

### Pre-submission Checklist

- [x] I have searched existing issues and discussions for an answer
- [x] I have read the documentation

PR #65, opened 2026-04-03
Simplify winsorization

FernandoRDLL:
## Summary
This addresses Issue #64 

## Change Type

- [X] Data/output change (affects computed outputs users download)
- [ ] Documentation only
- [X] Performance-impacting change

## Methodological Impact

None

## Data Impact

`return_cutoffs_daily.parquet` gains a new `eom` column (end-of-month date used as the internal grouping key). The existing `year` and `month` columns are preserved for backward compatibility.

## Breaking Changes

None — the output schema change is additive only.

## Tests

I added tests for the winsorization logic, including a test for missing cutoff rows verifying that GREATEST/LEAST in Polars SQL correctly preserves original return values when cutoff bounds are NULL.

## Checklist

- [X] I have run the tests locally (`uv run pytest`)
- [X] I have run the linter locally (`uv run ruff check code/ tests/`)
- [X] I have verified the pipeline runs successfully after my changes
- [X] I have updated documentation if needed
- [X] I have considered whether this change affects factor calculations

## Additional Notes

This provides a cleaner alternative to the past code. Makes the code simpler to read and aligns the winsorization method/syntax used across all the data. It's important to have the same syntax as in #38


FernandoRDLL:

theisij:
## Review Feedback

Hi Fernando,

Good PR — the GREATEST/LEAST simplification is clean and the test coverage is thorough. A few items to address:

### Required Changes

- [ ] **[Output schema] Keep `year` and `month` columns in `return_cutoffs_daily.parquet`.** I'd prefer not to introduce breaking changes to the output file schema unless there's a strong substantive reason. Switching from `(year, month)` to `(eom)` is a convenience simplification — the grouping is mathematically equivalent — but it breaks downstream consumers who rely on the current columns. The internal code can use `eom` for joins, but the output file should preserve the existing schema. If you feel there's a compelling reason to change it, let me know and we can discuss.

- [ ] **[PR description] Update "Data Impact" section.** Currently says "None," but if any output schema does change, it needs to be documented here.

- [ ] **[documentation/documentation.tex] Update if output schema changes.** The Daily Return Cutoffs section (lines 585–618) describes the current `year`/`month` columns. Keep documentation in sync with whatever the final output schema is.

- [ ] **[tests] Add a test for `add_cutoffs_and_winsorize` when cutoff rows are missing.** The GREATEST/LEAST refactoring relies on Polars SQL skipping NULLs in GREATEST/LEAST (unlike standard SQL). This works today but is engine-dependent. Add a test where the cutoffs join produces no match for a given `eom`, verifying the original return value is preserved, not NULLed.

- [ ] **[code/aux_functions.py] `drop_non_trading_days`** — bare `col()` in new line. Should be `pl.col()` per project convention (C5).

- [ ] **[code/aux_functions.py] Add type annotations to new `lower`/`upper` parameters** in `ap_factors` and `prep_data_factor_regs` (e.g., `lower: float = 0.001`).

- [ ] **Scope: `lower`/`upper` parameterization.** Issue #64 asks for simplification only. The configurable percentile bounds are a nice addition but expand the API beyond what was requested. Consider splitting into a separate PR or noting it explicitly as a bonus in the description.

- [ ] **Merge ordering with PR #63.** Both PRs touch overlapping code and you mention wanting "the same syntax as in #38." Which should merge first? Document the intended order to avoid conflicts.

- [ ] **Inline comment in `ap_factors` explaining why it uses CTE+JOIN** while `prep_data_factor_regs` uses window functions (Polars SQL vs DuckDB difference).

Let me know if you have questions — especially if there's a reason for the output schema change that I'm missing.

FernandoRDLL:
@theisij ready to go

theisij (approved):
Looks good — approving and merging.

PR #66, opened 2026-04-04
Add CSV output format support with quoted strings

capellini:
## Summary

Add optional CSV output format to preserve leading zeros in identifier columns (`gvkey`, `iid`) when exporting data for WRDS/SAS compatibility. The pipeline runs with parquet internally and converts to CSV at the end if `--output-format csv` is specified.

## Change Type

- [ ] Documentation only

## Methodological Impact

None

## Data Impact

Users can now choose CSV output format via `--output-format csv` flag in portfolio.py. CSV files use `quote_style="non_numeric"` to preserve leading zeros in string identifiers (gvkey, iid). Default remains parquet format - no change to existing behavior.

## Breaking Changes

None

## Tests

- Added `tests/unit/test_output_writer.py` with 25 tests covering:
  - Output format configuration (immutable "set once" pattern)
  - DataFrame writing with extension correction
  - CSV conversion with quoted strings and null handling
  - Streaming and shrink_dtype options
- Added `tests/unit/test_portfolio.py` with 4 tests for `parse_args()`

## Checklist

## Additional Notes

- Full pipeline tested on Yale HPC with CSV output - all 376 files converted successfully
- Validated CSV output matches parquet data exactly (row counts, column counts, numeric values)
- String quoting verified: gvkey `"004503"`, iid `"01"` preserve leading zeros
- Row ordering in output files is non-deterministic due to parallel processing (same as parquet) - this is existing behavior, not introduced by this PR


capellini:
@theisij After experimenting with a few conversion strategies, I decided the simplest way to do this was to do it at the very end with all interim processing still using Parquet files. If you'd like me to pursue a different strategy, please let me know.

I'm going to keep this in draft until I run this one more time on the Yale HPC to ensure the full pipeline runs without any problems.

capellini:
@theisij I ran the entire pipeline using both the parquet output and the CSV output and the files match. There was non-deterministic ordering between the parquet and CSV files, but that's the nature of our current algorithm (if we want to add determinism, we probably want to address it in a separate PR, as it's not really related to this PR -- which I can do).

The changes to portfolio.py are less than they actually seem. I'm really just wrapping things in a proper `__name__ == '__main__'` block and adding a `parse_args` function to parse arguments.

theisij:
## Review Feedback

Hi Rob,

Thanks for this PR — the CSV output feature is well-motivated and the `portfolio.py` refactoring into `main()` is a nice improvement.

Both copilot and Claude code flagged an issue with the current approach, where the parquet files are deleted once the .CSV version is executed. This may be your intented behavior, in which case it's fine, but I just wanted to make 100% sure that the stuff copilot and Claude Code flags is not a real concern with the modifications. If everything is as intended, then feel free to disregard, and I'll approve and merge the PR. Below is the changes suggested by copilot and Claude code. Please free to disregard if they are missing something about the original intent.

A few items to address before we can merge:

### Required Changes

- [ ] **[output_writer.py:213]** `convert_outputs_to_csv()` must not delete the original parquet files after conversion. Remove `parquet_file.unlink()` — CSV files should be written alongside the parquet files, not replace them.

- [ ] **[portfolio.py:1402]** Scope `convert_outputs_to_csv()` to only convert the files that `portfolio.py` actually writes (e.g., `data/processed/portfolios/`), not all of `data/processed/`. The current default recursively walks the entire tree including `characteristics/`, `return_data/`, `accounting_data/`, and `other_output/` — files written by `main.py` that shouldn't be touched by `portfolio.py`.

- [ ] **[portfolio.py:939]** Bug: the `pf_daily` concat comprehension filters with `"pf_returns" in sub_data` but then indexes `sub_data["pf_daily"]`. This should gate on `"pf_daily" in sub_data` instead. Can silently drop valid daily results or raise a KeyError.

- [ ] **[output_writer.py:213]** Add error handling to `convert_single_file()` so that a failed CSV conversion for one file doesn't leave the output directory in an inconsistent state. A try/except per file that collects and surfaces failures at the end would be sufficient.

- [ ] **[README.md:64]** Update the CSV output documentation to accurately describe the behavior once the above changes are made (CSVs written alongside parquet, scoped to portfolio outputs).

- [ ] **[portfolio.py:31]** Argparse epilog examples use `portfolio.py` — should be `code/portfolio.py` to match repo documentation and typical usage from the repo root.

- [ ] **[portfolio.py]** `parse_args()` and `main()` use single-line docstrings — project convention is the three-part `Description/Steps/Output` format (see CLAUDE.md).

capellini:
> Both copilot and Claude code flagged an issue with the current approach, where the parquet files are deleted once the .CSV version is executed. This may be your intented behavior, in which case it's fine, but I just wanted to make 100% sure that the stuff copilot and Claude Code flags is not a real concern with the modifications

I don't have a strong opinion, but I tend to disagree with copilot and Claude. If I specify output as CSV, I don't want any other artifacts/leftover files -- I just want the CSV files -- especially considering the sheer volume of data.

But you and @FernandoRDLL are much more familiar with this code base than I am and may forsee a use case where preserving these files are necessary. Barring that, I think that removing the parquet files is acceptable.

There are a few other comments that I'd like to review and I have to rebase this over the current main branch. In the meantime, if you and Fernando can let me know what you think about the necessity of keeping the parquet files, I would appreciate your advice.

capellini:
- Rebased main
- Fixed bug in the pf_daily concat comprehension
- Updated argparse epilog examples
- Changed docstrings to use Description/Steps/Output format

The other code review comments seem to not apply, since we don't want to keep parquet files around, if the user specifies CSV output.

theisij (approved):
Okay, good point. Approving and merging

PR #68, opened 2026-04-05
Apply default screening filters before save section (#40)

### LINKED ISSUE #40: [Feature]: Add main filters by default
## Background
Our guiding principle when create the JKP data was to return as much data as possible, and then have users filter the data in a way that suited our needs. 

In our own work, we almost always incorporated the suggested screens described below:

Image
 
However, it has become apparent that the vast majority of user want to employ exactly the same screens as us. And further, by not having the screens implemented as a default, we have cases where users were surprised to find that the data didn't match their expectation.

Therefore, I suggest we incorporate the suggested screens as a default in all code that we output.

## How should it be implemented
To have as little effect on the end output as possible, I suggest we add three functions `filter_dsf()`, `filter_msf`, and `filter_world` right before the saving section of `code/main.py`. That is, right before `save_main_data`, so something like 
`
filter_dsf()
filter_msf()
filter_world()
save_main_data(end_date)
save_daily_ret()
save_monthly_ret()
save_accounting_data()
save_output_files()
save_full_files_and_cleanup(clear_interim=True)
`
I don't think we need to filter the accounting data for now. 

### Use Case / Motivation

See above

_No response_

_No response_

### Willingness to Contribute

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

FernandoRDLL:
## Summary

Apply the standard screening filters (`obs_main=1`, `common=1`, `exch_main=1`, `primary_sec=1`) by default before the save section. Filter functions write to `*_output.parquet` files, preserving unfiltered source data in `interim/`. Removes the `world_data_unfiltered`/`world_data_filtered` split in favor of a single `world_data.parquet` output.

Closes #40

## Change Type

- [ ] Documentation only

## Methodological Impact

None — the same four screens were already recommended and used by nearly all users. This change makes them the default rather than optional.

## Data Impact

- Output `characteristics/` directory now contains a single `world_data.parquet` (equivalent to the previous `world_data_filtered.parquet`). The `world_data_unfiltered.parquet` file is no longer produced.
- Daily and monthly return files (`world_dsf.parquet`, `world_ret_monthly.parquet`) are now filtered by the same screens.
- Filter parameters are centralized in `config.py` (`MAIN_FILTERS` dict).

## Breaking Changes

- `world_data_unfiltered.parquet` and `world_data_filtered.parquet` are replaced by a single `world_data.parquet`. Users referencing the old filenames will need to update.
- Release notes have been updated to reflect this change.

## Tests

- Added `TestFilterFunctions` class with 8 tests: filtering correctness for all three functions, column preservation, source preservation, idempotency, all-pass and none-pass edge cases
- Added `MAIN_FILTERS` config tests (type, keys, values)
- Updated `TestSaveMainData` tests to reflect that filtering is now upstream
- All 137 unit tests pass

## Checklist

## Additional Notes

- Filter+save sequence is idempotent: re-running always reads from untouched source files in `interim/`
- `save_output_files()` changed from `mv` to `cp` to preserve interim files
- Documentation updated in LaTeX, Quarto, and HTML

This PR addresses #40 
@theisij please review and merge.


FernandoRDLL:
Re: comments 1-3 (existence checks for *_output.parquet in save_daily_ret, save_monthly_ret, save_full_files_and_cleanup):

  These are not applicable. This PR moves filtering upstream into dedicated filter_dsf(), filter_msf(), and filter_world()
  functions that produce the *_output.parquet files. These functions are called sequentially from main.py before the save
  functions — the *_output.parquet files are guaranteed to exist by the time the save functions run. Adding
  fallback-to-unfiltered logic would silently skip filtering when the caller forgets the filter step, which is the opposite of
  what we want: the whole point of this PR is to make filtering the default. A missing file should fail loudly (Polars already
  raises a clear FileNotFoundError) so the developer knows the pipeline is being run out of order.
  
  @theisij Ready to go

theisij (approved):
This is great. Thanks!

PR #71, opened 2026-04-07
Fix aug_msf_v2 idempotency bug (#56)

### LINKED ISSUE #56: [Bug]: Aug_msf_v2 fails on re-run without fresh WRDS download

main.py (data generation)

### Error Category

Other

### Bug Description

  aug_msf_v2() is not idempotent. It crashes with a duplicate column error if crsp_msf_v2.parquet already contains
  the mthaskhi and mthbidlo columns it tries to add.

### Steps to Reproduce

  1. Run the full pipeline (including download_raw_data_tables()) — succeeds
  2. Comment out download_raw_data_tables() and re-run — aug_msf_v2() fails:
        ibis.common.exceptions.IbisInputError: Duplicate column name 'mthaskhi' in result set

### Expected Behavior

Run twice without having to download everything

### Error Output / Stack Trace

```shell
Traceback (most recent call last):
  File "/gpfs/home/ffr7/jkp-data/code/main.py", line 50, in <module>
    aug_msf_v2()
  File "/gpfs/home/ffr7/jkp-data/code/aux_functions.py", line 774, in aug_msf_v2
    ).select([msf] + [m.mthaskhi, m.mthbidlo])
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfs/home/ffr7/jkp-data/.venv/lib/python3.11/site-packages/ibis/expr/types/joins.py", line 385, in select
    values = unwrap_aliases(values)
             ^^^^^^^^^^^^^^^^^^^^^^
  File "/gpfs/home/ffr7/jkp-data/.venv/lib/python3.11/site-packages/ibis/expr/types/relations.py", line 509, in
un>
    raise com.IbisInputError(
ibis.common.exceptions.IbisInputError: Duplicate column name 'mthaskhi' in result set
```

### Operating System

Linux

### Python Version

3.13.11

### Polars Version

Default uv 

### uv Version

0.10.9

### Available RAM

_No response_

### WRDS Authentication Method

Stored credentials (keyring)

### WRDS Two-Factor Authentication

- [x] Yes, I approved the 2FA request
- [ ] No 2FA prompt appeared
- [ ] 2FA failed or timed out
- [ ] Not applicable

### Additional Context

_No response_

### Pre-submission Checklist

FernandoRDLL:
## Summary

Fixes #56. `aug_msf_v2()` was overwriting `../raw/raw_tables/crsp_msf_v2.parquet` in place, which caused a duplicate-column crash (`Duplicate column name 'mthaskhi'`) on any re-run that skipped a fresh WRDS download. The function now writes its output to a separate file under `raw_data_dfs/`, leaving the raw table untouched.

Changes:
- `aug_msf_v2()` now writes to `raw_data_dfs/crsp_msf_v2_aug.parquet` instead of overwriting `../raw/raw_tables/crsp_msf_v2.parquet`.
- Moved the `aug_msf_v2()` call out of `main.py` and into `gen_raw_data_dfs()` so the augmentation lives alongside the other `raw_data_dfs/` generation steps.
- Updated downstream readers (`gen_crsp_sf("m")` and `firm_age()`) to read from the new augmented path.
- Updated the test fixture in `tests/unit/test_aux_functions.py` to write the monthly fixture to the new location.

## Change Type

- [ ] Documentation only

## Methodological Impact

None. The augmentation logic itself is unchanged — only the output destination and call site move.

## Data Impact

None. The same `mthaskhi`/`mthbidlo` columns are produced from the same daily inputs and joined onto the same monthly rows; only the file location changes.

## Breaking Changes

None for end users. Internally, `../raw/raw_tables/crsp_msf_v2.parquet` is no longer mutated by the pipeline, and the augmented monthly file now lives at `raw_data_dfs/crsp_msf_v2_aug.parquet`. Anyone with custom code reading the augmented columns from the old raw path will need to point at the new file.

## Tests

Updated the existing `_write_sf_fixture` helper in `tests/unit/test_aux_functions.py` to place the monthly fixture at `raw_data_dfs/crsp_msf_v2_aug.parquet` so the existing `gen_crsp_sf("m")` tests continue to exercise the new read path. No new test was added specifically for the idempotency fix because the bug was a side effect of in-place overwrites — eliminating the overwrite removes the failure mode entirely.

## Checklist

## Additional Notes

None.


theisij:
## Review Feedback

Hi Fernando,

Thanks for this PR — clean fix for the idempotency bug. Writing to a separate file instead of overwriting the raw table is the right approach. Two items to address:

### Required Changes

- [ ] **[aux_functions.py]** Add a unit test for `aug_msf_v2()` that verifies (a) the output file is created at `raw_data_dfs/crsp_msf_v2_aug.parquet` with the expected schema (`mthaskhi`, `mthbidlo` columns present), and (b) running it a second time does not raise an error. The structural fix is solid, but a direct regression test documents the idempotency contract and guards against future changes.

- [ ] **[aux_functions.py:6731]** `firm_age()` now reads from `raw_data_dfs/crsp_msf_v2_aug.parquet`, but it only uses `mthcaldt` — an original column that exists in the raw file `../raw/raw_tables/crsp_msf_v2.parquet`. This introduces an ordering dependency on `gen_raw_data_dfs()` where none existed before. Consider reading from the raw file instead, since `firm_age()` doesn't need the augmented columns.

FernandoRDLL:
@theisij Please review, I made the changes.
I don't think we need to make the change for the additional suggestion.

The suggestion is to revert firm_age() back to reading ../raw/raw_tables/crsp_msf_v2.parquet since it only needs mthcaldt. The reasoning given (no ordering dependency, no need for the augmented columns) is technically correct, but I think keeping firm_age() pointed at the augmented file is the better call:

  1. Single source of truth. After this PR, crsp_msf_v2_aug.parquet is the canonical monthly CRSP table the rest of the pipeline reads from — gen_crsp_sf("m") already reads from it (aux_functions.py:493). Having firm_age() reach back to the raw file means two different functions consume "monthly CRSP" from two different paths. That's the kind of inconsistency that causes confusion later when someone wonders which file is authoritative.
  2. The "ordering dependency" already exists. firm_age() is called from main.py well after gen_raw_data_dfs() runs, and gen_raw_data_dfs() is what produces crsp_msf_v2_aug.parquet. So in practice the ordering is already enforced by main.py. Reverting firm_age() to the raw file wouldn't actually decouple anything — it would just create a second, looser contract that's easy to violate accidentally.
  3. The augmented file is a strict superset. It contains every column the raw file has, plus two more. Reading mthcaldt from it produces identical results. There's no correctness or performance cost — mthaskhi/mthbidlo are projected away by the group_by("permco").agg(crsp_first=_.mthcaldt.min()) anyway.
  4. Future-proofing. If someone later adds another augmentation step to aug_msf_v2() (e.g., a corrected mthcaldt for a CRSP data fix), firm_age() would silently use stale dates if it's still reading the raw file. Pointing at the augmented file means firm_age() automatically benefits from any future fixes layered into aug_msf_v2().

theisij:
@FernandoRDLL can you revolve the merging conflicts? Then I'll approve and merge

theisij (approved):

Issue #59, opened 2026-04-01
[Bug]: combine_crsp_comp_sf() runs out of memory


FernandoRDLL:
### Which script were you running?

main.py (data generation)

### Error Category

Memory / OOM error

### Bug Description

This function has OOM events.

### Steps to Reproduce

Run the function in main alone. It runs out of memory. Not when it's done in the full main.

### Expected Behavior

Finish the function and write outputs

### Error Output / Stack Trace

```shell
[I will assign this issue to myself]
```

### Operating System

Linux in SOM HPC

### Python Version

3.13.11

### Polars Version

Default version in uv

### uv Version

0.10.9

### WRDS Authentication Method

Stored credentials (keyring)

### WRDS Two-Factor Authentication

- [ ] Yes, I approved the 2FA request
- [ ] No 2FA prompt appeared
- [ ] 2FA failed or timed out
- [ ] Not applicable

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this bug has not already been reported
- [x] I am using the latest version of the code from the main branch
- [x] I have read the README and followed the setup instructions

Issue #69, opened 2026-04-05
[Bug]: Non-deterministic deduplication in Compustat stock file pipeline


FernandoRDLL:
### Which script were you running?

main.py (data generation)

### Error Category

Computation error

### Bug Description

Three functions in the Compustat stock file pipeline contain non-deterministic deduplication logic inherited from the original SAS code. When duplicate rows exist for the same key, the row that survives is arbitrary, meaning results can differ across runs, machines, or library versions.

**Affected locations:**

1. **`gen_comp_msf()`** (`aux_functions.py` line 1284): `.distinct(on=["gvkey", "iid", "eom"])` does not specify `keep=`, so DuckDB/Ibis picks an arbitrary row among duplicates. Duplicates can occur when both SECD rows exist for the same `{gvkey, iid, eom}` (the `n==2` filter only prefers source=1 over source=2, but doesn't handle same-source duplicates).

2. **`add_primary_sec()`** (`aux_functions.py` lines 1416–1447): The three range-joins on `prihistrow`, `prihistusa`, and `prihistcan` can multiply rows when date ranges overlap for the same `gvkey`. The `SELECT DISTINCT` on line 1417 collapses exact-duplicate rows but cannot resolve cases where overlapping ranges produce genuinely different `prihist*` values for the same `{gvkey, iid, datadate}`.

3. **`gen_returns_df()`** (`aux_functions.py` line 1512): `.unique(["gvkey", "iid", "datadate"])` does not specify `keep=`, so Polars picks an arbitrary row among duplicates.

### Steps to Reproduce

1. Run `uv run python code/main.py`
2. Observe that `gen_comp_msf()` calls `.distinct(on=["gvkey", "iid", "eom"])` without a deterministic tie-breaker
3. Observe that `add_primary_sec()` range-joins can fan out rows when prihist date ranges overlap, and `SELECT DISTINCT` does not guarantee a deterministic winner
4. Observe that `gen_returns_df()` calls `.unique(["gvkey", "iid", "datadate"])` without `keep=`

### Expected Behavior

Deduplication should be deterministic — the same input data should always produce the same output, regardless of execution environment. A tie-breaking rule (e.g., sort by `datadate` and specify `keep="last"`) should be applied at each dedup step.

### Error Output / Stack Trace

N/A — no crash. The issue is silent non-determinism in output data.

### Operating System

N/A (affects all platforms)

### Python Version

N/A (affects all versions)

### Polars Version

N/A (affects all versions)

[...]

PR #72, opened 2026-04-07
Rewrite combine_crsp_comp_sf in DuckDB to fix OOM (#59)

FernandoRDLL:
## Summary

- Rewrites `combine_crsp_comp_sf()` in DuckDB against a persistent `.ddb` file so the merge runs out-of-core instead of materializing the full CRSP+Compustat security files in Python memory. Removes the four Polars helpers (`prepare_crsp_sfs_for_merging`, `prepare_comp_sfs_for_merging`, `gen_temp_sf`, and the per-frequency writer) in favor of a single SQL implementation.
- Replaces the previous non-deterministic `.unique()` dedup with `ROW_NUMBER() OVER (PARTITION BY id, eom ORDER BY source_crsp DESC, primary_sec DESC)`, which deterministically prefers CRSP rows and, on ties within Compustat, the primary security.
- Adds `tests/unit/test_combine_crsp_comp_sf.py` (49 tests across 7 classes) covering each sub-operation in isolation and comparing the DuckDB output against a Polars reference on ~120K monthly + ~1M daily synthetic rows.

Addresses #59.

## Change Type

- [ ] Documentation only
- [x] Performance-impacting change

## Methodological Impact

None. Factor definitions and characteristic calculations are unchanged. The function still produces `__msf_world.parquet` and `world_dsf.parquet` with the same schema and the same `obs_main` rule (prefer CRSP when multiple observations exist for a `(gvkey, iid, eom)`).

## Data Impact

I re-ran the full pipeline on the new branch and the outputs change only minimally relative to the previous version. The non-deterministic dedup that this PR fixes was only reachable on the small subset of `(id, eom)` partitions where more than one row survived the union, so very few observations are actually affected.

How the previous dedup was non-deterministic: the old code did `sf_df.join(obs_main, ...).unique(sort_vars)` where `sort_vars` was `[\"id\", \"eom\"]` (monthly) or `[\"id\", \"date\"]` (daily). Polars' `.unique(subset=...)` keeps an arbitrary representative row from each duplicate group — without an explicit ordering and `keep=` argument, which row survives depends on the underlying execution order (streaming partition boundaries, thread interleaving, input row order). So when a `(id, eom)` partition had multiple candidate rows — e.g. several Compustat rows that disagreed on `primary_sec` — the row that won could change from run to run, and downstream characteristics for that handful of observations could flip with it.

The new SQL pins this down explicitly: `ROW_NUMBER() OVER (PARTITION BY id, eom ORDER BY source_crsp DESC, primary_sec DESC)` and then `WHERE _rn = 1`. `source_crsp DESC` is a no-op in practice today because CRSP and Compustat use disjoint id spaces (raw permno vs. `'1'/'2'/'3' || gvkey || iid[0:2]`), so a partition is always entirely CRSP or entirely Compustat — it is kept to document the intended preference and to stay correct if the id schemes are ever unified. The tie-break that does the real work is `primary_sec DESC`: when multiple Compustat rows share the same `(id, eom)`, the one flagged as the primary security wins, which is both the principled answer and stable across engines and row orderings.

## Breaking Changes

None intentional. The output schema and column semantics are unchanged; only the specific row chosen on a small number of dedup ties may differ from prior runs, and in the deterministic direction.

## Tests

- New file: `tests/unit/test_combine_crsp_comp_sf.py` — 49 tests in 7 classes that exercise each step of the new SQL pipeline (CRSP/Compustat normalization, monthly union, `ret_exc_lead1m` lead, `obs_main` derivation, monthly dedup, daily dedup, end-to-end). Each step is also cross-checked against a Polars reference implementation on ~120K monthly + ~1M daily synthetic rows.
- All existing unit tests still pass under `uv run pytest tests/unit/`.

## Checklist

## Additional Notes

The `aux_combine_sf.ddb` file is created next to the parquet outputs and removed at the end of the function; if a previous run crashed it will be overwritten on the next call. Memory ceiling on the SOM HPC node was the original motivation (#59) — DuckDB now spills to disk through the `.ddb` file rather than holding the unioned monthly/daily frames in RAM.


FernandoRDLL:
This is also like 5x faster.

theisij:
@FernandoRDLL: can you show an example of duplicates that we want to remove? (as per our conversation)

theisij:
## Review Feedback

Hi Fernando,

Thanks for this PR — the DuckDB rewrite is well-structured, the dedup fix is a real correctness improvement, and the 51-test suite is thorough. A few items to address before we can merge:

### Required Changes

- [ ] **[aux_functions.py:1978–1992]** The `obs_main` table has one row per `sf_world_m` record, so `(id, eom)` is not unique. When LEFT JOINed back to `sf_world_m`, this creates a many-to-many intermediate result that `ROW_NUMBER()` then collapses. The final output is correct, but the transient row inflation works against the OOM fix. Consider deduplicating obs_main before the join:
  ```sql
  CREATE TABLE obs_main AS
  SELECT id, eom, MAX(obs_main) AS obs_main
  FROM (
      SELECT id, eom,
          CAST(CASE
              WHEN cnt IN (0, 1) THEN 1
              WHEN cnt > 1 AND source_crsp = 1 THEN 1
              ELSE 0
          END AS INT) AS obs_main
      FROM (
          SELECT id, source_crsp, eom,
              COUNT(gvkey) OVER (PARTITION BY gvkey, iid, eom) AS cnt
          FROM sf_world_m
      ) sub
  ) dedup
  GROUP BY id, eom
  ```

- [ ] **[aux_functions.py:1886–2098]** The DuckDB connection and `.ddb` file cleanup aren't guarded against exceptions. If any SQL step raises, `con.close()` and the `rm` are skipped, potentially leaving a locked `.ddb` file that breaks the next run. Wrap the body in `try/finally`:
  ```python
  con = duckdb.connect("aux_combine_sf.ddb")
  try:
      # ... all SQL steps ...
  finally:
      con.close()
      Path("aux_combine_sf.ddb").unlink(missing_ok=True)
  ```

- [ ] **[test_combine_crsp_comp_sf.py:1245]** `test_crsp_dsf_normalization` asserts `df["id"] == df.select(pl.col("id")).to_series()` which compares a series to itself — it can never fail. Compare against the raw `permno` from the source parquet instead:
  ```python
  raw = pl.read_parquet(toy_dir / "crsp_dsf.parquet")
  assert (df["id"] == raw["permno"].cast(df["id"].dtype)).all()
  ```

- [ ] **[test_combine_crsp_comp_sf.py:1810–1812]** The NaN/null masking in `test_monthly_numerical_values_match` uses `np.where(d is None, ...)` — a scalar `is` check that never detects elementwise nulls. This works by coincidence because Polars `.to_numpy()` already maps nulls to NaN for float columns, but it's misleading. Simplify to:
  ```python
  np.testing.assert_allclose(d, p, rtol=1e-10, atol=1e-12, equal_nan=True,
                             err_msg=f"Monthly column {c} mismatch")
  ```

- [ ] **[Duplicate examples — response to my question]** Can you provide a concrete example from real data showing `(id, eom)` duplicates that the dedup removes? From code analysis, the mechanism is: Compustat records with the same `(gvkey, iid, eom)` but differing `primary_sec` produce the same constructed `id` — the new `ROW_NUMBER()` deterministically picks `primary_sec=1`. A real-data example would confirm this is the only source of duplicates.

- [ ] **[test_combine_crsp_comp_sf.py:~300–475]** The `_polars_combine_crsp_comp_sf` reference implementation uses bare `col()` via `from polars import col` (~30 occurrences). Per CLAUDE.md, new code should use `pl.col()`.

- [ ] **[aux_functions.py:1851]** Add `-> None` return type annotation to `combine_crsp_comp_sf()`. The function body is completely rewritten, so this aligns with the convention for new code.

- [ ] **[aux_functions.py:1885, 2098]** Replace `os.system("rm -f aux_combine_sf.ddb")` with `Path("aux_combine_sf.ddb").unlink(missing_ok=True)` — more portable and avoids spawning a subshell. (This naturally folds into the try/finally fix above.)

- [ ] **[test_combine_crsp_comp_sf.py:295–298]** The reference implementation docstring says it "mirrors the old helpers exactly" but it actually implements deterministic dedup (matching the *new* intended behavior). Consider updating the wording to "deterministic reference for intended behavior" to avoid future confusion.

FernandoRDLL:
> @FernandoRDLL: can you show an example of duplicates that we want to remove? (as per our conversation)
Hey @theisij:

All (id, eom) duplicates originate in add_primary_sec(), not in combine_crsp_comp_sf itself. The range-joins against
  prihistrow/prihistusa/prihistcan produce row multiplication when a gvkey has overlapping date windows, and SELECT DISTINCT
  doesn't collapse them because the resulting primary_sec values differ.

  - __comp_sf2.parquet (input to add_primary_sec): 0 duplicates
  - comp_msf.parquet (output): 405 duplicate (gvkey, iid, datadate) pairs
  - Only two gvkeys affected: 327360 (404 pairs) and 339031 (1 pair)
  - 361 pairs are fully identical rows; 44 pairs differ only on primary_sec (one has 1, the other 0)

  The ORDER BY primary_sec DESC tie-break in combine_crsp_comp_sf correctly resolves all 44 meaningful duplicates. The obs_main
  GROUP BY id, eom added in this PR avoids the transient many-to-many row inflation during the JOIN.

  The upstream fix (deduplicating inside add_primary_sec itself) is a issue #69 
  
  I already incorporated your comments in here so it's ready to go

FernandoRDLL:
What we did in this branch:

Before the rewrite, duplicate (id, eom) rows could resolve in arbitrary order depending on the engine's internal row ordering. Now ROW_NUMBER() ... ORDER BY source_crsp DESC, primary_sec DESC makes a deterministic choice:

  1. Prefer CRSP over Compustat (though id spaces don't currently overlap, so this is defensive)
  2. Prefer primary_sec = 1 over 0 on ties

  The 51 tests validate this, including test_dedup_prefers_primary_sec_monthly and test_dedup_prefers_primary_sec_daily which
  specifically check the tie-break logic.

theisij (approved):
This looks good. Nice catch. I'll approve and merge

PR #85, opened 2026-04-14
Add unified CLI with src/ layout

### LINKED ISSUE #82: [Feature]: Add unified CLI entry point (`jkp`) using Typer
## Feature Description

Add a unified command-line interface (`jkp`) to consolidate all pipeline operations under a single entry point with subcommands, similar to `aws`, `gh`, and `uv`.

**Current state:**
```bash
uv run python code/main.py --persistent-connection
uv run python code/portfolio.py
```

**Proposed:**
```bash
jkp build --persistent-connection
jkp portfolio
jkp connect --reset
```

Commands:
- `jkp build` — Run the full data generation pipeline (wraps `main.py`)
- `jkp portfolio` — Generate factor portfolios (wraps `portfolio.py`)
- `jkp connect` — Test/configure WRDS connection

## Use Case / Motivation

- **Discoverability**: `jkp --help` shows all available commands
- **Tab completion**: Built-in shell completion for commands and options
- **Consistency**: Shared global options and standardized interface
- **Easier onboarding**: Single tool to learn, not scattered scripts
- **Professional polish**: Matches industry-standard CLI patterns

## Proposed Implementation (Optional)

- Use [Typer](https://typer.tiangolo.com/) for the CLI framework
- Create `code/cli.py` as the entry point
- Refactor `main.py` and `portfolio.py` to expose importable functions (`run_pipeline()`, `run_portfolio()`)
- Add `[project.scripts]` entry point in `pyproject.toml`: `jkp = "code.cli:app"`
- Preserve backward compatibility with direct `uv run python code/main.py` invocation

## Alternatives Considered

- **Click**: More verbose than Typer, which is built on Click but uses type hints for less boilerplate
- **argparse**: Already used in `main.py`, but doesn't support subcommands as ergonomically
- **No CLI**: Keep the current `uv run python code/...` pattern, but this doesn't scale well as more entry points are added

capellini:
## Summary

- Replace procedural script entry points (`python code/main.py`, `python code/portfolio.py`) with a Typer-based CLI providing three commands: `jkp build`, `jkp portfolio`, `jkp connect`
- Move `code/` to `src/jkp/data/` using PEP 420 implicit namespace packages, enabling future `jkp.*` packages (jkp.api, jkp.website, etc.)
- Add mandatory positional `output_dir` argument so pipeline output is user-specified rather than hardcoded to a `data/` directory inside the repo
- Bundle static reference data (Siccodes49.txt, cluster_labels.csv, country_classification.xlsx, factor_details.xlsx) as package resources via `importlib.resources`
- Remove `data/` directory from the repo (created at runtime in the user-specified output directory)

Closes #82

## Change Type

- [ ] Documentation only

## Methodological Impact

None

## Data Impact

None. The pipeline produces identical output — verified via a full end-to-end run on the Yale HPC (`jkp build data` + `jkp portfolio data`). Country characteristic files are identical in shape and columns to the previous pipeline output.

## Breaking Changes

- The `code/` directory is removed. Users must use `jkp build <output_dir>` and `jkp portfolio <output_dir>` instead of `python code/main.py` and `python code/portfolio.py`.
- The `data/` directory is no longer checked into the repo. Output is written to the user-specified directory.
- Python imports change from `from aux_functions import ...` to `from jkp.data.aux_functions import ...`.

## Tests

- Added `tests/unit/test_cli.py` with 15 tests covering help output, argument parsing, command routing, and missing argument validation for all three CLI commands
- Added `tests/unit/test_portfolio.py` with 2 tests verifying output format parameter forwarding
- All 295 tests pass locally and on Yale HPC
- Full pipeline run completed successfully on Yale HPC using the new CLI and namespace package

## Checklist

### Architecture

The package uses a `src/` layout with PEP 420 implicit namespace packages:

```
src/
  jkp/                    # namespace package (no __init__.py)
    data/                 # main package
      __init__.py
      cli.py              # Typer CLI entry point
      paths.py            # DataPaths dataclass + resource accessors
      resources/          # bundled static data
        Siccodes49.txt
        cluster_labels.csv
        country_classification.xlsx
        factor_details.xlsx
      main.py             # run_pipeline()
      portfolio.py        # run_portfolio()
      aux_functions.py    # core library
      ...
```

### Key design decisions

- **Namespace package**: `jkp` is the namespace, `jkp.data` is the package. Future packages (`jkp.api`, `jkp.website`, `jkp.wrds`) can share the namespace.
- **Positional output_dir**: Required argument, not optional with a default. Users must explicitly specify where output goes, which is necessary for an installable package (site-packages is not a writable output location).
- **Kept `os.chdir()`**: The pipeline uses two `os.chdir()` calls to set working directory (interim, then processed). These are parameterized via `DataPaths` but not removed, to avoid touching ~200 hardcoded relative path references throughout `aux_functions.py`.
- **Bundled resources**: Static reference data is shipped with the package via `importlib.resources`. `portfolio.py` now reads `cluster_labels.csv` and `country_classification.xlsx` from the package instead of downloading from GitHub at runtime.


capellini:
Leaving this in draft until I receive feedback from @theisij and @FernandoRDLL about implementation (see #82).

theisij:
This is great, Rob. Thanks a lot!! Moving to this format and, eventually, a full-blown package will take the codebase to the next step. 

My only question: How will the way we improve the codebase change when moving to a package? Currently, we're just merging PRs as they are ready, but I imagine we'll need to be a bit more strict about versioning and the like once we transition to a package format. 

@FernandoRDLL: let me know if you have any inputs on this

Other than that, let me know when the PR is ready for review (the only tiny thing I'd flag now is that we should probably update the README once these new changes take effect)!

capellini:
@FernandoRDLL I've added [an authors section](https://github.com/bkelly-lab/jkp-data/pull/85/changes#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R12-R17) to the pyproject.toml file. Would you like to be included? If so, what name should I use in the file?

capellini:
> How will the way we improve the codebase change when moving to a package? Currently, we're just merging PRs as they are ready, but I imagine we'll need to be a bit more strict about versioning and the like once we transition to a package format.

 There are a multitude of decisions to make in regards to the package management/maintenance, but the ones that come to mind are:
- Branching strategy - how do we manage bugs, features, and other development? Does main always represent a releasable state? There are several branching strategies and we'd have to choose one.
- Release process - When and how to cut package releases (on every commit, every day, every week, etc)?
- Semver management - What constitutes a breaking change? How long do we support previous releases? PR title prefixes indicating semver impact (e.g., feature, fix, docs, refactor, test, etc)
- Dependency management - How do we manage our dependencies? When do we bump versions? Security patches.
- CI/CD - We'd need some formal tests to increase confidence that we're not breaking things as we release new versions of our package
- Documentation - Keeping docs up to date, migration guides for major version bumps.

But at the end of the day, we're still going to be creating issues/PRs and merging them. I can manage most of the formality (though the process would be documented for continuity purposes).

We can discuss and evolve our decisions for the items I listed above as we go. But there are a couple of other changes that we would need before I'd go about publishing on PyPI. There are some things we do in our code that we shouldn't (e.g., use `os.system`). I plan to submit issues and PRs to fix these. We should also seriously consider prioritizing our testing.

capellini:
> we should probably update the README once these new changes take effect

Absolutely -- the README.md was updated. The data/README.md file is now a package resource that will be copied into the pipeline output data directory when the pipeline is run.

capellini:
@theisij Ready for review.

slurm/submit_job_som_hpc.slurm
```diff
@@ -23,13 +23,16 @@ export OMP_NUM_THREADS=${SLURM_CPUS_PER_TASK}
 export MKL_NUM_THREADS=${SLURM_CPUS_PER_TASK}
 export NUMEXPR_NUM_THREADS=${SLURM_CPUS_PER_TASK}
 
+# Activate the project's virtual environment so the jkp CLI is on PATH
+source .venv/bin/activate
+
 # Create char data
 # Set PERSISTENT_WRDS_CONNECTION=1 to use a single WRDS connection (reduces MFA on NAT-rotated networks)
 if [ "${PERSISTENT_WRDS_CONNECTION:-0}" = "1" ]; then
```
theisij, line 31: Is there any reasonable use case where one would not like to have a persistent connection? It seems to me like users would also want to have a persistent connection. If there are no reasonable cases, we could consider making '--persistent-connection' the default

src/jkp/data/aux_functions.py
```diff
@@ -67,22 +74,35 @@ def wrapper(*args, **kwargs):
...
+        Folders created on disk (no return value). Working directory set to paths.interim_dir.
     """
-    os.chdir(os.path.join(os.path.dirname(__file__), "..", "data/interim"))
-    os.system(
-        "mkdir -p raw_data_dfs ../raw/raw_tables ../processed/characteristics ../processed/return_data ../processed/accounting_data ../processed/other_output"
+    import shutil
+
+    from .paths import get_data_readme_path
+
+    paths.interim_dir.mkdir(parents=True, exist_ok=True)
```
theisij, line 94: It actually feels a bit envasive to create a new directory on a users computer. Will something like this be accepted once we make it into a package? I guess it's okay as long as we're explcit about it, but I just wanted to flag it

src/jkp/data/aux_functions.py
```diff
@@ -67,22 +74,35 @@ def wrapper(*args, **kwargs):
...
+    (paths.processed_dir / "characteristics").mkdir(parents=True, exist_ok=True)
+    (paths.processed_dir / "return_data" / "daily_rets_by_country").mkdir(
+        parents=True, exist_ok=True
     )
-    os.system("mkdir -p ../processed/return_data/daily_rets_by_country")
+    (paths.processed_dir / "accounting_data").mkdir(parents=True, exist_ok=True)
+    (paths.processed_dir / "other_output").mkdir(parents=True, exist_ok=True)
+    (paths.processed_dir / "portfolios").mkdir(parents=True, exist_ok=True)
+    shutil.copy2(get_data_readme_path(), paths.base_dir / "README.md")
+    os.chdir(paths.interim_dir)
```
theisij, line 105: You mentioned our use of os.chdir, which is quite ugly, but I agree with you that we should leave this here for now, and change it later

CLAUDE.md
```diff
@@ -15,11 +15,11 @@ For complete setup and contribution instructions, see [CONTRIBUTING.md](CONTRIBU
...
 
 # Run linting
-uv run ruff check code/ tests/
-uv run ruff format --check code/ tests/
+uv run ruff check src/jkp/data/ tests/
+uv run ruff format --check src/jkp/data/ tests/
 
 # Run type checking (informational, not blocking in CI)
-uv run pyright code/
+uv run pyright src/jkp/data/
```
theisij, line 22: Should this be 'uv run pyright ..' or just 'pyright ..'. In CONTRIBUTING.md it's the latter.

CONTRIBUTING.md
```diff
@@ -46,24 +46,24 @@ We use [ruff](https://docs.astral.sh/ruff/) for linting and formatting. Before s
...
+ruff format src/jkp/data/ tests/
 ```
 
 ## Type Checking
 
 We use [pyright](https://github.com/microsoft/pyright) for static type checking:
 
 ```bash
-uv run pyright code/
+pyright src/jkp/data/
```
theisij, line 68: Should this be 'uv run pyright ..' or 'pyright ..'. In CLAUDE.md it's the former.

pyproject.toml
```diff
@@ -1,26 +1,61 @@
...
-description = "Add your description here"
+description = "Global factor data generation pipeline from Jensen, Kelly, and Pedersen (2023)."
 readme = "README.md"
+license = "MIT"
 requires-python = ">=3.11"
+authors = [
+    {name = "Bryan Kelly"},
+    {name = "Theis Jensen"},
+    {name = "Lasse Pedersen"},
+    {name = "Rob Capellini"},
```
theisij, line 16: Let's add Fernando here

theisij (dismissed):
This looks great!! My comments are mostly clarifying and shouldn't stop the PR

theisij:
> > How will the way we improve the codebase change when moving to a package? Currently, we're just merging PRs as they are ready, but I imagine we'll need to be a bit more strict about versioning and the like once we transition to a package format.
> 
> There are a multitude of decisions to make in regards to the package management/maintenance, but the ones that come to mind are:
> 
> * Branching strategy - how do we manage bugs, features, and other development? Does main always represent a releasable state? There are several branching strategies and we'd have to choose one.
> * Release process - When and how to cut package releases (on every commit, every day, every week, etc)?
> * Semver management - What constitutes a breaking change? How long do we support previous releases? PR title prefixes indicating semver impact (e.g., feature, fix, docs, refactor, test, etc)
> * Dependency management - How do we manage our dependencies? When do we bump versions? Security patches.
> * CI/CD - We'd need some formal tests to increase confidence that we're not breaking things as we release new versions of our package
> * Documentation - Keeping docs up to date, migration guides for major version bumps.
> 
> But at the end of the day, we're still going to be creating issues/PRs and merging them. I can manage most of the formality (though the process would be documented for continuity purposes).
> 
> We can discuss and evolve our decisions for the items I listed above as we go. But there are a couple of other changes that we would need before I'd go about publishing on PyPI. There are some things we do in our code that we shouldn't (e.g., use `os.system`). I plan to submit issues and PRs to fix these. We should also seriously consider prioritizing our testing.

Thanks a lot for the elaborate explanation! 

I've admittedly pulled Fernando a bit away from testing, but let's reprioritize some of his development time towards testing. Do you have some specific tests in mind, or is is more about making sure that we reach something close to 100% test coverage? 

Also, given that Opus 4.7 was released today, I was thinking it might make sense to have a structured review process whenever a new SOTA model is released. Currently, a lot of our code is written with coding agents, so if something better comes out, it would be good to have specific parts of the code that it checks for possible improvements. For example, we could establish a procedure for reviewing all tests, with specific guidance on what we're looking for (maintaining a consistent style, removing redundancy, making sure we capture edge cases, etc.). This is probably a task for later, but I wanted to flag it while the idea was fresh in mind.

slurm/submit_job_som_hpc.slurm
capellini: Good question. The two modes have different trade-offs:

**Default (no --persistent-connection)**: Uses `postgres_scan()` which creates a new TCP connection per table download. This is actually the more robust mode. Each query is independent, so if one table download fails, the others aren't affected. It also doesn't hold a long-lived database connection open for the entire ~35 minute download phase. On most networks this works fine and only triggers a single MFA prompt.

**`--persistent-connection`**: Uses DuckDB's `ATTACH` to maintain a single PostgreSQL connection for all ~26 table downloads. This was added specifically for HPC clusters with NAT IP rotation (like Yale's Bouchet cluster), where the gateway assigns a random outbound IP to each new TCP connection. WRDS interprets connections from different IPs as different locations and triggers MFA for each, resulting in ~26 MFA prompts instead of 1. The trade-off is that a single long-lived connection is less resilient: if the connection drops mid-download, all remaining tables fail.

Additionally, I'd like to parallelize the WRDS downloads in a future PR, which would be incompatible with persistent connection mode, since parallel downloads require independent connections per table. The default mode already provides this functionality.

So the default is the safer, more portable, and more future-proof mode. `--persistent-connection` solves a specific infrastructure problem at the cost of resilience and future parallelism.

src/jkp/data/aux_functions.py
capellini: The key difference from the old behavior is that previously the pipeline wrote to `data/` inside the repo silently, which is arguably worse. A user cloning the repo would have data written into their git working tree with no warning.

Now the user explicitly specifies the output directory as a required positional argument (`jkp build data/`), so they're opting in. We also added a `--force` flag. If the directory already contains data, the CLI prompts for confirmation before overwriting (unless `--force` is passed). So the directory creation is always explicitly requested by the user, never implicit.

src/jkp/data/aux_functions.py
capellini: Agreed. I have a list of things that need to get done before we publish to PyPI, and this is one of them. (I will publish a main issue about this soon)

The plan would be to replace the bare relative paths with explicit path construction so os.chdir() can be removed entirely.

CLAUDE.md
capellini: CLAUDE.md keeps uv run because it's instructions for Claude Code specifically, where uv run ensures the right environment regardless of shell state. CONTRIBUTING.md dropped the uv run prefix to be indifferent as to the choice of tool.

The assumption is that the contributor has the package and its dev dependencies installed, however they choose to manage that. Both are intentional, but I'll add a note to CONTRIBUTING.md clarifying that commands assume the project's dependencies are available in your current environment.

CONTRIBUTING.md
capellini: See other comment RE: differences between CLAUDE.md and CONTRIBUTING.md.

pyproject.toml
capellini: Absolutely, that's my intention. I've actually [asked Fernando about this](https://github.com/bkelly-lab/jkp-data/pull/85#issuecomment-4257609134). However, I'm ashamed to admit that I do not know his full name!

Once I hear back from him, I'll be sure to place his name in there and make the change to CONTRIBUTING.md and push another commit to this pull request.

@FernandoRDLL Please let me know how you'd like to be listed in the authors section when you have a moment.

capellini:
> I've admittedly pulled Fernando a bit away from testing, but let's reprioritize some of his development time towards testing. Do you have some specific tests in mind, or is is more about making sure that we reach something close to 100% test coverage?

I'll provide my thoughts on this in the GitHub issue.  Fernando's input would also be key since he is the primary contributor to the Python codebase. He probably has insights into which pieces are more critical on the factor estimation side.

As long as we're focused on getting as close to 100% test coverage as possible in the near future, I don't think this would be a blocker to being published on PyPI. But getting there sooner rather than later is a good goal and would help us move faster and more confidently.
 
> Also, given that Opus 4.7 was released today, I was thinking it might make sense to have a structured review process whenever a new SOTA model is released. Currently, a lot of our code is written with coding agents, so if something better comes out, it would be good to have specific parts of the code that it checks for possible improvements. For example, we could establish a procedure for reviewing all tests, with specific guidance on what we're looking for (maintaining a consistent style, removing redundancy, making sure we capture edge cases, etc.). This is probably a task for later, but I wanted to flag it while the idea was fresh in mind.

Yes, that's a great point. Right now, I think we would be overwhelmed with numerous comments that would be redundant with the comments I'm about to post about prerequisites to PyPI publishing.  As long as we have a centralized plan going forward and we're all on the same page, I think this is a worthwhile effort. We just have to make sure that we don't work redundantly or work in ways that would create a lot of merge conflicts.

But I will add creating such a structured review to my todo list.

slurm/submit_job_som_hpc.slurm
theisij: Great, thanks for the elaboration

pyproject.toml
theisij: Sorry about the later reply here. It's: Fernando Reyes De La Luz

src/jkp/data/aux_functions.py
theisij: Okay, that's a good point. No problem then

theisij (approved):
Looks good! Approviong and merging

PR #110, opened 2026-04-23
Change command from 'jkp connect' to 'uv run jkp connect'

FernandoRDLL:
Updated command to connect to WRDS credentials.
@theisj and @capellini we need 'uv run' so that we don't sync the whole environment Please approve

## Change Type

- [ ] Documentation only

## Checklist

- [ ] I have run the tests locally (`pytest`)
- [ ] I have run the linter locally (`ruff check src/jkp/data/ tests/`)
- [ ] I have updated documentation if needed
- [ ] I have considered whether this change affects factor calculations

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ This repo contains Python code to generate the global dataset of factor returns,
 
    - To save your WRDS credentials, navigate to the `jkp-data/` folder and run:
      ```sh
-     jkp connect
+     uv run jkp connect
      ```
      Kindly follow the prompts.
 


capellini:
> @TheisJ and @capellini we need 'uv run' so that we don't sync the whole environment

We're trying to be indifferent to the user's choice of package manager, since we're moving this to a PyPI python package. I made some comments in CONTRIBUTING.md about this, too. If you prefer `uv`, then by all means prefix your commands with `uv run`, but I don't believe we should mandate that our users use `uv`.

Pro tip: if you just activate the uv virtualenv, you no longer have to prefix everything with `uv run` and can still take advantage of tab completion with the `jkp` utility:

```python
source .venv/bin/activate
jkp ...
```

I also see this in the README.md, which I missed

```
   - To save your WRDS credentials, navigate to the `jkp-data/` folder and run:
```

There is no longer a need to navigate to any directory. In fact, using certain package managers (.e.g, `pip`), the install directory will be somewhere else (e.g., `/usr/lib...`, or buried in the user's home directory).

The idea is that users now run the `jkp` commands and specify an output directory wherever they'd like. This is definitely a paradigm shift -- going from an internal package to a PyPI package.

I'm a bit swamped at the moment, but I will definitely take a closer look at the README.md. Thank you for bringing this to my attention, @FernandoRDLL. If there's anything I'm overlooking, please let me know.

FernandoRDLL:
@copilot please review

capellini:
@FernandoRDLL I was actually just about to close this issue. We should not be mentioning `uv` at all in our README.md.

What is really needed is a complete revamp of the README.md which doesn't mention `uv` or navigating to a directory at all. Do you want me to do this?

FernandoRDLL:
That would be great, thank you

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #130, opened 2026-04-30
Stop forcing PlaintextKeyring at import; require explicit opt-in

### LINKED ISSUE #101: [PyPI v0.1.0] T1.3: PlaintextKeyring — stop forcing at import, add env-var fallback
## Problem

`wrds_credentials.py` sets `PYTHON_KEYRING_BACKEND=keyrings.alt.file.PlaintextKeyring` at module import time. Any user who imports from `jkp.data` has their keyring backend silently switched to plaintext-on-disk for the rest of the process — no warning, no documentation, security implication not surfaced.

## Solution

Tiered credential resolution:
1. **Default:** use the system keyring (`keyring.get_password`). On macOS/Linux desktop this is Keychain/Secret Service.
2. **Fallback:** environment variables `WRDS_USERNAME` / `WRDS_PASSWORD` (standard for CI and HPC).
3. **Opt-in plaintext:** require `JKP_ALLOW_PLAINTEXT_KEYRING=1`. When set, emit a loud warning and activate `PlaintextKeyring` only inside the credential-resolution function (not globally).
4. **Never** mutate `os.environ["PYTHON_KEYRING_BACKEND"]` at import time.

Document the precedence in README and `jkp connect --help`.

## Files

- `src/jkp/data/wrds_credentials.py`
- `README.md`
- `src/jkp/data/cli.py` (connect subcommand help text)
- `tests/unit/test_wrds_credentials.py` (new or expanded)

## Acceptance

```bash
python -c "import os; env_before = dict(os.environ); from jkp.data import wrds_credentials; assert os.environ == env_before"
```
- Env-var fallback works in a clean container
- Plaintext only activates with the opt-in flag and prints a warning

capellini:
## Summary

`src/jkp/data/wrds_credentials.py` previously set `PYTHON_KEYRING_BACKEND=keyrings.alt.file.PlaintextKeyring` at module-import time, which silently changed the keyring backend for every process that imported anything from `jkp.data`. That backend swap is appropriate for headless environments (HPC compute nodes, minimal Docker images) where no system keyring daemon is available, but doing it silently for every user is undocumented behavior.

This PR replaces that with a tiered, explicit credential resolver:

1. `WRDS_USERNAME` / `WRDS_PASSWORD` environment variables (containers, shared service accounts).
2. The system keyring (Keychain, Secret Service, Credential Vault) — default for interactive sessions.
3. The file-backed keyring (`keyrings.alt.file.PlaintextKeyring`, a mode-600 file under `~/.local/share/python_keyring/`) — selected only when `JKP_ALLOW_PLAINTEXT_KEYRING=1` is set, with a warning emitted on first use (once per process) so the backend change is never silent.

The package never mutates `os.environ` at import time. Documentation updated in README and `jkp connect --help`. The Slurm example script now opts in to the file-backed keyring explicitly, matching the HPC compute-node use case.

This is a Tier 1 prerequisite for publishing the package to PyPI.

Closes #101.

## Change Type

- [ ] Documentation only

## Methodological Impact

None.

## Data Impact

None.

## Breaking Changes

Existing users on machines with a working system keyring see no behavior change other than the backend no longer being silently overridden.

Existing users who relied on the previous silent activation of the file-backed keyring (for example on HPC compute nodes that have no keyring daemon) need to set `JKP_ALLOW_PLAINTEXT_KEYRING=1` in their environment. The updated `slurm/submit_job_som_hpc.slurm` example shows the pattern, and the warning emitted on first use makes the required action obvious.

## Tests

Added five new tests in `tests/unit/test_wrds_credentials.py`:

- `test_import_does_not_mutate_environ` — the regression test for the original bug.
- `test_env_vars_take_precedence` — env-var path returns those values without touching the keyring.
- `test_env_partial_falls_through_to_keyring` — only one of the two env vars set falls through to keyring resolution.
- `test_plaintext_opt_in_emits_warning` — the opt-in flag fires the documented warning.
- `test_plaintext_no_optin_no_warning` — without the opt-in, no warning, no backend swap.

All five pass; full unit suite passes (370 tests, 0 failures).

Manual validation on a Yale HPC compute node (job 571841) covered four scenarios with no `$HOME` writes:

1. Module import on a compute node does not mutate the environment.
2. Env-var precedence returns the env-var values without touching the keyring.
3. The file-backed keyring round-trips a stored password (write then read), with the documented warning emitted on each backend swap.
4. The smoke test never touches any real `$HOME` files (verified via a scratch directory and a sentinel check).

## Checklist

- [x] I have verified the pipeline runs successfully after my changes — manual smoke test on Yale compute node verified the credential resolution paths end to end
- [x] I have updated documentation if needed — README, `jkp connect --help`, and the Slurm example script
- [x] I have considered whether this change affects factor calculations — it does not

## Additional Notes

The mode-600 file written by `keyrings.alt.file.PlaintextKeyring` has the same security posture as any other user-private dotfile (e.g. an exported `~/.wrds_secrets`); the concern this PR addresses is the silent backend mutation at import time, not the storage format itself.

### DIFF
--- a/src/jkp/data/wrds_credentials.py
+++ b/src/jkp/data/wrds_credentials.py
@@ -1,8 +1,35 @@
-import os
+"""WRDS credential resolution.
 
-os.environ["PYTHON_KEYRING_BACKEND"] = "keyrings.alt.file.PlaintextKeyring"
+Credential precedence (highest first):
+
+1. ``WRDS_USERNAME`` / ``WRDS_PASSWORD`` environment variables. Useful for
+   container deployments and shared service accounts where the source of
+   truth shouldn't be a per-user filesystem location.
+2. The system keyring (Keychain on macOS, Secret Service on Linux desktop,
+   Credential Vault on Windows). The default for interactive sessions.
+3. The file-backed keyring (``keyrings.alt.file.PlaintextKeyring``), which
+   stores the password in a mode-600 file under ``~/.local/share/python_keyring/``.
+   This is appropriate for headless environments (HPC compute nodes,
+   minimal Docker images) where no system keyring daemon is available — the
+   stored secret has the same security posture as any other 600-mode dotfile
+   in the user's home directory.
+
+   This backend is not selected automatically. The previous implementation
+   set ``PYTHON_KEYRING_BACKEND=keyrings.alt.file.PlaintextKeyring`` at
+   module-import time, which silently changed the keyring backend for every
+   process that imported anything from ``jkp.data`` — surprising, undocumented,
+   and unnecessary on machines where a system keyring is available. Now the
+   user must opt in with ``JKP_ALLOW_PLAINTEXT_KEYRING=1``, and the swap
+   happens only inside this module's credential-resolution functions, never
+   at import time.
+"""
+
+from __future__ import annotations
+
 import argparse
 import getpass
+import os
+import warnings
 from dataclasses import dataclass
 from pathlib import Path
 
@@ -11,6 +38,11 @@ import keyring
 SERVICE_NAME = "WRDS"
 LAST_USER_FILE = Path.home() / ".wrds_user"  # remembers last username
 
+# Environment variable names.
+ENV_USERNAME = "WRDS_USERNAME"
+ENV_PASSWORD = "WRDS_PASSWORD"
+ENV_ALLOW_PLAINTEXT = "JKP_ALLOW_PLAINTEXT_KEYRING"
+
 
 @dataclass(frozen=True)
 class Credentials:
@@ -18,23 +50,61 @@ class Credentials:
     password: str
 
 
+def _maybe_use_file_keyring() -> None:
+    """If the user explicitly opts in, switch the keyring backend to the
+    file-backed ``keyrings.alt.file.PlaintextKeyring``.
+
+    Called at credential-resolution time, never at import. The backend stores
+    the password in a mode-600 file under ``~/.local/share/python_keyring/``;
+    the security posture is the same as any other user-private dotfile.
+    """
+    if os.environ.get(ENV_ALLOW_PLAINTEXT) != "1":
+        return
+    warnings.warn(
+        f"{ENV_ALLOW_PLAINTEXT}=1 is set: switching the keyring backend to "
+        "keyrings.alt.file.PlaintextKeyring for this process. The WRDS "
+        "password will be read from / written to a mode-600 file under "
+        "~/.local/share/python_keyring/.",
+        stacklevel=2,
+    )
+    from keyrings.alt.file import PlaintextKeyring  # noqa: PLC0415
+
+    keyring.set_keyring(PlaintextKeyring())
+
+
+def _credentials_from_env() -> Credentials | None:
+    """Return env-var credentials if both ``WRDS_USERNAME`` and ``WRDS_PASSWORD`` are set."""
+    user = os.environ.get(ENV_USERNAME)
+    password = os.environ.get(ENV_PASSWORD)
+    if user and password:
+        return Credentials(user, password)
+    return None
+
+
 def get_wrds_credentials() -> Credentials:
+    """Resolve WRDS credentials following the documented precedence order.
+
+    Steps:
+      1. If ``WRDS_USERNAME`` and ``WRDS_PASSWORD`` are both set, return those.
+      2. Otherwise, look up the saved username in ``~/.wrds_user`` (prompt
+         interactively on first run).
+      3. Fetch the password from the (possibly plaintext-opt-in) keyring;
+         prompt and store on first run.
     """
-    Automatically retrieves credentials for wrds.
-    - On first run: asks for username and password/token, stores them.
-    - On later runs: loads both silently from the system keyring.
-    """
-    # Try to remember the last-used username
+    env_creds = _credentials_from_env()
+    if env_creds is not None:
+        return env_creds
+
+    _maybe_use_file_keyring()
+
     if LAST_USER_FILE.exists():
         username = LAST_USER_FILE.read_text().strip()
     else:
         username = input(f"Username for {SERVICE_NAME}: ").strip()
         LAST_USER_FILE.write_text(username)
 
-    # Try to retrieve the stored password for this username
     password = keyring.get_password(SERVICE_NAME, username)
 
-    # If not found, prompt once and store it securely
     if not password:
         password = getpass.getpass(f"Password or token for {username} at {SERVICE_NAME}: ")
         keyring.set_password(SERVICE_NAME, username, password)
@@ -43,10 +113,9 @@ def get_wrds_credentials() -> Credentials:
     return Credentials(username, password)
 
 
-def reset_credentials(full_reset: bool = False):
-    """
-    Clears stored username and optionally removes password from keyring.
-    """
+def reset_credentials(full_reset: bool = False) -> None:
+    """Clear the stored username and (optionally) remove the password from the keyring."""
+    _maybe_use_file_keyring()
     if LAST_USER_FILE.exists():
         username = LAST_USER_FILE.read_text().strip()
         LAST_USER_FILE.unlink()
--- a/src/jkp/data/wrds_credentials.py
+++ b/src/jkp/data/wrds_credentials.py
@@ -54,6 +54,12 @@ def _maybe_use_file_keyring() -> None:
     """If the user explicitly opts in, switch the keyring backend to the
     file-backed ``keyrings.alt.file.PlaintextKeyring``.
 
+    Opt-in: set the ``JKP_ALLOW_PLAINTEXT_KEYRING`` environment variable to
+    exactly the string ``"1"``. Any other value (including ``"true"``,
+    ``"yes"``, ``"True"``, or ``"0"``) is treated as not set and the system
+    keyring is used. We use a strict comparison rather than truthy parsing so
+    that the opt-in is unambiguous and matches the documented form.
+
     Called at credential-resolution time, never at import. The backend stores
     the password in a mode-600 file under ``~/.local/share/python_keyring/``;
     the security posture is the same as any other user-private dotfile.
--- a/src/jkp/data/wrds_credentials.py
+++ b/src/jkp/data/wrds_credentials.py
@@ -133,7 +133,8 @@ def _keyring_backend() -> Iterator[_WrdsKeyring]:
 
 
 def _credentials_from_env() -> Credentials | None:
-    """Return env-var credentials if both ``WRDS_USERNAME`` and ``WRDS_PASSWORD`` are set."""
+    """Return env-var credentials if both ``WRDS_USERNAME`` and ``WRDS_PASSWORD`` are set to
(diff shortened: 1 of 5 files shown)


capellini:
@FernandoRDLL Could I please get a code review for these changes? Thank you.

src/jkp/data/wrds_credentials.py
```diff
@@ -11,30 +38,79 @@
...
+    _maybe_use_file_keyring()
+
     if LAST_USER_FILE.exists():
         username = LAST_USER_FILE.read_text().strip()
     else:
         username = input(f"Username for {SERVICE_NAME}: ").strip()
         LAST_USER_FILE.write_text(username)
 
-    # Try to retrieve the stored password for this username
     password = keyring.get_password(SERVICE_NAME, username)
```
FernandoRDLL: I think we need to improve the error message when no keyring backend is available.

Right now, a user running in a headless environment without a configured keyring will get keyring’s generic NoKeyringError("No recommended backend was available..."), which doesn’t provide any guidance on how to proceed. Since this change introduces JKP_ALLOW_PLAINTEXT_KEYRING as an opt-in, it would be helpful to catch NoKeyringError and re-raise with a message that points users to that environment variable (and to WRDS_USERNAME / WRDS_PASSWORD as an alternative).

That could be done once and reused in both get_wrds_credentials() and reset_credentials().

src/jkp/data/wrds_credentials.py
```diff
@@ -43,10 +119,9 @@ def get_wrds_credentials() -> Credentials:
     return Credentials(username, password)
 
 
-def reset_credentials(full_reset: bool = False):
-    """
-    Clears stored username and optionally removes password from keyring.
-    """
+def reset_credentials(full_reset: bool = False) -> None:
+    """Clear the stored username and (optionally) remove the password from the keyring."""
+    _maybe_use_file_keyring()
```
FernandoRDLL: This path has the same issue as the credential lookup: on a headless system without a configured keyring, the delete_password call below raises NoKeyringError, and the existing except clause only covers PasswordDeleteError, so the error propagates uncaught from jkp connect --reset.

If a shared NoKeyringError handler is added for get_wrds_credentials(), it would be worth applying it here as well.

src/jkp/data/wrds_credentials.py
```diff
@@ -11,30 +38,79 @@
...
+    that the opt-in is unambiguous and matches the documented form.
+
+    Called at credential-resolution time, never at import. The backend stores
+    the password in a mode-600 file under ``~/.local/share/python_keyring/``;
+    the security posture is the same as any other user-private dotfile.
     """
-    Automatically retrieves credentials for wrds.
-    - On first run: asks for username and password/token, stores them.
-    - On later runs: loads both silently from the system keyring.
+    if os.environ.get(ENV_ALLOW_PLAINTEXT) != "1":
```
FernandoRDLL, line 74: The strict comparison against "1" is documented, so this is more of a usability suggestion. Right now, a user who sets JKP_ALLOW_PLAINTEXT_KEYRING=true (a fairly common shell convention) gets the opt-in silently ignored and will later run into a NoKeyringError with no indication of why. It might be worth emitting a warning when the variable is set to a truthy-looking value other than "1".

On a related note, the new tests cover "1" and the unset case, but not the rejection of values like "true" or "yes". Since the README explicitly documents that behavior, a small test for it would be a nice addition.

src/jkp/data/wrds_credentials.py
```diff
@@ -11,30 +38,79 @@
...
+    from keyrings.alt.file import PlaintextKeyring  # noqa: PLC0415
+
+    keyring.set_keyring(PlaintextKeyring())
+
+
+def _credentials_from_env() -> Credentials | None:
+    """Return env-var credentials if both ``WRDS_USERNAME`` and ``WRDS_PASSWORD`` are set."""
+    user = os.environ.get(ENV_USERNAME)
+    password = os.environ.get(ENV_PASSWORD)
+    if user and password:
```
FernandoRDLL, line 140: Minor edge case: because this uses a truthiness check, an environment variable set to an empty string (for example export WRDS_PASSWORD= from a templating mistake in CI) is treated as unset and silently falls through to the keyring path. The docstring describes the condition as "if both are set", so is not None checks would match that more closely. Alternatively, failing loudly on a partial or empty env configuration might be preferable to a silent fallthrough.

src/jkp/data/wrds_credentials.py
```diff
@@ -11,30 +38,79 @@
...
+         interactively on first run).
+      3. Fetch the password from the (possibly plaintext-opt-in) keyring;
+         prompt and store on first run.
     """
-    # Try to remember the last-used username
+    env_creds = _credentials_from_env()
+    if env_creds is not None:
+        return env_creds
+
+    _maybe_use_file_keyring()
```
FernandoRDLL: Not blocking, just a maintainability note. Every function that touches the keyring needs to remember to call _maybe_use_file_keyring() first (currently here and in reset_credentials()). With two callers that's fine, but a future keyring-using function that omits the call would break the opt-in silently on headless systems. Caching the swap once (for example with functools.cache) at a single resolution point would remove that implicit contract.

FernandoRDLL (commented):
Hey Rob, here are the comments

src/jkp/data/wrds_credentials.py
```diff
@@ -1,8 +1,35 @@
-import os
+"""WRDS credential resolution.
```
FernandoRDLL, line 1: Small documentation note: the credential precedence list now appears in four places (the README, the cli.py docstring, this module docstring, and the test file docstring). If the precedence ever changes, it's easy for one of those copies to be missed. It might be worth keeping this docstring as the canonical version and having the others reference it.

src/jkp/data/wrds_credentials.py
capellini: Agreed -- good catch. Done in 5ab638f. Added a shared `_keyring_backend()` context manager that catches keyring's generic `NoKeyringError` and re-raises it with guidance on the two ways to handle creds without a system keyring installed (`WRDS_USERNAME`/`WRDS_PASSWORD` and `JKP_ALLOW_PLAINTEXT_KEYRING=1`). Covered by `test_get_wrds_credentials_missing_backend_raises_guidance`.

src/jkp/data/wrds_credentials.py
capellini: Fixed in 5ab638f using the same `_keyring_backend()` wrapper, so `jkp connect --reset` now surfaces the same guidance instead of crashing uncaught on a headless box. The existing `PasswordDeleteError` handling ("no entry to delete") is preserved separately inside the facade. `NoKeyringError` is a different exception and is deliberately not caught there, so on a headless box it propagates up to the `_keyring_backend()` context manager, which re-raises it with the actionable guidance, rather than being misreported as an absent entry. Covered by `test_reset_credentials_missing_backend_raises_guidance`, with `test_reset_credentials_missing_entry_is_handled` guarding the no-entry path.

src/jkp/data/wrds_credentials.py
capellini: Agreed, and went a bit further than `functools.cache`. The swap is now in a `@functools.cache`'d `_ensure_keyring_backend()` (so it runs at most once per process -- env var read and warning emitted only on first use), and all keyring access routes through the `_keyring_backend()` context manager, which yields a small WRDS-scoped `_WrdsKeyring` facade.

Now `get_wrds_credentials()` and `reset_credentials()` no longer reference the keyring library or its exceptions at all. There's now a single, obvious way to access the keyring, so it's much harder for a future function to skip the opt-in by accident and a direct keyring call that bypasses it stands out in review.

src/jkp/data/wrds_credentials.py
capellini: I like the idea of a test for non-"1" values to confirm they're rejected and don't swap the backend. That's a nice guard against the check getting loosened down the road, so I'll add that.

On the warning itself, I went back and forth but I'm leaning against it, and I wanted to lay out the reasoning in case I'm missing something. My worry is that it nudges us toward over-engineering for a case the docs already cover, and the generous hint we added in 5ab638f now reinforces. The trickier part is that env vars are always strings, so a simple bool(var) check doesn't actually capture intent (bool("no") is True, for instance). So to do it properly we'd need an allow-list of "truthy-looking" tokens ("true", "yes", "on", "y", "t", and so on). We could just check for "true", but that feels a bit arbitrary, and I worry it could backfire: someone sees no warning for a value we didn't think to include and assumes they're all set when they aren't.

That said, this is very much a judgment call, and if you feel strongly about it I'm happy to implement something covering the most common cases -- just let me know.

src/jkp/data/wrds_credentials.py
capellini: Drift across four copies is a real risk. Rather than make the module docstring canonical, I made `jkp connect --help` the single source of truth, since that's the one surface a user can actually run to see the order. Done in 11c418b. The full precedence list (now including the exactly-"1" opt-in and the once-per-process warning note) lives in the connect command docstring, and the other three references (README, the wrds_credentials module docstring, and the test docstring) now point to `jkp connect --help` instead of restating it. The README keeps its HPC/Slurm opt-in guidance since that's operational and not a restatement of the order. Since the README is also the first place developers normally go for documentation, I also kept a very brief note on the ordering there, with a note to run `jkp connect --help` for more details.

src/jkp/data/wrds_credentials.py
capellini: That's a fair edge case to raise. This one's intentional, and I think the truthiness check is actually the safer behavior, so I'd lean toward keeping it.

Switching to `is not None` would make `WRDS_PASSWORD=` (empty) count as "set," so we'd return an empty password and hand `""` straight to WRDS, which then fails at connection time with a more opaque auth error. Falling through instead keeps a blank value from being treated as a real credential.

On failing loudly for a partial/empty config: Though tempting, it's a behavior change with its own ambiguity (is an empty string a "partial config" or just unset?), and the cost of the silent fallthrough is lower now than when you flagged it: after the `NoKeyringError` change in 5ab638f, a CI box that sets `WRDS_USERNAME` but leaves `WRDS_PASSWORD` empty falls through, hits the missing-backend path, and now gets the guidance error that explicitly names `WRDS_USERNAME`/`WRDS_PASSWORD`, pointing the user straight back at the variables they misconfigured. So the mistake still surfaces clearly, just one step later.

Correct, the docstring says "if both are set" while the code requires non-empty. Strictly speaking, an empty var is still "set", but treating an empty value as "not provided" is a common convention (it's what the shell's `${VAR:-default}` form does), so the truthiness check matches that expectation. I'll tighten the docstring to "set to non-empty values" to make it explicit.

Let me know if you feel strongly about the fail-loud option and I can revisit it.

capellini:
@FernandoRDLL Thank you for the review. Great feedback. I have a few open comments. If you feel strongly about them, I'll revisit them. If not, I can merge these changes into main after you approve the review.

FernandoRDLL (approved):

FernandoRDLL:
@capellini I just merged!

PR #133, opened 2026-05-01
Fix wall-clock date dependency in aux_functions (closes #131)

### LINKED ISSUE #131: [Data]: comp_hgics uses date.today() to fill open-ended GICS intervals, making output depend on wall-clock date
### Type of Discrepancy

Differs from previous run

### Affected Data Type

Industry portfolios (`gics` column propagating into firm characteristics).

### Country/Region Affected

Multiple countries — Compustat GICS history feeds the `gics` column on every world-level row.

### Comparison Baseline

Previous run of this codebase.

### Detailed Description

`comp_hgics` in `src/jkp/data/aux_functions.py` calls `date.today()` to fill the `indthru` (industry-record end-date) field for open-ended GICS intervals — those where Compustat records no end date because the company is still in that industry "as of today":

```python
indthru_date = (
    pl.lit(data[["indfrom"]].max()[0, 0])
    if data[["indfrom"]].max()[0, 0] > date.today()
    else pl.lit(date.today())   # <-- here
)
...
data = (
    data.with_columns(
        indthru=pl.when(c1 & c2).then(indthru_date).otherwise(col("indthru"))
    )
    .select(["gvkey", pl.date_ranges("indfrom", "indthru").alias("date"), "gics"])
    .explode("date")
    ...
)
```

The expression then feeds `pl.date_ranges("indfrom", "indthru").explode("date")`, so any change to `indthru` produces additional or fewer per-day rows in the exploded output. Because `date.today()` returns the wall-clock date of execution, two runs that fall on different dates produce different output: each currently-active company gets one extra exploded row per day that passes between runs. With ~250,000 currently-active Compustat companies, even a one-day gap between runs produces ~250,000 extra rows in `na_hgics.parquet`.

### Sample of Problematic Data

Two pipeline runs starting at 21:55 on the same evening — one finished at 04:17 the next morning, the other at 07:39:

| file | run-1 rows | run-2 rows | diff |
|---|---|---|---|
| `na_hgics.parquet` | 118,777,457 | 119,028,242 | +250,785 |
| `g_hgics.parquet` | 263,610,039 | 263,681,349 | +71,310 |
| `comp_hgics.parquet` | 366,812,635 | 367,120,166 | +307,531 |
| `comp_ind.parquet` | 632,525,829 | 632,678,806 | +152,977 |

These extra rows propagate into `__msf_world2`, `__msf_world3`, `world_msf`, and the final `processed/characteristics/{country}.parquet` files, where they manifest as a different `gics`-null count between runs (e.g. 5,986,318 vs 5,978,348 in `world_data_output.parquet` — about 8,000 securities-months that gain a non-null `gics` only on the later-running side).

### Code to Reproduce

The bug manifests on any rerun where the second run extends past midnight from the first. Equivalently, calling `comp_hgics` twice in a single process while monkeypatching `datetime.date.today` to return different values produces different row counts in `na_hgics.parquet` / `g_hgics.parquet`.

### Date Code Was Run

2026-04-28 21:55 → 2026-04-29 04:17 (run 1) and 2026-04-28 21:55 → 2026-04-29 07:39 (run 2). Yale HPC.

### Additional Context

Surfaced while comparing two pipeline runs as part of PyPI-readiness work.

**Why this matters:** The pipeline isn't reproducible across runs unless they all begin and end on the same date — a fragile property that quietly breaks comparison testing and the planned golden-snapshot test (#41).

**Proposed fix:** Replace `date.today()` with `END_DATE` from `src/jkp/data/config.py`. `END_DATE` is already the documented end of the dataset; aligning the GICS panel's open-end fill to it keeps everything coherent and removes the wall-clock dependency.

**Adjacent occurrences in the same file** (these don't currently cause data divergence — they build lookup tables LEFT-JOINed against real data so excess rows are unused — but they should get the same `END_DATE`-based fix for consistency and clarity):

- `merge_roll_apply_daily_results` line 7977: `date_idx = datetime.datetime.today().month + datetime.datetime.today().year * 12`
- `gen_aux_maps` line 8384: `date_aux = datetime.datetime.today().month + datetime.datetime.today().year * 12`

A grep audit (`grep -rEn "date\.today\(\)|datetime\.now\(\)|datetime\.today\(\)" src/jkp/data/`) found no other call sites that affect output.

### Acceptance

- `comp_hgics` produces identical output for the same input regardless of when it's run.
- The two adjacent occurrences in `merge_roll_apply_daily_results` and `gen_aux_maps` get the same treatment for consistency.

FernandoRDLL:
## Summary

Fixes #131.

`comp_hgics` filled open-ended GICS `indthru` with `date.today()` and then exploded into a daily panel, so every day between two runs added ~250k rows to `na_hgics.parquet` and leaked non-determinism into the final country characteristic files. `merge_roll_apply_daily_results` and `gen_aux_maps` had the same wall-clock pattern in lookup-table date indices.

Replaced all three with `END_DATE` from `config.py` — the documented dataset endpoint. Output is now deterministic. Removed the now-unused `import datetime`.

## Change Type

## Methodological Impact

None. Only the fill value for open-ended GICS intervals changes (dataset endpoint instead of wall clock).

## Data Impact

`na_hgics.parquet` / `g_hgics.parquet` (and downstream `gics` values in country files) will differ from pre-fix runs that spanned more than one wall-clock day — those runs weren't reproducible to begin with. The other two sites feed LEFT-JOINed lookups, so no joined values change.

## Breaking Changes

None at the current `END_DATE`. Future bumps of `END_DATE` extend outputs — single source of truth instead of three wall-clock reads.

## Tests

All 360 unit tests pass. No new test added: one-line substitution per site, and the grep audit below covers the determinism property.

## Checklist

- [x] Tests pass locally (`pytest`)
- [x] Lint clean (`ruff check src/jkp/data/ tests/`)
- [ ] Full pipeline run — skipped (needs WRDS + ~450 GB RAM)
- [x] Docstring updated (`comp_hgics` step 2)
- [x] Considered factor calculation impact

## Additional Notes

\`\`\`
grep -rEn "date\.today\(\)|datetime\.now\(\)|datetime\.today\(\)" src/jkp/data/
\`\`\`
returns no hits.

### DIFF
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -7963,7 +7963,7 @@ def merge_roll_apply_daily_results():
         Merge rolling regression daily results into one dataset.
 
     Steps:
-        1) Build date index from earliest to current month.
+        1) Build date index from earliest to END_DATE month.
         2) Load id_int mapping and all '__roll*' parquet files (sorted for
            deterministic join order).
         3) Outer join them on (id_int, aux_date).
@@ -8373,7 +8373,7 @@ def gen_aux_maps(sfx):
 
     Steps:
         1) Map suffix to k: {'_21d':1,'_126d':6,'_252d':12,'_1260d':60} or int(sfx).
-        2) Build aux_date range from start index to current month index.
+        2) Build aux_date range from start index to END_DATE month index.
         3) Create grouped mappings via group_mapping_dfs(date_idx, k).
 
     Output:
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import datetime
 import functools
 import operator
 import os
@@ -21,7 +20,7 @@ import polars_ols  # noqa: F401 - required for least_squares method on polars ex
 from ibis import _
 from polars import col
 
-from .config import MAIN_FILTERS
+from .config import END_DATE, MAIN_FILTERS
 
 
 def fl_none():
@@ -2307,7 +2306,7 @@ def comp_hgics(lib):
 
     Steps:
         1) Load raw file (national/global); replace null gics with -999 sentinel.
-        2) Compute row counts and terminal rows; set open-ended indthru to (max(indfrom), or today).
+        2) Compute row counts and terminal rows; set open-ended indthru to (max(indfrom) or END_DATE).
         3) Create date ranges [indfrom, indthru]; explode; unique per (gvkey,date).
         4) Write to na_hgics.parquet or g_hgics.parquet.
 
@@ -2329,8 +2328,8 @@ def comp_hgics(lib):
     )
     indthru_date = (
         pl.lit(data[["indfrom"]].max()[0, 0])
-        if data[["indfrom"]].max()[0, 0] > date.today()
-        else pl.lit(date.today())
+        if data[["indfrom"]].max()[0, 0] > END_DATE
+        else pl.lit(END_DATE)
     )
     c1 = col("n") == col("n_aux")
     c2 = col("indthru").is_null()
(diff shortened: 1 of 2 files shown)


capellini (approved):
Added a test file, then ran the /review-pr Claude Code command and corrected a few small nits. Ready for merge.

PR #127, opened 2026-04-28
Add __version__ and CLI --version flag

capellini:
## Summary

Exposes the installed package version through `jkp.data.__version__` via `importlib.metadata`, and adds a `--version` flag to the top-level `jkp` CLI that prints the version and exits. Falls back to `0.0.0+unknown` when the package is imported without being installed (e.g. directly from a source tree).

The `--version` option is registered as eager so it short-circuits ahead of any subcommand parsing, even if future app-level options become required.

This is a Tier 0 prerequisite for publishing the package to PyPI: users and bug reports need a way to learn which version is installed.

Closes #96.

## Change Type

- [x] Infrastructure change (CI, config, dependencies)

## Methodological Impact

None.

## Data Impact

None.

## Breaking Changes

None.

## Tests

Added two tests under a new `TestVersionFlag` class in `tests/unit/test_cli.py`:

- `test_version_prints_package_version` — invokes `jkp --version`, asserts exit code 0 and that `__version__` appears in the output.
- `test_version_does_not_invoke_subcommand` — invokes `jkp --version build /tmp/whatever` with `run_pipeline` patched, asserts the subcommand is not invoked (i.e. eager short-circuit works).

All 16 CLI tests pass locally (14 existing + 2 new). Acceptance criteria from the issue both verified manually:

```
$ uv run jkp --version
0.1.0
$ uv run python -c "import jkp.data; print(jkp.data.__version__)"
0.1.0
```

## Checklist

- [x] I have run the tests locally (`pytest`)
- [x] I have run the linter locally (`ruff check src/jkp/data/ tests/`)
- [x] I have verified the pipeline runs successfully after my changes — `jkp --version`, `jkp --help`, and `jkp build --help` all behave correctly
- [x] I have updated documentation if needed
- [x] I have considered whether this change affects factor calculations — it does not

## Additional Notes

The `__init__.py` resolves the version at import time using `importlib.metadata.version("jkp-data")`. This keeps `pyproject.toml` as the single source of truth for the version string — there is no separate constant in code that has to be kept in sync.

Issue #92, opened 2026-04-19
Publish jkp-data to PyPI (v0.1.0)


capellini:
## Goal

Prepare `jkp-data` for its first public release on PyPI as version `0.1.0` ("public preview, API may evolve").

## Why

We want wider adoption and usage of the package. Professional polish is important to attract both users and potential maintainers. Publishing to PyPI is the friction-reduction step that unlocks external use — users can install with any package manager that supports PyPI (pip, uv, conda, etc.).

Publishing also lays the groundwork for a plug-in infrastructure: with `jkp.data` as a proper namespace package on PyPI, future extension packages (e.g., `jkp.api`, `jkp.website`) can depend on it and extend the CLI without forking the core.

## Scope

This is a tracking issue. Each sub-issue represents a concrete work item, triaged into three priority tiers:

- **Tier 0 (P0)** — true blockers for a professional public release
- **Tier 1 (P1)** — quality and security concerns a careful reviewer will notice; strongly recommended before v0.1.0
- **Tier 2 (P2)** — polish; can ship after v0.1.0 if time-pressed

## Data usage note

`jkp-data` is MIT-licensed orchestration code. It does **not** distribute WRDS, CRSP, Compustat, or IBES data. Users must bring their own WRDS subscription. This distinction must be clear in all user-facing documentation before publishing to PyPI.

PR #134, opened 2026-05-03
Define public API via __all__ with lazy module __getattr__

### LINKED ISSUE #102: [PyPI v0.1.0] T1.4: Define __all__ in __init__.py
## Problem

The package has no declared public import surface. `from jkp.data import *` exposes whatever happens to be imported. Documentation generators (Sphinx, pdoc) have no signal for what's API vs. internal.

## Solution

After T0.3 adds `__version__`, define `__all__` listing the explicit public entry points:
```python
__all__ = ["__version__", "app", "DataPaths"]  # expand as needed
```

Keep the list deliberately small — we're committing to these names across the v0.x line. The scope discussion (what counts as public) is the real work here.

## Files

- `src/jkp/data/__init__.py`

## Acceptance

- `python -c "from jkp.data import *; print(dir())"` shows the intended symbols

capellini:
> **⚠️ Stacked on top of #127** ("Add `__version__` and CLI `--version` flag"). The diff against `main` therefore includes `src/jkp/data/cli.py` and `tests/unit/test_cli.py` from #127 — those should be reviewed as part of #127, not here. The substantive changes in *this* PR are confined to:
> - `src/jkp/data/__init__.py` (the `__all__` declaration + lazy `__getattr__` + `__dir__`)
> - `tests/unit/test_public_api.py` (new 8-test file)
>
> Once #127 lands, this PR's diff will collapse to those two files. Please merge #127 first.

## Summary

Adds an explicit, minimal public-API surface for the v0.x line so star-imports and doc generators have a clear signal for what is supported, and so internals can change without breaking users. Resolves #102.

The five public names — `__version__`, `app`, `DataPaths`, `run_pipeline`, `run_portfolio` — are exposed lazily through a module-level `__getattr__` ([data-model hook for customizing module attribute access](https://docs.python.org/3/reference/datamodel.html#customizing-module-attribute-access)) rather than eagerly re-exported, so `import jkp.data` does not pull `polars`/`duckdb`/`ibis` and `jkp.data.__version__` stays cheap. `__dir__` is overridden to advertise *exactly* `__all__` — `dir(jkp.data)` no longer leaks `version`, `PackageNotFoundError`, `_LAZY_ATTRS`, or dunders into tab-completion.

The five names cover: the package version (Python convention); the Typer `app` (so plug-in packages can compose it via `add_typer`); `DataPaths` (extension packages need it to integrate with the directory layout); and the two top-level entry points (so notebook/script users don't have to go through the CLI). Internals (per-characteristic helpers in `aux_functions`, output writers, credential plumbing) remain private — adding to `__all__` is a one-way door for v0.x.

## Change Type

- [ ] Documentation only

(Packaging-surface change — declares the v0.x public API. Marked Infrastructure as the closest fit; this is part of the PyPI v0.1.0 work tracked under #92.)

## Methodological Impact

None. No factor definitions, calculations, or numerical outputs are touched.

## Data Impact

None. No characteristic, return, or portfolio output changes.

## Breaking Changes

None. This is purely additive: every name that was previously reachable on `jkp.data` is still reachable. The lazy `__getattr__` fires only on attribute access of the five `__all__` names, and submodule paths like `from jkp.data.aux_functions import ...` are unaffected.

## Tests

New `tests/unit/test_public_api.py` (8 tests):

- `TestPublicApiSurface` — `__all__` matches the documented set; every name resolves to a non-`None` object; `from jkp.data import *` exposes exactly those names (subprocess assertion); unknown attribute raises `AttributeError`; `dir(jkp.data)` is *exactly* the public API.
- `TestLazyImportBehavior` — fresh-subprocess negative tests proving `import jkp.data` does NOT load `polars`, and that `jkp.data.__version__` lookup does NOT load `polars`. An inverse sanity check confirms touching `jkp.data.run_pipeline` DOES load `polars` (catches a buggy resolver returning `None`).
- Subprocess tests strip `COV_*` env vars so pytest-cov's subprocess hook doesn't eagerly import the measured package and falsely fail the lazy-import tests.

## Checklist

- [x] I have run the tests locally (`uv run pytest tests/unit/ --no-cov` — 370/370 passing)
- [x] I have run the linter locally (`uv run ruff check src/jkp/data/ tests/` clean; `ruff format --check` clean)
- [ ] I have verified the pipeline runs successfully after my changes — not run end-to-end here; the change is import-surface only and does not alter any pipeline code path. Pipeline regression continues under the chdir-removal comparison runs on Yale.
- [x] I have updated documentation if needed — none required; the surface is self-documenting via `__all__`/`dir()`, and the inline comment in `__init__.py` explains the lazy-loading rationale.
- [x] I have considered whether this change affects factor calculations — it does not.

## Additional Notes

- Spec conformance with [Customizing module attribute access](https://docs.python.org/3/reference/datamodel.html#customizing-module-attribute-access): `__getattr__` returns the resolved value or raises `AttributeError(f"module {__name__!r} has no attribute {name!r}")`; `__dir__` returns a `list[str]`. Caching via `globals()[name] = value` after first resolution means subsequent lookups bypass `__getattr__` (Python only invokes it when normal attribute lookup fails).
- Submodule names (`jkp.data.cli`, `jkp.data.main`, …) may still surface in IPython/Jedi tab-completion via package-path introspection — that is Python's package machinery, not driven by `__dir__`.
- This is referenced from the v0.1.0 plan (`PYPI_PUBLICATION_PLAN.md` T1.4).

### DIFF
--- a/src/jkp/data/__init__.py
+++ b/src/jkp/data/__init__.py
@@ -1,8 +1,101 @@
 """JKP Factor Data generation pipeline."""
 
-from importlib.metadata import PackageNotFoundError, version
+# Use underscore-prefixed aliases so the `importlib.metadata` helpers don't
+# leak into `dir(jkp.data)` / tab-completion as if they were part of our API.
+from importlib.metadata import PackageNotFoundError as _PackageNotFoundError
+from importlib.metadata import version as _version
+from typing import Any as _Any  # underscore-aliased so it doesn't leak into dir()
 
+# IMPORTANT: `__version__` must be assigned BEFORE `_LAZY_ATTRS` / `__getattr__`
+# below. Submodules loaded by the lazy resolver (notably `.cli`) do
+# `from . import __version__` at import time; if that name isn't already in
+# this module's `__dict__` when the lazy import fires, those submodules will
+# observe a partially-initialized package and fail. Keep this assignment at
+# the top of the file.
 try:
-    __version__ = version("jkp-data")
-except PackageNotFoundError:  # pragma: no cover
+    __version__ = _version("jkp-data")
+except _PackageNotFoundError:  # pragma: no cover
     __version__ = "0.0.0+unknown"
+
+del _PackageNotFoundError, _version
+
+
+# Public API surface for the v0.x line. Adding a name here commits to keeping
+# it importable from `jkp.data` for the duration of v0.x — bumping these names
+# requires a major version bump.
+#
+# We deliberately keep this list small. The five names below cover:
+#   - the package version string (Python convention),
+#   - the Typer CLI app object (so future plug-in packages can compose it via
+#     `add_typer`),
+#   - the `DataPaths` dataclass (extension packages need it to integrate with
+#     the pipeline's directory layout),
+#   - the two top-level pipeline entry points (so users can drive the pipeline
+#     from notebooks / scripts without going through the CLI).
+#
+# Other internals (per-characteristic helpers in `aux_functions`, output
+# writers, credential plumbing) remain private until we have a clearer reason
+# to commit them as public — adding a name here is a one-way door for v0.x.
+__all__ = [
+    "__version__",
+    "app",
+    "DataPaths",
+    "run_pipeline",
+    "run_portfolio",
+]
+
+
+# We expose the names above via lazy module-level `__getattr__` (see Python's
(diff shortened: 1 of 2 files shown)


capellini:
Ready to merge after Claude-assisted review.

PR #139, opened 2026-05-04
Tighten dep pins to lockfile range and check lock in CI

### LINKED ISSUE #135: [Bug]: pyproject.toml mismatch with uv.lock — no safeguard catches silent major bumps
## Bug Description

`pyproject.toml` declares loose lower-bound version pins, while `uv.lock` contains the exact versions actually used by CI and contributors. The two are mismatched on `main`, with no safeguard catching the gap.

## Current state on `main` (1795a40)

`pyproject.toml`:
```
ibis-framework[duckdb,postgres]>=11.0
polars>=1.34.0
```

`uv.lock`:
```
ibis-framework  12.0.0
polars          1.40.1
```

The lock satisfies the pyproject ranges, so `uv` accepts the project. But pyproject no longer describes what is shipped or tested. CI uses `uv sync --frozen` (lockfile wins) so it stays green; anyone running plain `uv sync`, or any future `uv lock` regen, can resolve to a different version silently — including across major boundaries.

## How this got here

PR #114 (`d00529c`, "Replace polars-ds Dimson β with pure-Polars closed-form") regenerated `uv.lock` as a side-effect of dropping `polars-ds`. With loose pyproject lower bounds, the resolver picked latest-in-range and bumped ibis 11→12 and polars 1.34→1.40 invisibly. PR description didn't mention the bumps; no test exercised the new APIs; review didn't flag them.

## Environment (where the symptoms surfaced)

- OS: macOS 14 (Darwin 25.4.0); CI: Ubuntu (matrix per `ci.yml`)
- Python: 3.11.15 / 3.12 (matrix)
- ibis-framework: 12.0.0 (lock); pyproject `>=11.0`
- polars: 1.40.1 (lock); pyproject `>=1.34.0`

## Pre-submission checklist

- Searched existing issues — no duplicate
- Reproduced on latest `main` (`1795a40`)
- Setup followed per `CONTRIBUTING.md`

### LINKED ISSUE #141: ibis 12 dropped `ibis.duckdb` shortcut — 17 call sites in aux_functions.py break at runtime
## Summary

After the ibis 11 → 12 upgrade (locked in #114, pinned in #139), every call to `ibis.duckdb.connect(...)` raises `AttributeError: module 'ibis' has no attribute 'duckdb'`. 17 call sites in `src/jkp/data/aux_functions.py` are affected, and **8 unit tests fail** as a direct consequence.

## Reproduction

```bash
uv run python -c "import ibis; ibis.duckdb.connect()"
# AttributeError: module 'ibis' has no attribute 'duckdb'.
```

## Failing tests

```
tests/unit/test_aug_msf_v2.py::TestAugMsfV2::test_output_has_askhi_bidlo_columns
tests/unit/test_aug_msf_v2.py::TestAugMsfV2::test_askhi_bidlo_values_for_tr_rows
tests/unit/test_aug_msf_v2.py::TestAugMsfV2::test_askhi_bidlo_null_for_non_tr_rows
tests/unit/test_aug_msf_v2.py::TestAugMsfV2::test_idempotent_on_rerun
tests/unit/test_aug_msf_v2.py::TestAugMsfV2::test_preserves_original_columns
tests/unit/test_aux_functions.py::test_gen_crsp_sf_exposes_ticker_after_senames_join[m]
tests/unit/test_aux_functions.py::test_gen_crsp_sf_exposes_ticker_after_senames_join[d]
tests/unit/test_aux_functions.py::test_aug_msf_v2_writes_augmented_file_and_is_idempotent
```

All 8 fail with the same root cause.

## Root cause

`ibis-framework` 12 removed the implicit `ibis.duckdb` namespace accessor. The duckdb backend is still available, but it must be loaded explicitly:

```python
# Old (works in ibis 11.x, broken in 12.x):
con = ibis.duckdb.connect(threads=os.cpu_count())

# New (works in ibis 12.x):
from ibis.backends.duckdb import Backend
con = Backend().do_connect(threads=os.cpu_count())
```

Or, equivalently:

```python
import ibis.backends.duckdb  # registers the backend
con = ibis.duckdb.connect(...)  # now resolvable
```

## Affected call sites

`src/jkp/data/aux_functions.py` — 17 occurrences:

```
242, 287, 325, 380, 520, 890, 1112, 1223, 1296, ...
```

(complete list via `grep -n "ibis\.duckdb\.connect" src/jkp/data/aux_functions.py`)

## CI status

These 8 tests exercise the WRDS/CRSP DuckDB-backed paths (via `ibis`). They've been failing on `main` since the ibis upgrade landed; the failure was treated as pre-existing in subsequent PRs and never blocked merges. #139 pins the ibis major to 12 to prevent further drift, but does not touch the call sites.

## Acceptance criteria

- All 8 listed tests pass under ibis 12.x.
- `uv run pytest tests/unit/test_aug_msf_v2.py tests/unit/test_aux_functions.py` returns green.
- Pipeline-level smoke run (`jkp build` on a small synthetic dataset, if available) still produces output.

## Related

- #114 — bumped ibis 11 → 12 in the lockfile (the original drift)
- #139 — pinned the major; does not address the runtime breakage
- #135 — closed by #139

FernandoRDLL:
Closes #135
Closes #141

## Simple explanation

This PR fixes a quiet drift between `pyproject.toml` and `uv.lock`, and prevents it from happening again. Three things happen:

1. **Dep pins now match the lockfile.** Before this PR, pyproject said we accepted ibis `>=11.0` and polars `>=1.34.0`, but the lock had silently moved to ibis 12 and polars 1.40 during an unrelated PR. CI passed because it runs `uv sync --frozen` and trusts the lock, but pyproject no longer described what we actually shipped. Pins are now `ibis>=12,<13` and `polars>=1.40,<1.41`.

2. **CI now refuses to silently re-drift.** A `uv lock --check` step is added to the lint job. The next time pyproject and the lock disagree, CI fails and the diff shows up in PR review instead of slipping through.

3. **The 8 mysterious ibis test failures are explained and fixed.** They weren't a code bug. They were stale-venv fallout from the original ibis bump — a corrupted `duckdb-1.4.3.dist-info/` directory that broke `importlib.metadata.version("duckdb")`. After this PR's pinned versions and a fresh `uv sync`, all 8 tests pass.

### Files changed

- `pyproject.toml` — pinned `ibis-framework[duckdb,postgres]>=12.0,<13` and `polars>=1.40,<1.41`. The polars upper cap is intentionally tight: 1.40 already renamed `pl.PartitionByKey` to `pl.PartitionBy` mid-minor, so opt-in to 1.41 is preferable to surprise.
- `.github/workflows/ci.yml` — added `uv lock --check` step in the lint job.
- `uv.lock` — regenerated. Diff is just the new spec strings (plus two unrelated marker tidy-ups `uv` threw in). No version bumps.

### Why the drift happened

PR #114 was meant to swap out `polars-ds` for a closed-form Dimson β. As a side effect, `uv lock` re-resolved the world and walked ibis from 11 to 12 and polars from 1.34 to 1.40. Nothing in the PR description, the diff, or the test changes flagged it — the loose lower bounds in pyproject just let it happen, and lockfile diffs are noisy enough that nobody caught it in review.

Two things broke downstream as a result: a feature branch using the old `pl.PartitionByKey` name failed on rebase, and a contributor with a long-lived local `.venv` ended up with a corrupted ibis install that fooled `importlib.metadata` into reporting zero backends. Both were fallout from the same silent bump.

The fix is structural rather than reactive: make pyproject describe the major we actually support, and make CI complain the next time pyproject and the lock disagree. Bumping a major now becomes an explicit, reviewable decision instead of a side effect.

### Resolving the 8 ibis test failures (#141)

Eight tests in `test_aug_msf_v2.py` and `test_aux_functions.py` had been failing on stale venvs with `AttributeError: module 'ibis' has no attribute 'duckdb'`. Diagnosis:

- The ibis 11 → 12 dep bump left a stale `duckdb-1.4.3.dist-info/` directory alongside the new `duckdb-1.5.2.dist-info/` in long-lived venvs. `uv pip install --reinstall` couldn't repair it because `RECORD` was missing on the older entry; uv ended up renaming files inside it (`METADATA 2`, `RECORD 2`, …).
- With the corrupt dist-info present, `importlib.metadata.version("duckdb")` returned `None`. Ibis 12's duckdb backend calls that during `_post_connect` to gate a behavior change (`vparse(self.version) > vparse("1")`), so any `ibis.duckdb.connect(...)` raised before the connection was usable.
- Not a bug in `aux_functions.py` — a venv state problem caused by the silent dep bump.

After this PR, a clean `rm -rf .venv && uv sync --all-groups` resolves to ibis 12.0.0 + duckdb 1.5.2 (the locked majors), and all 8 previously-failing tests pass (`test_aug_msf_v2.py` 5/5; `test_aux_functions.py::test_gen_crsp_sf_exposes_ticker_after_senames_join` both `m`/`d`; `test_aux_functions.py::test_aug_msf_v2_writes_augmented_file_and_is_idempotent`).

The pin tightening + `uv lock --check` CI gate prevents this exact failure mode from recurring: the next ibis major bump can't slip through without an explicit pyproject change.

## Test plan

- [x] `uv sync --frozen --all-groups` on a clean venv resolves cleanly
- [x] `uv lock --check` reports lockfile in sync with the new pyproject
- [x] `uv run ruff check src/jkp/data/ tests/` passes
- [x] `uv run ruff format --check src/jkp/data/ tests/` passes
- [x] Full unit suite passes after clean venv (the 8 ibis tests now green)
- [ ] CI green on this PR


capellini:
Thank you for surfacing this issue @FernandoRDLL -- dependency management is a serious issue that I had been deferring until now. I appreciate all of your hard work on this -- it can help us hash out a very important nuance. And it was a great catch that we need to bump the versions in pyproject.toml.

I would be hesitant to commit to including any `<` restriction in pyproject.toml. Pinning to minor versions in particular would necessitate a regular update cadence that we may not want to sign up for.

We're treating "drift" between pyproject.toml and uv.lock as a problem here, but I'm not sure that it is (though uv.lock versions should be compatible with pyproject.toml versions). The pyproject.toml file is the authoritative source for versioning in the project, as it's recognized by all package managers and is what our users will use to install the latest versions of our dependencies, including upstream performance enhancements and bug fixes. Conversely, the uv.lock file is a research artifact for historical record-keeping.

I still think that we need both files due to the bespoke nature of this project -- it's not only a python package, but also a source of truth for data that we generate. So we need to support people using this package, both (1) with the latest enhancements and bug fixes and (2) to reproduce results on a specific date.

So the pyproject.toml file serves as the most current configuration and the uv.lock file serves as a historical reference of the specific locked versions of packages that we used on a given date to produce our specific output. These need not be in sync at all times and the pyproject.toml file should be more forgiving than the uv.lock file.

A more durable solution to prevent failing tests on major version bumps is to install and configure dependabot. I haven't done that yet because dependabot is most useful, and least annoying, when you have a test suite that is comprehensive enough to catch errors and allow auto merge and we're still working on that.

In this way, being less restrictive often helps, as it forces us to deal with breaking changes in small increments as they surface, as opposed to multiple version upgrades that often arise when pinning versions for lengthy periods of time.

That's not to say that we will never need to pin specific versions. Often, packages let bugs slip through that necessitate pinning to specific versions, but only temporarily. As a concrete example: does the ibis-framework bump actually necessitate a polars cap at all? The PR frames `<1.41` as preferring opt-in over surprise, but if opting in to 1.41 is the goal, should we just use `polars>=1.41`?

Though I can definitely understand the initial instinct to pin versions, in my experience, it's best to enforce as few restrictions as possible and, only if necessary, tighten those later (e.g. major versions, security patches, our code changes that necessitate upgrades). This allows us to focus our efforts on true value-added enhancements and bug fixes, as opposed to version bump maintenance patches.

We probably also want to include a separate github actions workflow to use `pip` as a test of that workflow, in addition to the `uv` workflow.

capellini:
Apologies for the long post, but dependency management is a huge topic and this repository very unique! I should probably detail my thoughts in a markdown document for historical purposes.

Of course, I'm also interested in everyone's opinions on the matter -- these are just my ideas based on my experiences. They aren't meant to be authoritative statements.

FernandoRDLL:
Hey @capellini , I softened the filter. Is this correct?

@theisij  Do you have any suggestions?

capellini:
Thanks for the changes @FernandoRDLL - this is better, for sure. I'm starting to think that we should formally separate the two concerns that I mentioned above. 

I'll post a separate issue with my thoughts. We need to eventually decide on what to do, but I don't want to hold up this PR while we decide. I'll review this PR and then we can hash out the details of how to properly package this in a separate issue.

capellini (approved):
This looks good to claude and me, pending a deeper discussion on project structure in #142.

PR #143, opened 2026-05-08
Merge branch 'main' of https://github.com/bkelly-lab/jkp-data

tomassamaj:
## Change Type

- [ ] Documentation only

## Checklist

- [ ] I have run the tests locally (`pytest`)
- [ ] I have run the linter locally (`ruff check src/jkp/data/ tests/`)
- [ ] I have updated documentation if needed
- [ ] I have considered whether this change affects factor calculations


FernandoRDLL:
Hello, this seems to be a more personal version of the repo rather than something we should merge. Is that correct?

PR #148, opened 2026-05-11
Use pds.lin_reg for Dimson β

### LINKED ISSUE #144: [Feature]: Revisit polars-ds for beta_dimson now that new release is out
## Feature Description

Swap the closed-form sufficient-statistics OLS used for Dimson β back to a `polars-ds` (`pds.lin_reg`) implementation, using the new release of polars-ds.

Current impl lives at `src/jkp/data/aux_functions.py:8809-8841` (`_dimson_exprs` + `dimsonbeta`), with the Cramer's-rule solver in `_solve_beta_sum_sym3` at `aux_functions.py:8784-8806`. It aggregates the 6 upper-triangle covariances of `(mktrf_lg1, mktrf, mktrf_ld1)` plus the 3 `Xᵀy` terms per `(id_int, group_number)`, then solves a 3×3 system in-expression.

## Use Case / Motivation

The closed-form path was introduced because both `pds.lin_reg` and `polars-ols` `mode=\"coefficients\"` were slow at ~10⁷-group cardinality in earlier polars-ds releases (per-group plugin dispatch + list/struct allocation dominated wall time). With the new polars-ds release, this overhead may be substantially reduced.

If `pds.lin_reg` is now competitive, switching back would:

- Replace ~60 lines of bespoke 3×3 OLS math + caching with a single `pds.lin_reg(...).list.head(3).list.sum()` expression.
- Remove a hand-rolled near-singularity check (`rcond > 1e-12`) in favor of the plugin's internal handling.
- Align Dimson β with the other regression-based stats and lower the maintenance surface.

## Proposed Implementation

1. Add `polars-ds` to `pyproject.toml` (currently not a dependency — only `polars-ols>=0.3.5` is pinned).
2. Replace `_dimson_exprs` / `_solve_beta_sum_sym3` / `dimsonbeta` body in `aux_functions.py` with a `pds.lin_reg` call returning the coefficient list, then `list.head(3).list.sum()` for β_sum.
3. Benchmark the new impl against the current closed-form version on a representative slice (e.g. 252d window, full CRSP universe) before merging — wall time and peak memory.
4. Confirm bit-for-bit (or within numerical tolerance) parity on `beta_dimson_21d` against current outputs via the golden test suite.

## Alternatives Considered

- Keep closed-form sufficient-stats OLS (status quo).
- Use `polars-ols` `mode=\"coefficients\"` — prior testing showed struct-field unpacking was worse than the polars-ds list path at this cardinality.

## Additional Context

Only proceed with the swap if benchmarks show polars-ds is competitive with or faster than the closed-form path. If the new release still loses at this scale, close as wontfix and keep current impl.

FernandoRDLL:
Closes #144

## Simple explanation

Dimson β used to be a one-line call to polars-ds. We wrote our own 3×3 OLS solver by hand — about fifty lines of Cramer's rule— because polars-ds 0.10.2 was too slow at our group counts. Polars-ds 0.11.2 has a speed up for that function (I made that), so we're going back. The new version is a single regression call. It's slower than the hand-rolled solver (about 6×), it will only add like 5 min to the pipeline but code is much easier to read.

### Implementation

`src/jkp/data/aux_functions.py`:

- Added `import polars_ds as pds`.
- Removed `_solve_beta_sum_sym3` (Cramer's-rule 3×3 solver) and `_dimson_exprs` (cached agg-expr builder).
- `dimsonbeta` reduced from ~58 lines (incl. helpers) to 25 lines:

  ```python
  def dimsonbeta(df, sfx, __min):
      name = f"beta_dimson{sfx}"
      beta_expr = pl.col("coeffs").list.head(3).list.sum()
      return (
          df.group_by(["id_int", "group_number"])
          .agg(coeffs=pds.lin_reg(
              "mktrf", "mktrf_ld1", "mktrf_lg1",
              target="ret_exc", add_bias=True, solver="cholesky",
          ))
          .select("id_int", "group_number", beta_expr.alias(name))
          .filter(pl.col(name).is_not_null() & pl.col(name).is_not_nan())
      )
  ```

### Solver choice

`pds.lin_reg` exposes three solvers: `svd`, `qr`, `cholesky`. Bench at 100k groups × 25 obs:

| solver | median wall | notes |
|---|---:|---|
| svd | 0.664 s | slowest; full rank-revealing |
| qr | 0.557 s | rank-revealing |
| cholesky | 0.554 s | fastest; SPD `XᵀX` only |

Cholesky picked. Three near-iid market-return regressors over 21 trading days → `XᵀX` is full-rank in practice. qr/cholesky give identical full-LS answer; svd differs only for rank-deficient designs that don't occur on real CRSP data.

### Dependency

`pyproject.toml`: added `polars-ds>=0.11.2`. Resolved to 0.11.2 in `uv.lock`.

### Tests

`tests/unit/test_rolling_daily_metrics.py::TestDimsonbeta` — 6/6 pass.

Renamed `test_dimsonbeta_drops_singular_design` → `test_dimsonbeta_handles_singular_design`. `pds.lin_reg` returns a finite min-norm solution on rank-deficient designs instead of null; the drop-on-singular guard was added alongside the closed-form rewrite in #114 and isn't a pre-existing contract.

### Benchmark

`scripts/benchmark_dimsonbeta.py` — synthetic data, 25 obs/group, median of 5 runs:

| n_groups | obs | closed | pds | ratio | max\|Δ\| |
|---:|---:|---:|---:|---:|---:|
| 1e3 | 25k | 1.5 ms | 5.9 ms | 3.95× | 9e-16 |
| 1e4 | 250k | 11 ms | 60 ms | 5.36× | 9e-16 |
| 1e5 | 2.5M | 88 ms | 0.58 s | 6.57× | 1e-15 |
| 1e6 | 25M | 0.92 s | 6.02 s | 6.52× | 2e-15 |

~6.5× slower at scale, absolute cost still small (≤10s at 1M groups), outputs match to machine epsilon. Worth it for the readability — one `pds.lin_reg` call beats a hand-rolled 3×3 Cramer's-rule solver any day.

## Test plan

- [x] \`uv run pytest tests/unit/test_rolling_daily_metrics.py::TestDimsonbeta\` — 6 passed
- [x] \`uv run pytest tests/unit/\` — 372 passed
- [x] \`uv run ruff check src/jkp/data/ tests/\`
- [x] \`uv run pyright src/jkp/data/aux_functions.py\` — same 39 pre-existing errors (no new)
- [ ] CI green

### DIFF
--- a/scripts/benchmark_dimsonbeta.py
+++ b/scripts/benchmark_dimsonbeta.py
@@ -44,7 +44,16 @@ def dimsonbeta_pds(df: pl.DataFrame) -> pl.DataFrame:
     beta_expr = pl.col("coeffs").list.head(3).list.sum()
     return (
         df.group_by(["id_int", "group_number"])
-        .agg(coeffs=pds.lin_reg("mktrf", "mktrf_ld1", "mktrf_lg1", target="ret_exc", add_bias=True))
+        .agg(
+            coeffs=pds.lin_reg(
+                "mktrf_lg1",
+                "mktrf",
+                "mktrf_ld1",
+                target="ret_exc",
+                add_bias=True,
+                solver="cholesky",
+            )
+        )
         .select("id_int", "group_number", beta_expr.alias(name))
         .filter(pl.col(name).is_not_null() & pl.col(name).is_not_nan())
     )
--- a/scripts/benchmark_dimsonbeta.py
+++ b/scripts/benchmark_dimsonbeta.py
@@ -1,0 +1,140 @@
+"""Benchmark Dimson β: polars-ds `pds.lin_reg` vs closed-form sufficient-stats OLS.
+
+Generates synthetic data matching the `dimsonbeta` input schema (~25 obs per
+(id_int, group_number) group), sweeps over group counts, and reports median
+wall-clock per impl plus max abs diff on β.
+
+Run: `uv run python scripts/benchmark_dimsonbeta.py`
+"""
+
+from __future__ import annotations
+
+import functools
+import statistics
+import time
+
+import numpy as np
+import polars as pl
+import polars_ds as pds
+
+
+def make_data(n_groups: int, obs_per_group: int = 25, seed: int = 0) -> pl.DataFrame:
+    rng = np.random.default_rng(seed)
+    n = n_groups * obs_per_group
+    mkt = rng.standard_normal(n)
+    mkt_ld = rng.standard_normal(n)
+    mkt_lg = rng.standard_normal(n)
+    eps = 0.1 * rng.standard_normal(n)
+    ret_exc = 0.3 * mkt + 0.2 * mkt_ld + 0.1 * mkt_lg + eps  # true β_sum = 0.6
+    return pl.DataFrame(
+        {
+            "id_int": np.repeat(np.arange(n_groups, dtype=np.int64), obs_per_group),
+            "group_number": np.zeros(n, dtype=np.int64),
+            "mktrf": mkt,
+            "mktrf_ld1": mkt_ld,
+            "mktrf_lg1": mkt_lg,
+            "ret_exc": ret_exc,
+        }
+    )
+
+
+# --- impl A: new (polars-ds) -------------------------------------------------
+def dimsonbeta_pds(df: pl.DataFrame) -> pl.DataFrame:
+    name = "beta"
+    beta_expr = pl.col("coeffs").list.head(3).list.sum()
+    return (
+        df.group_by(["id_int", "group_number"])
+        .agg(coeffs=pds.lin_reg("mktrf", "mktrf_ld1", "mktrf_lg1", target="ret_exc", add_bias=True))
+        .select("id_int", "group_number", beta_expr.alias(name))
+        .filter(pl.col(name).is_not_null() & pl.col(name).is_not_nan())
+    )
+
+
+# --- impl B: closed-form (current pre-revert impl, inlined) ------------------
+def _solve_beta_sum_sym3(c00, c01, c02, c11, c12, c22, v0, v1, v2):
+    col0 = (c00, c01, c02)
+    col1 = (c01, c11, c12)
+    col2 = (c02, c12, c22)
+    v = (v0, v1, v2)
+
+    def det(a, b, c):
+        return (
+            a[0] * (b[1] * c[2] - b[2] * c[1])
+            - b[0] * (a[1] * c[2] - a[2] * c[1])
+            + c[0] * (a[1] * b[2] - a[2] * b[1])
+        )
+
+    det_S = det(col0, col1, col2)
+    num = det(v, col1, col2) + det(col0, v, col2) + det(col0, col1, v)
+    rcond = det_S.abs() / (c00 * c11 * c22).abs()
+    return pl.when(rcond > 1e-12).then(num / det_S).otherwise(None)
+
+
+@functools.cache
+def _dimson_exprs():
+    X = ("mktrf_lg1", "mktrf", "mktrf_ld1")
+    y = "ret_exc"
+    pairs = [(X[i], X[j]) for i in range(3) for j in range(i, 3)] + [(x, y) for x in X]
+    agg = tuple(
+        (pl.var(a) if a == b else pl.cov(a, b)).alias(f"m{k}") for k, (a, b) in enumerate(pairs)
+    )
+    beta = _solve_beta_sum_sym3(*(pl.col(f"m{k}") for k in range(9)))
+    return agg, beta
+
+
+def dimsonbeta_closed(df: pl.DataFrame) -> pl.DataFrame:
+    name = "beta"
+    agg, beta = _dimson_exprs()
+    return (
+        df.group_by(["id_int", "group_number"])
+        .agg(*agg)
+        .select("id_int", "group_number", beta.alias(name))
+        .filter(pl.col(name).is_not_null())
+    )
+
+
+# --- bench runner ------------------------------------------------------------
+def time_runs(fn, df, n_runs: int = 5) -> float:
+    fn(df)  # warm-up
+    samples = []
+    for _ in range(n_runs):
+        t0 = time.perf_counter()
+        fn(df)
+        samples.append(time.perf_counter() - t0)
+    return statistics.median(samples)
+
+
+def max_abs_diff(a: pl.DataFrame, b: pl.DataFrame) -> float:
+    joined = a.rename({"beta": "beta_a"}).join(
+        b.rename({"beta": "beta_b"}), on=["id_int", "group_number"], how="inner"
+    )
+    return joined.select((pl.col("beta_a") - pl.col("beta_b")).abs().max()).item()
+
+
+def main() -> None:
+    rows = []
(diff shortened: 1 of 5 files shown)


src/jkp/data/aux_functions.py
```diff
@@ -8781,61 +8782,30 @@ def mktcorr(df, sfx, __min):
...
-        .select("id_int", "group_number", beta.alias(name))
-        .filter(pl.col(name).is_not_null())
+        .agg(
+            coeffs=pds.lin_reg(
+                "mktrf",
+                "mktrf_ld1",
+                "mktrf_lg1",
+                target="ret_exc",
+                add_bias=True,
+                solver="cholesky",
```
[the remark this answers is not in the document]
FernandoRDLL: Keeping `solver="cholesky"` intentionally — it's the fast path here and the production input is full-rank in normal conditions. `test_dimsonbeta_handles_singular_design` (e6f1686) covers the near-singular case: polars-ds's Cholesky tolerates the near-singular X'X produced by collinear regressors via tiny numerical noise on the diagonal and still returns a finite β. The test locks that behavior in so any future solver swap that silently regresses callers will be caught.

src/jkp/data/aux_functions.py
```diff
@@ -8781,61 +8782,30 @@ def mktcorr(df, sfx, __min):
...
     return (
         df.group_by(["id_int", "group_number"])
-        .agg(*agg)
-        .select("id_int", "group_number", beta.alias(name))
-        .filter(pl.col(name).is_not_null())
+        .agg(
+            coeffs=pds.lin_reg(
+                "mktrf",
+                "mktrf_ld1",
+                "mktrf_lg1",
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in a82a952 — reordered the args to `pds.lin_reg("mktrf_lg1", "mktrf", "mktrf_ld1", ...)` so the call matches the docstring's mktrf_{-1,0,+1} convention. Sum-of-coefficients is invariant to order, but the symmetric ordering is easier to read against the docstring.

tests/unit/test_rolling_daily_metrics.py
```diff
@@ -1217,22 +1217,24 @@ def test_dimsonbeta_empty_input_returns_empty(self):
         result = dimsonbeta(df, "_21d", __min=15)
         assert len(result) == 0
 
-    def test_dimsonbeta_drops_singular_design(self):
-        """Singular X'X (perfectly collinear regressors) yields a null beta and is dropped."""
+    def test_dimsonbeta_handles_singular_design(self):
+        # pds.lin_reg uses QR (rank-revealing) and returns a finite min-norm
+        # solution when regressors are collinear, rather than a null beta.
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in e6f1686 — replaced the inaccurate "uses QR (rank-revealing)" comment with a regression-rationale note. The test still passes under `solver="cholesky"` (polars-ds tolerates the near-singular X'X in this fixture via tiny numerical noise on the diagonal) and locks the finite-output behavior so any future solver swap that silently regresses callers will be caught.

scripts/benchmark_dimsonbeta.py
```diff
@@ -0,0 +1,140 @@
...
+    )
+
+
+# --- impl A: new (polars-ds) -------------------------------------------------
+def dimsonbeta_pds(df: pl.DataFrame) -> pl.DataFrame:
+    name = "beta"
+    beta_expr = pl.col("coeffs").list.head(3).list.sum()
+    return (
+        df.group_by(["id_int", "group_number"])
+        .agg(coeffs=pds.lin_reg("mktrf", "mktrf_ld1", "mktrf_lg1", target="ret_exc", add_bias=True))
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in e6f1686 — pinned the benchmark's `dimsonbeta_pds` to the same `solver="cholesky"`, `add_bias=True`, and arg order (`mktrf_lg1, mktrf, mktrf_ld1`) as production, so benchmark numbers reflect actual pipeline behavior.

FernandoRDLL:
@theisij Ready to go

tests/unit/test_rolling_daily_metrics.py
```diff
@@ -1217,22 +1217,27 @@ def test_dimsonbeta_empty_input_returns_empty(self):
         result = dimsonbeta(df, "_21d", __min=15)
         assert len(result) == 0
 
-    def test_dimsonbeta_drops_singular_design(self):
-        """Singular X'X (perfectly collinear regressors) yields a null beta and is dropped."""
+    def test_dimsonbeta_handles_singular_design(self):
+        # Regression check: on a collinear design, dimsonbeta returns a finite
```
theisij, line 1221: I don't fully understand this, so I'm just working off the function description. It sounds like you want the function to return a finite number if the inputs are collinear. Wouldn't it make more sense to return an error?

theisij (commented):
I have one tiny comment about the test

tests/unit/test_rolling_daily_metrics.py
FernandoRDLL: Fair question. I think there are two options:

1. **Raise on collinearity.** Cleanest semantically, but the call site runs across ~thousands of (id, eom) groups per day for the full panel. A single degenerate 21d window (low-variance period, illiquid stock, etc.) would crash the whole `roll_apply_daily` step. Too risky.
2. **Return null → drop.** This is what the implementation did pre-#148 (closed-form Cramer's rule returned NaN when `det(X'X)≈0`, the `is_not_null & is_not_nan` filter then dropped the row). Semantically equivalent to "no estimate available," consistent with other missing-data paths in the pipeline, and `roll_apply_daily` downstream already tolerates missing observations.

I'd push for option 2. What do you think?

theisij: I would be in favor of returning null or NA (so I guess option 2). But just to be clear, that's not what you're currently doing right?

FernandoRDLL: Correct — right now it does not return null. With `solver="cholesky"` a collinear/degenerate window comes back as a finite min-norm β (the `is_not_null & is_not_nan` filter never fires), so as it stands it always returns a number. The current test asserts that finite return, which is the opposite of what you want.

I looked into getting option 2 properly: none of the polars-ds solvers (cholesky/qr/svd) return null on a singular design — they all hand back a finite min-norm solution. The clean fix would be upstream: I could open a PR on polars-ds to make `lin_reg` return null on a rank-deficient design (I already contributed the 0.11.2 speedup, so I know that code path). Then we'd get the null semantics for free with no local workaround.

That said — in real CRSP data these degenerate windows basically never happen. A perfectly collinear or near-constant 21-day window of lead/current/lag market returns doesn't occur in practice (and the upstream `__min` obs requirement filters the thin cases). So this is really a theoretical edge case, not something that affects production output. The synthetic test is the only place it shows up.

So I'd suggest: keep cholesky as-is for now, and I'll open the polars-ds PR to add proper null-on-singular handling so it's correct at the source.

theisij: Okay, that works. I agree that it's unlikely to happen in practice. Thanks

theisij (approved):

PR #151, opened 2026-05-11
Add unit tests for scaling/ratio helpers (#149)

### LINKED ISSUE #149: Add unit tests for scaling/ratio functions in aux_functions.py
## Feature Description

Add unit tests for scaling and ratio helper functions in `src/jkp/data/aux_functions.py` that currently lack test coverage.

Target functions (lines 4866–5119):

- `scale_me` (4866)
- `scale_mev` (4884)
- `mean_year` (4905)
- `temp_liq_rat` (4930)
- `temp_rat_other` (4950)
- `temp_rat_other_spc` (4968)
- `update_ni_inc_and_decrease` (5050)
- `calculate_consecutive_earnings_increases` (5078)
- `compute_capex_abn` (5119)

## Use Case / Motivation

These functions feed accounting characteristic calculations downstream. No direct unit coverage today — regressions only caught via golden/integration runs, which are slow and obscure root cause. Direct tests would:

- Pin numerical behavior (null handling, edge cases, ratio denominators).
- Catch regressions at PR time instead of full-pipeline run.
- Document expected semantics for new contributors.

## Proposed Implementation

- Add tests to `tests/unit/` (new file `test_scaling_ratios.py` or extend existing module).
- Use `make_dataframe` fixture and `tolerance` / `assert_series_equal` per `tests/README.md`.
- Verify: standard cases, all-null inputs, zero denominators (should not raise — `safe_div` semantics), single-row groups for `mean_year` / `calculate_consecutive_earnings_increases`.
- Cross-check `temp_*` helpers against a hand-computed small example.

## Willingness to Contribute

Yes, I can implement this myself and submit a PR.

## Pre-submission Checklist

- [x] Searched existing issues
- [x] Confirmed feature does not already exist

FernandoRDLL:
Closes #149

## Simple explanation

Nine scaling and ratio helper functions in aux_functions.py previously had no direct unit test coverage.

As a result, regressions in this area only surfaced during golden or integration runs, which are slower and make it harder to trace failures back to the underlying issue.

This PR adds 89 focused unit tests in a new dedicated test file so that future regressions fail immediately during PR validation, with clear test names like test_zero_sale_returns_null instead of downstream parquet diffs several layers later.

The tests cover behavior that is especially easy to break accidentally, including:

* null propagation,
* count-threshold gating,
* zero or negative denominators,
* FX scaling behavior,
* alias parameterization,
* cleanup of intermediate columns, and
* per-firm partition semantics.

No production code was changed in this PR; it only adds test coverage.

### Coverage

`tests/unit/test_scaling_ratios.py` — one `TestX` class per function, mirroring `tests/unit/test_accounting_formulas.py`:

| Function | Lines | Tests |
|---|---:|---:|
| `scale_me` | 4866–4882 | 15 |
| `scale_mev` | 4885–4902 | 10 |
| `mean_year` | 4905–4927 | 9 |
| `temp_liq_rat` | 4930–4946 | 8 |
| `temp_rat_other` | 4949–4965 | 7 |
| `temp_rat_other_spc` | 4968–4985 | 7 |
| `update_ni_inc_and_decrease` | 5050–5073 | 8 |
| `calculate_consecutive_earnings_increases` | 5076–5116 | 12 |
| `compute_capex_abn` | 5119–5152 | 12 |
| **Total** | | **89** |

### Behavior pinned

- `scale_me` / `scale_mev`: zero `me_company`/`mev` → null; `fx == 0` → 0 (not null); negative denominator flows through (no sign guard); `.replace("_x", "")` is a literal substring replace (`xrd_x` → `xrd_me`).
- `mean_year`: 3-stage fallback (avg → current → lag → null), partitioned by `gvkey`/`curcd`.
- `temp_liq_rat` / `temp_rat_other`: count gate is strict `>12`; count=12 returns null, count=13 passes.
- `temp_rat_other_spc`: numerator `(cogs + invt − invt.shift(12)).over(['gvkey','curcd'])`; null `cogs` or null `invt` lag propagates.
- `compute_capex_abn`: count gate strict `>36`; uses `safe_div` mode 3, so non-positive `sale_x` yields null upstream; zero `sale_x` at any of rows {current, t-12, t-24, t-36} returns null.
- `calculate_consecutive_earnings_increases`: count gate `>=33` is **necessary but not sufficient** — with monotone data starting at index 0, the earliest row where all 8 lag positions (0, 3, …, 21) have non-null `ni_inc` is row 33 (count=34). Row 32 (count=33) returns null because `shift(21)` lands on row 11 which has no 12-month lag. Pinned by `test_count_gate_boundary`.

### Observed quirk

`update_ni_inc_and_decrease` calls `.shift(lag)` without `.over(['gvkey','curcd'])`, so the shift is global to the sorted DataFrame. First row of firm B reads last row of firm A. `test_multi_firm_shift_is_global` asserts this so any future `.over(...)` refactor is intentional rather than incidental. Not fixed in this PR — out of scope for a tests-only change.

### Helpers / fixtures

- Local `_monthly_dates`, `_single_firm_panel`, `_ni_panel`, `_streak_frame` builders.
- Reuses `tolerance` fixture from `tests/conftest.py` — `TIGHT` for single-op `scale_*`, `FINANCIAL_RATIOS` (`STANDARD`) for ratio asserts.

## Test plan

- [x] `uv run python -m pytest tests/unit/test_scaling_ratios.py --no-cov` — 89 passed (~0.15 s)
- [x] `uv run python -m pytest tests/unit/ --no-cov` — 449 passed
- [x] `uv run ruff check tests/unit/test_scaling_ratios.py` — clean
- [x] `uv run ruff format --check tests/unit/test_scaling_ratios.py` — formatted
- [ ] CI green

### DIFF
--- a/tests/unit/test_scaling_ratios.py
+++ b/tests/unit/test_scaling_ratios.py
@@ -16,7 +16,6 @@ from datetime import date
 
 import numpy as np
 import polars as pl
-import pytest
 
 from jkp.data.aux_functions import (
     calculate_consecutive_earnings_increases,
@@ -36,7 +35,7 @@ from jkp.data.aux_functions import (
 
 
 def _monthly_dates(n: int, start_year: int = 2010) -> list[date]:
-    """Generate n month-end dates starting Jan of start_year."""
+    """Generate n monthly dates on the 28th starting Jan of start_year."""
     out: list[date] = []
     for i in range(n):
         year = start_year + i // 12
@@ -56,24 +55,6 @@ def _single_firm_panel(n: int, **cols: list) -> pl.DataFrame:
     for k, v in cols.items():
         data[k] = v
     return pl.DataFrame(data)
-
-
-@pytest.fixture
-def panel_13() -> pl.DataFrame:
-    """13-row single-firm panel (minimum length for shift(12))."""
-    return _single_firm_panel(13)
-
-
-@pytest.fixture
-def panel_24() -> pl.DataFrame:
-    """24-row single-firm panel."""
-    return _single_firm_panel(24)
-
-
-@pytest.fixture
-def panel_37() -> pl.DataFrame:
-    """37-row single-firm panel (minimum length for capex_abn count>36 gate)."""
-    return _single_firm_panel(37)
 
 
 # =============================================================================
--- a/tests/unit/test_scaling_ratios.py
+++ b/tests/unit/test_scaling_ratios.py
@@ -1,0 +1,832 @@
+"""
+Tests for scaling and ratio helper functions in aux_functions.py.
+
+Covers FX-adjusted scaling (scale_me, scale_mev), year-mean helpers
+(mean_year), parameterized ratio builders (temp_liq_rat, temp_rat_other,
+temp_rat_other_spc), and consecutive-earnings / abnormal-capex DataFrame
+transforms. Pins null-handling, count-gate, and zero-denominator behavior
+so downstream accounting characteristics cannot silently regress.
+
+Paper Reference: Jensen, Kelly, Pedersen (2023), "Is There a Replication Crisis in Finance?"
+"""
+
+from __future__ import annotations
+
+from datetime import date
+
+import numpy as np
+import polars as pl
+import pytest
+
+from jkp.data.aux_functions import (
+    calculate_consecutive_earnings_increases,
+    compute_capex_abn,
+    mean_year,
+    scale_me,
+    scale_mev,
+    temp_liq_rat,
+    temp_rat_other,
+    temp_rat_other_spc,
+    update_ni_inc_and_decrease,
+)
+
+# =============================================================================
+# Local Helpers / Fixtures
+# =============================================================================
+
+
+def _monthly_dates(n: int, start_year: int = 2010) -> list[date]:
+    """Generate n month-end dates starting Jan of start_year."""
+    out: list[date] = []
+    for i in range(n):
+        year = start_year + i // 12
+        month = (i % 12) + 1
+        out.append(date(year, month, 28))
+    return out
+
+
+def _single_firm_panel(n: int, **cols: list) -> pl.DataFrame:
+    """Build a sorted monthly panel for one gvkey/curcd with arbitrary columns."""
+    data: dict[str, list] = {
+        "gvkey": [1] * n,
+        "curcd": ["USD"] * n,
+        "datadate": _monthly_dates(n),
+        "count": list(range(1, n + 1)),
+    }
+    for k, v in cols.items():
+        data[k] = v
+    return pl.DataFrame(data)
+
+
+@pytest.fixture
+def panel_13() -> pl.DataFrame:
+    """13-row single-firm panel (minimum length for shift(12))."""
+    return _single_firm_panel(13)
+
+
+@pytest.fixture
+def panel_24() -> pl.DataFrame:
+    """24-row single-firm panel."""
+    return _single_firm_panel(24)
+
+
+@pytest.fixture
+def panel_37() -> pl.DataFrame:
+    """37-row single-firm panel (minimum length for capex_abn count>36 gate)."""
+    return _single_firm_panel(37)
+
+
+# =============================================================================
+# scale_me
+# =============================================================================
+
+
+class TestScaleMe:
+    """Tests for scale_me() — FX-adjusted scaling by me_company."""
+
+    def test_basic_positive(self, tolerance):
+        df = pl.DataFrame({"ni_x": [100.0], "fx": [1.0], "me_company": [50.0]})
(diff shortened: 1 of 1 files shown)


tests/unit/test_scaling_ratios.py
```diff
@@ -0,0 +1,832 @@
...
+
+
+def _monthly_dates(n: int, start_year: int = 2010) -> list[date]:
+    """Generate n month-end dates starting Jan of start_year."""
+    out: list[date] = []
+    for i in range(n):
+        year = start_year + i // 12
+        month = (i % 12) + 1
+        out.append(date(year, month, 28))
+    return out
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in c86a73f: docstring now says "monthly dates on the 28th" — matches actual day argument.

tests/unit/test_scaling_ratios.py
```diff
@@ -0,0 +1,832 @@
...
+    """24-row single-firm panel."""
+    return _single_firm_panel(24)
+
+
+@pytest.fixture
+def panel_37() -> pl.DataFrame:
+    """37-row single-firm panel (minimum length for capex_abn count>36 gate)."""
+    return _single_firm_panel(37)
+
+
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in c86a73f: removed panel_13/panel_24/panel_37 fixtures (and the now-unused pytest import). Dead code — tests construct panels inline via _single_firm_panel.

FernandoRDLL:
@theisij Ready to go

tests/unit/test_scaling_ratios.py
```diff
@@ -0,0 +1,813 @@
...
+        np.testing.assert_allclose(df.select(scale_me("ni_x"))["ni_me"][0], 5.0, **tolerance.TIGHT)
+
+    def test_fx_greater_than_one(self, tolerance):
+        df = pl.DataFrame({"ni_x": [100.0], "fx": [1.25], "me_company": [50.0]})
+        np.testing.assert_allclose(df.select(scale_me("ni_x"))["ni_me"][0], 2.5, **tolerance.TIGHT)
+
+    def test_fx_zero_returns_zero(self, tolerance):
+        """fx=0 is not guarded — numerator becomes 0, result is 0 (not null)."""
+        df = pl.DataFrame({"ni_x": [100.0], "fx": [0.0], "me_company": [50.0]})
+        np.testing.assert_allclose(df.select(scale_me("ni_x"))["ni_me"][0], 0.0, **tolerance.TIGHT)
```
theisij, line 119: Can you just explain what's happening here? It sounds like whenever there's a zero FX entry, the result becomes zero. That sounds strange. It seems like dividing by zero should result in an error. Can you explain why it's not?

theisij: I actually have a bunch of questions about the scale_me function more generally:
1. Why do we need a dedicated function for scaled by "me" (or any other numerator for that matter). Can't we just use the built-in division function?
2. Why is fx part of the scale_me function? It feels like "scale_me" should do one thing: scale a number by me. But with fx being part of the function, it sounds like scale_me is also doing fx conversion
There may be good answers to these questions, but it would be good to hear about your thought process

FernandoRDLL:
Good question — there's no division by zero here. `fx` is in the numerator, not the denominator. The formula is `(var * fx) / me_company`, and `fx` is the exchange rate to USD (built in `compustat_fx`, see line 410). So if `fx = 0` the numerator just becomes 0 and you get 0 back. The only thing that could divide by zero is `me_company`, and that's guarded — `me_company == 0` returns null via `fl_none()`.

**1. Why a dedicated function instead of plain division?**

Mainly to avoid repeating the same pattern everywhere. Each `*_me` characteristic needs the FX conversion, the `me_company != 0` guard, and the `xrd_x` → `xrd_me` renaming. Putting it in one helper means adding a new scaled variable is a single call instead of rewriting the whole expression, and it's much easier to audit since the guard and naming logic live in one place. Plain `/` would lose both the guard and the naming.

**2. Why is fx inside scale_me?**

It's a currency thing. `var` is in the firm's local currency (`curcd`) and `me_company` is in USD, so dividing them directly would mix currencies. `fx` converts the numerator to USD first so the units line up. You're right that it's doing two things at once — this is how the original SAS code was structured, with the FX conversion folded into the scaling step, so I kept it that way for parity. We could open an issue later to do all the FX conversion up front (convert everything to USD once, then scale) and take `fx` out of these helpers. Happy to file that if you want.

tests/unit/test_scaling_ratios.py
FernandoRDLL: Please look at the comment below.

theisij (approved):

PR #171, opened 2026-05-21
Build FF3/FF5/UMD/HXZ4/Mispricing factor returns + characteristics via gen_*_data and ap_factor_model_data

### LINKED ISSUE #157: [Feature]: Build FF3 monthly factor returns
We need a faithful build of the monthly FF3 factors (SMB and HML) following Fama-French's original construction as documented in their 1993 JFE paper and the Ken French data library.

The build should respect the canonical conventions: NYSE/AMEX/NASDAQ ordinary commons, annual June reconstitution, NYSE breakpoints for both the size and B/M sorts, 2×3 independent sort, value-weighting, and a July-to-June holding period.

The output should land in `ff_factors_monthly.parquet` with the standard schema (eom + excntry + factor columns).

### LINKED ISSUE #158: [Feature]: Build FF3 daily factor returns
Daily counterpart of #157. We need the FF3 daily factors built to match Ken French's daily series.

The portfolios should be formed once a year at the June reconstitution, and daily value-weighted returns computed within the held portfolios. Between rebalances the weights drift naturally with the daily returns, following the Ken French daily convention.

Output should be `ff_factors_daily.parquet`.

### LINKED ISSUE #159: [Feature]: Build FF5 monthly factor returns
We need a faithful build of the FF5 monthly factors — RMW (profitability) and CMA (investment), plus an FF5-style SMB — following Fama-French (2015, JFE).

The construction follows the same conventions as FF3 (USA universe, NYSE breakpoints, annual June reconstitution, 2×3 independent sorts, value-weighting, July-to-June holding) but with operating profitability and asset growth as the sort signals. The FF5 version of SMB averages across the nine portfolios from the B/M, OP, and INV sorts, rather than the three from the B/M sort alone.

Operating profitability and asset growth should follow the definitions in the 2015 paper — no jkp-specific substitutes for OP.

Output cols in `ff_factors_monthly.parquet` (alongside FF3 cols).

### LINKED ISSUE #160: [Feature]: Build FF5 daily factor returns
Daily counterpart of #159. Same construction as the monthly FF5 build, but with daily value-weighted returns within the held portfolios and weights drifting between June rebalances per the Ken French daily convention.

Output goes into `ff_factors_daily.parquet` alongside the FF3 cols.

### LINKED ISSUE #161: [Feature]: Build UMD (momentum) monthly factor returns
We need a faithful build of the Carhart momentum factor (UMD / MOM / WML) following Jegadeesh-Titman (1993) and Carhart (1997).

Unlike the FF factors this one rebalances monthly, not annually. At the end of each month stocks are sorted on the 12-2 cumulative return (cumulative return from twelve months ago through two months ago, skipping the most recent month to avoid short-term reversal). The 2×3 sort uses NYSE median for size and NYSE 30/70 for prior return, independently. The factor is the average of the two winner portfolios minus the average of the two loser portfolios, value-weighted using end-of-month ME for the next month's return.

Output col in `ff_factors_monthly.parquet`.

### LINKED ISSUE #162: [Feature]: Build UMD (momentum) daily factor returns
Daily counterpart of #161. Same monthly rebalancing of the 2×3 size × 12-2 prior-return sort, but with daily value-weighted returns within the held portfolios.

Output col in `ff_factors_daily.parquet`.

### LINKED ISSUE #163: [Feature]: Build HXZ4 monthly factor returns
We need a faithful build of the Hou-Xue-Zhang q-factor monthly returns (size, investment, profitability) following HXZ (2015, RFS).

The construction uses a 2×3×3 independent sort on size × investment × profitability, with NYSE breakpoints for the USA. Size and investment sorts rebalance annually at June; the profitability bucket updates monthly as new quarterly earnings data lands (so ROE responds to the most recent quarterly release).

Output: `hxz_factors_monthly.parquet` with size, investment, and profitability factor columns.

### LINKED ISSUE #164: [Feature]: Build HXZ4 daily factor returns
Daily counterpart of #163. The portfolio assignments (size × investment × profitability bucket) are held within the calendar month, and daily value-weighted returns are computed within the held buckets. Weights drift naturally between rebalances.

Output: `hxz_factors_daily.parquet`.

### LINKED ISSUE #165: [Feature]: Build Stambaugh-Yuan mispricing monthly factor returns
We need a faithful build of the Stambaugh-Yuan (2017, JF) mispricing factors — MGMT, PERF, and the mispricing-specific SMB — for the USA.

The construction relies on 11 anomalies, percentile-ranked cross-sectionally each month. The MGMT leg averages over the management anomalies (asset growth, accruals, composite issue, NOA, asset growth, stock issue); the PERF leg averages over the performance anomalies (distress, O-score, ROA, momentum, gross profitability). Stocks need at least three active anomaly pcts per leg.

Portfolios are formed via double-sort: size (NYSE median) × score (20/80 percentile breaks). MGMT and PERF factors take the low-minus-high spread per size bucket averaged across sizes. The mispricing SMB is small minus big within the middle-score bucket shared by both legs.

The benchmark is the SAS v1 build from MisprProject, which itself is a faithful replication of the SY 2017 paper using SIZ-era CRSP.

Output: `mp_factors_monthly.parquet`.

### LINKED ISSUE #166: [Feature]: Build Stambaugh-Yuan mispricing daily factor returns
Daily counterpart of #165. Same monthly portfolio formation as the monthly mispricing build, with daily value-weighted returns within the held buckets and assignments held for the calendar month.

We also want the world branch — per-excntry portfolios using per-country size medians (no NYSE breakpoint for ROW).

Output: `mp_factors_daily.parquet` for both USA and the rest of the world.

### LINKED ISSUE #167: [Feature]: Build FF3 stock-level characteristics
We need the stock-level FF3 characteristics — market equity (ME) and book-to-market (BE/ME) — built per Fama-French (1993, JFE) variable definitions.

ME is end-of-June market equity used for the size sort and stamped to the July-through-June holding window. BE/ME pairs fiscal-year book equity (year ending in the prior calendar year) with end-of-December market equity. Book equity follows the Davis-Fama-French waterfall.

The id should be Int64 BIGINT, matching the jkp `world_msf.id` convention (raw permno for USA, encoded gvkey-iid for ROW).

Output: `ff_characteristics.parquet`.

### LINKED ISSUE #168: [Feature]: Build FF5 stock-level characteristics
We need the additional FF5 characteristics — operating profitability (OP) and investment (asset growth, INV) — built per Fama-French (2015, JFE) variable definitions.

Both come from annual Compustat data at fiscal year ending in the prior calendar year, stamped to the June formation date.

These columns sit alongside the FF3 chars in `ff_characteristics.parquet`. The OP definition should follow the 2015 paper directly, not jkp-specific substitutes.

### LINKED ISSUE #169: [Feature]: Build HXZ4 stock-level characteristics
We need the stock-level HXZ characteristics — market equity (ME), investment-to-assets (IA), and quarterly return on equity (ROE) — built per Hou-Xue-Zhang (2015, RFS).

ME and IA are pulled from annual data and stamped to the June formation date. ROE is the key quarterly characteristic — it pairs quarterly net income with lagged book equity from the quarterly Compustat file, and it updates each month as new quarterly earnings are released (via the announcement date `rdq`, not `datadate`). Book equity follows the Davis-Fama-French waterfall.

Output: `hxz_characteristics.parquet`.

### LINKED ISSUE #170: [Feature]: Build Stambaugh-Yuan mispricing stock-level characteristics
We need the per-stock mispricing scores — MGMT and PERF — built per Stambaugh-Yuan (2017, JF).

Each score is the average of the percentile rank (1-100) across that leg's anomalies, computed cross-sectionally each month-end on the USA universe. Positive anomalies (those where larger values indicate overpricing rather than underpricing) are ranked in descending order so that higher percentiles consistently mean more overpriced. Stocks need at least three active anomaly pcts per leg, otherwise the score is null.

The benchmark is the per-stock mispricing scores from the SAS v1 build (MisprProject).

Output: `mp_characteristics.parquet`.

### LINKED ISSUE #178: Add documentation for replication release
## Summary

Release documentation for the new factor replications (FF3/FF5, UMD, HXZ4, Stambaugh-Yuan mispricing, and DHS) shipped under `documentation/ap_factor_replication/`.

## Status

Delivered in PR #171: `jkp_factor_replication.qmd` (rendered to `jkp_factor_replication.html`). The note:

- documents each replicated factor, its source paper, and the methodology choices;
- benchmarks the US replications against the published reference series, with per-factor cumulative-return panels tabbed by model (axes shared across a model's tabs so switching tabs moves only the plotted lines);
- states the sample windows (with month + year) and notes that our code produces the factor return series and characteristics through the most recent available date, even where the reference series stops earlier (e.g. the DHS reference ends December 2024);
- covers reasons for divergence (CRSP Legacy vs CRSP V2, Compustat restatements) and includes a Harvard-style references section;
- includes an appendix replicating the daily data.

## Scope remaining

- International (ROW) parity notes, if/when reference series become available.

### LINKED ISSUE #180: [Bug]: quality_minus_junk crashes with Polars 32-bit row index overflow on world panel
## Bug Description

`quality_minus_junk` in `aux_functions.py` crashes the full-world pipeline with `pyo3_runtime.PanicException: Polars' maximum length reached`. Four `how="full"` joins (lines 10202, 10222–10224) materialize an intermediate that exceeds Polars' 32-bit row index limit on the world panel (~95 countries × ~600 months × N firms, accumulated over the 16-iteration loop).

## Script

main.py (data generation) — specifically `quality_minus_junk` called after `merge_world_data_prelim`.

## Error Category

Computation error (Polars row-limit / join planner blowup).

## Error Output

```
pyo3_runtime.PanicException: Polars' maximum length reached.
  at aux_functions.py:10202
  qmj = qmj.join(__z, how="full", coalesce=True, on=[...])
```

Observed on Slurm job 784160 running the full world pipeline.

## Root Cause

`z_ranks(qmj, ...)` is filtered from `qmj` itself — its `(excntry, eom, id)` keys are a strict subset of `qmj`'s keys. The `how="full"` joins are semantically over-broad; with `coalesce=True` they return the same row set a left join would, but the full-outer intermediate the planner materializes blows past the 32-bit row-index cap.

## Fix

Switch all four joins (lines 10202, 10222–10224) from `how="full"` to `how="left"`. Same output, no blowup. Already landed in commit b66b774 on branch `factors_rep` (PR #171).

## Steps to Reproduce

1. Run full `jkp build` on world panel with all ~95 countries.
2. Pipeline completes `merge_world_data_prelim`.
3. Crashes inside `quality_minus_junk` at the first full-join.

## Expected Behavior

`quality_minus_junk` returns the QMJ characteristic frame for the full world panel without overflowing Polars' row-index limit.

## Environment

- OS: Linux (Slurm cluster)
- Python: 3.11
- Polars: 1.x (32-bit row index build)
- RAM: ~450 GB (not the binding constraint — row-count limit is)

## Additional Context

Fix shipped on `factors_rep` branch (commit b66b774) while working on PR #171. Filing this issue retroactively for traceability.

### LINKED ISSUE #236: [Feature]: Build DHS FIN monthly factor returns
We need a faithful build of the monthly DHS FIN (financing) factor following Daniel, Hirshleifer, and Sun (2020).

FIN is built from the overlap of net stock issuance (NS) and composite share issuance (IR): annual June reconstitution, NYSE breakpoints, a size × financing sort, value-weighting, and a July-to-June holding period.

The output should land in `dhs_factors_monthly.parquet` as `fin_dhs` (eom + excntry + factor columns).

### LINKED ISSUE #237: [Feature]: Build DHS FIN daily factor returns
Daily counterpart of the DHS FIN (financing) factor from Daniel, Hirshleifer, and Sun (2020).

The same June-formed FIN portfolios are value-weighted on daily returns.

The output should land in `dhs_factors_daily.parquet` as `fin_dhs` (date + excntry + factor columns).

### LINKED ISSUE #238: [Feature]: Build DHS PEAD monthly factor returns
We need a faithful build of the monthly DHS PEAD (post-earnings-announcement drift) factor following Daniel, Hirshleifer, and Sun (2020).

PEAD is built from the cumulative abnormal return (ABR) around quarterly earnings announcements: a size × ABR sort with monthly rebalancing, value-weighting.

The output should land in `dhs_factors_monthly.parquet` as `pead_dhs` (eom + excntry + factor columns).

### LINKED ISSUE #239: [Feature]: Build DHS PEAD daily factor returns
Daily counterpart of the DHS PEAD (post-earnings-announcement drift) factor from Daniel, Hirshleifer, and Sun (2020).

The monthly-formed PEAD portfolios are value-weighted on daily returns.

The output should land in `dhs_factors_daily.parquet` as `pead_dhs` (date + excntry + factor columns).

### LINKED ISSUE #240: [Feature]: Build DHS stock-level characteristics
Write the per-stock DHS sort variables used to form the FIN and PEAD factors.

Columns: `ns_dhs` (net stock issuance), `ir_dhs` (composite share issuance), and `abr_dhs` (cumulative abnormal return around quarterly earnings announcements).

The output should land in `dhs_characteristics.parquet` keyed by `id, eom` (id is Int64 BIGINT, matching the jkp world_msf.id convention).

### LINKED ISSUE #244: [Feature]: Add Davis-Fama-French Moody's book-equity data to the FF pipeline
The FF book-equity ladder from Compustat only reaches back to the Compustat coverage start (~1950s/60s). To match Fama-French's published FF3 back to July 1926, the pipeline needs the Davis-Fama-French hand-collected Moody's book-equity data (distributed via [Ken French's data library](https://mba.tuck.dartmouth.edu/pages/faculty/ken.french/data_library.html#BookEquity)).

## Scope
- Parse the Moody's/DFF book-equity file into a long panel (permno, year, be).
- Union it with the Compustat book equity: `be = coalesce(dff.be, comp.be)`, aligning DFF be(t) to Dec(t-1) ME so the June formation pairs correctly.
- Gate behind a flag so the splice is explicit, and extend the FF3 SMB/HML (and BE/ME chars) back to June 1926.

## Notes
- `be` is in $M, identical to Compustat, so the beme scale is unchanged.
- Modern Compustat backfill creates (permno, year) collisions; resolve with DFF-first coalesce.

FernandoRDLL:
## Summary

Adds the factor-return and stock-level characteristic builders for the six canonical asset-pricing models — FF3, FF5, UMD, HXZ4, Stambaugh-Yuan mispricing, and DHS behavioral factors — along with the orchestrator (`ap_factor_model_data`) that outer-merges them into unified AP outputs, and the release documentation for the replication.

### Factor returns

- **Closes #157** (FF3 monthly) — `gen_ff_data` builds the FF3 SMB/HML factors on the USA universe with annual June reconstitution, NYSE breakpoints, and 2×3 independent value-weighted sorts. Writes `ff_factors_monthly.parquet`.
- **Closes #158** (FF3 daily) — same `gen_ff_data` build emits the daily counterpart in `ff_factors_daily.parquet` with daily VW returns within June-formed portfolios.
- **Closes #159** (FF5 monthly) — `gen_ff_data` extends the FF3 machinery to the FF5 SMB, RMW, and CMA factors, reusing the same June/breakpoint logic with operating profitability and asset growth as additional sort signals.
- **Closes #160** (FF5 daily) — daily FF5 factors in `ff_factors_daily.parquet` alongside FF3.
- **Closes #161** (UMD monthly) — `gen_ff_data` adds the Carhart momentum factor as a monthly-rebalanced 2×3 size × 12-2 prior-return sort. Writes `umd_ff` to the monthly output.
- **Closes #162** (UMD daily) — daily UMD factor alongside the monthly version.
- **Closes #163** (HXZ4 monthly) — `gen_hxz_data` builds the q-factor model factors (size, IA, ROE) via a 2×3×3 independent sort with NYSE breakpoints. The ROE bucket updates monthly as new quarterly earnings land. Writes `hxz_factors_monthly.parquet`.
- **Closes #164** (HXZ4 daily) — `gen_hxz_data` emits the daily counterpart in `hxz_factors_daily.parquet`.
- **Closes #165** (Mispricing monthly) — `gen_mispricing_data` builds the Stambaugh-Yuan 11-anomaly + CHS DISTRESS factors via percentile ranking and double-sort on size × MGMT/PERF score. Writes `mp_factors_monthly.parquet`. Uses pure CIZ DLRET semantics (trusts `mthret`, no Shumway / `delret` overlay).
- **Closes #166** (Mispricing daily) — `gen_mispricing_data` adds the daily build for both USA and 33 ROW countries via `_mp_world_build_portfolios_daily`. Writes `mp_factors_daily.parquet`.
- **Closes #236** (DHS FIN monthly) — `gen_dhs_data` builds the FIN (financing) factor from the overlap of net stock issuance and composite share issuance, with annual June reconstitution, NYSE breakpoints, and value-weighting. Writes `fin_dhs` to `dhs_factors_monthly.parquet`.
- **Closes #237** (DHS FIN daily) — daily `fin_dhs` in `dhs_factors_daily.parquet`, value-weighting the June-formed FIN portfolios on daily returns.
- **Closes #238** (DHS PEAD monthly) — `gen_dhs_data` builds the PEAD (post-earnings-announcement drift) factor from the cumulative abnormal return (ABR) around quarterly earnings announcements, via a monthly-rebalanced size × ABR sort. Writes `pead_dhs` to `dhs_factors_monthly.parquet`.
- **Closes #239** (DHS PEAD daily) — daily `pead_dhs` in `dhs_factors_daily.parquet`.

### Characteristics

- **Closes #167** (FF3 chars) — `gen_ff_data` writes the FF3 stock-level chars (ME, BE/ME) to `ff_characteristics.parquet`. `id` is Int64 BIGINT, matching the jkp `world_msf.id` convention.
- **Closes #168** (FF5 chars) — same writer adds the FF5 chars (OP, INV) alongside the FF3 chars in the same file.
- **Closes #169** (HXZ4 chars) — `gen_hxz_data` writes the HXZ chars (ME, IA, ROE) to `hxz_characteristics.parquet`. ROE updates monthly with new quarterly releases.
- **Closes #170** (Mispricing chars) — `gen_mispricing_data` writes the per-stock MGMT and PERF mispricing scores to `mp_characteristics.parquet`.
- **Closes #240** (DHS chars) — `gen_dhs_data` writes the per-stock DHS sort variables (`ns_dhs`, `ir_dhs`, `abr_dhs`) to `dhs_characteristics.parquet`.

### Orchestrator

- `ap_factor_model_data` outer-joins the four monthly factor parquets on `[excntry, eom]` (adding `mktrf`), the four daily factor parquets on `[excntry, date]`, and the four characteristic parquets on `[id, eom]`, producing `ap_factors_monthly.parquet`, `ap_factors_daily.parquet`, and `ap_factors_characteristics.parquet`. These feed downstream regression code (`prep_data_factor_regs`, `market_beta`, `residual_momentum`, `prepare_daily`).
- `generate_factor_models` / `combine_factor_models` are the single source of truth for the model set: one ordered `{prefix: generator}` mapping fixes both the run order and the order the per-model outputs feed the merge.

### Documentation

- **Closes #178** — release documentation for the replication, shipped under `documentation/ap_factor_replication/` (`jkp_factor_replication.qmd`, rendered `jkp_factor_replication.html`): per-factor cumulative-return panels benchmarked against the published reference series, sample windows, reasons for divergence, and a Harvard-style references section.

### Other changes in this PR

- **Closes #244** (FF Moody's book equity) — `gen_ff_data` unions the Davis-Fama-French hand-collected Moody's book equity (`load_dff_be` / `_ff_merge_dff_be`, gated by `FF_USE_DFF_BE`) with Compustat BE to extend the FF3 SMB/HML factors and BE/ME characteristic back to June 1926.
- Renamed the FF factor output columns `hml`/`rmw`/`cma` → `hml_ff`/`rmw_ff`/`cma_ff` for cross-model consistency (`smb_ff` → `smb_ff3` with new `smb_ff5`); downstream `prep_data_factor_regs` / `prepare_daily` alias back where needed. See `CHANGELOG_DATA.md`.
- Added the source-paper citation to each factor-generator docstring (FF 1993/2015, HXZ 2015, Stambaugh-Yuan 2017, DHS 2020), plus the WRDS "Fama-French Factors (CIZ Format)" and Ken French book-equity data-library URLs in `gen_ff_data`.
- Compustat quarterly dedup tie-break aligned with the SAS NODUPKEY ordering.
- ME is now read directly from the CIZ `mthcap` field (no `prc * shrout` round-trip).
- `DataPaths` threaded through `prep_data_factor_regs`, `market_beta`, `residual_momentum`, and `prepare_daily` to match the main-branch convention.
- The `excntry` column was dropped from the per-model characteristic parquets since `id` is already globally unique.
- DuckDB progress bars are silenced on the mispricing connections.

### Bug fixes included

- **Closes #180** (`quality_minus_junk` Polars 32-bit row-index overflow) — switched four `how="full"` joins to `how="left"` in `quality_minus_junk` (commit b66b774). `z_ranks(qmj, ...)` keys are a strict subset of `qmj`'s keys, so the full-outer joins were semantically over-broad and the planner's intermediate blew past Polars' max row count on the world panel. Left-join returns identical output without overflow.

### Follow-up test coverage

This PR ships the factor builders but no tests. Test work is tracked in eight follow-up issues opened against this PR:

- #172 — Unit tests for FF3 (`gen_ff_data`)
- #173 — Golden-fixture tests for FF3 (`gen_ff_data`)
- #174 — Unit tests for HXZ4 (`gen_hxz_data`)
- #175 — Golden-fixture tests for HXZ4 (`gen_hxz_data`)
- #176 — Unit tests for Mispricing (`gen_mispricing_data`)
- #177 — Golden-fixture tests for Mispricing (`gen_mispricing_data`)
- #241 — Unit tests for DHS (`gen_dhs_data`)
- #242 — Golden-fixture tests for DHS (`gen_dhs_data`)


src/jkp/data/aux_functions.py
```diff
@@ -7632,1334 +10065,3651 @@ def acc_chars_list():
...
+           each monthly factor parquet on (excntry, eom); sink.
+        2) Daily:   same with date key, mkt_daily_path.
+        3) Chars:   outer-join each chars parquet on (excntry, id, eom); sink.
+
     Output:
-        LazyFrame with f'beta_dimson{sfx}'.
+        - <out_monthly>: [excntry, eom, mktrf] + union of columns from each
+                          factor parquet in monthly_factor_inputs.
+        - <out_daily>:   [excntry, date, mktrf] + union of daily_factor_inputs.
+        - <out_chars>:   [excntry, id, eom]    + union of chars_inputs.
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in 2d355e1 — docstring + Output schema for `<out_chars>` now say `[id, eom]` (excntry dropped by per-model chars writers).

documentation/ap_factor_replication/jkp_factor_replication.qmd
```diff
@@ -0,0 +1,360 @@
...
+
+This document benchmarks the **US** replications against the original /
+WRDS-distributed series. For the international panels, we have made
+discretionary choices in the construction of each factor that try to
+emulate the spirit of the original US procedure (e.g., NYSE-style size
+breakpoints replaced with country-specific breakpoints, accounting
+conventions adapted to each excntry's Compustat coverage, etc.).
+
+Factor construction is implemented in `ap_factor_model_data` in
+[`src/jkp/data/aux_functions.py`](../src/jkp/data/aux_functions.py),
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in 2d355e1 — link now `../../src/jkp/data/aux_functions.py` so it resolves from `documentation/ap_factor_replication/`.

tests/golden/generate_ff_golden.py
```diff
@@ -0,0 +1,59 @@
+"""Regenerate the FF (FF3/FF5/UMD) golden parquets.
+
+Reads slices from ``tests/golden/fixtures/wrds_slices/``, stages them
+into a temporary DataPaths layout, runs ``gen_ff_data``, and writes the
+results under ``tests/golden/fixtures/ff/``.
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in 2d355e1 — docstring now references `tests/golden/fixtures/synthetic_wrds/` (`SYNTHETIC_SLICES_DIR`) instead of the obsolete `wrds_slices/` path.

src/jkp/data/aux_functions.py
```diff
@@ -6971,320 +6892,2835 @@ def gen_misp_exp(var_list, min_fcts):
...
+    """Stambaugh-Yuan 11-anomaly mispricing factor pipeline (incl. CHS DISTRESS), US + WORLD.
+
+    Always runs both branches; outputs concatenated with `excntry` column.
+    Writes 3 outputs to CWD:
+      mp_factors_monthly.parquet  — eom, excntry, smb_mispricing, mispricing_mgmt, mispricing_perf
+      mp_factors_daily.parquet    — date, excntry, smb_mispricing, mispricing_mgmt, mispricing_perf (US + world)
+      mp_characteristics.parquet  — id, eom, mispricing_mgmt, mispricing_perf
+    """
+    global _MP_RAW, _MP_INTERIM
+    _MP_RAW = paths.raw_tables_dir.as_posix()
```
capellini: `gen_mispricing_data` mutates the module-level globals `_MP_RAW` and `_MP_INTERIM` to point them at the caller's `paths.raw_tables_dir` and `paths.interim_dir`. Every `_mp_*` helper downstream then reads these globals to find its input files.

This re-introduces the exact anti-pattern that PR #132 was specifically written to remove. The body of #132 spells out the problem: "library code shouldn't mutate process-global state — it silently breaks callers, leaves state corrupt on exception, and makes the functions un-runnable in parallel." All three of those concerns apply here. Two concurrent calls to `gen_mispricing_data` (e.g., in a test suite or a future parallel pipeline) would race on the globals. An exception partway through `gen_mispricing_data` leaves stale paths set for any later unrelated `_mp_*` use in the same process. And the `@functools.cache` decorators on `_mp_load_funda` / `_mp_load_fundq` (see separate comment) interact badly with this — the first caller's paths get baked into the cache for the lifetime of the process.

Could the mispricing pipeline be refactored to pass `raw_dir: Path` and `interim_dir: Path` as explicit arguments through the `_mp_*` helpers, mirroring how `gen_ff_data` and `gen_hxz_data` handle this internally? That would eliminate the globals entirely and bring this code in line with the post-#132 convention.

src/jkp/data/aux_functions.py
```diff
@@ -6971,320 +6892,2835 @@ def gen_misp_exp(var_list, min_fcts):
...
+#   ccmxpf_lnkhist,funda,fundq,crsp_a_indexes_msp500,crsp_a_indexes_acti}.parquet.
+# ============================================================================
 
-    Steps:
-        1) Load/Filter monthly stock panel; keep id/eom/excntry + factor inputs.
-        2) Iteratively join rank-normalized columns per variable with proper direction.
-        3) Build mispricing_mgmt from vars_mgmt and mispricing_perf from vars_perf using gen_misp_exp.
-        4) Keep id/eom + both composites and write parquet.
+_MP_RAW = "../raw/raw_tables"
+_MP_INTERIM = "."  # CWD per jkp convention (pipeline chdirs to interim_dir)
```
capellini: The trailing comment here — `# CWD per jkp convention (pipeline chdirs to interim_dir)` — describes the world before PR #132 was merged. The pipeline no longer chdirs anywhere; that's the headline change in #132 (removed both `os.chdir` calls and every `os.system` invocation). Anyone reading this comment today would be misled into thinking the chdir convention is still in effect, which could cause them to write new code that depends on it.

Either remove the comment or replace it with a note explaining that the default value is a fallback and that `gen_mispricing_data` overrides both globals at runtime (though removing the globals entirely is what the separate comment above is suggesting).

tests/golden/test_gen_mispricing_data_golden.py
```diff
@@ -0,0 +1,51 @@
...
+            regen_or_compare(
+                actual=test_paths.interim_dir / name,
+                golden=GOLDEN_DIR / name,
+                key_cols=keys,
+                regen=regen,
+            )
+        )
+    if regen:
+        pytest.skip("regenerated golden parquets")
+    assert not failures, "\n".join(failures)
```
capellini, line 67: Heads-up that this test currently passes vacuously. The committed golden fixtures `tests/golden/fixtures/mispricing/mp_factors_daily.parquet` and `mp_factors_monthly.parquet` are both zero rows (5-column schema only), and `gen_mispricing_data` running on the synthetic universe also produces zero rows for those two factor outputs — most likely because the synthetic data has roughly 500 US firms plus 50 firms per ROW country, which isn't enough to populate the size×score buckets past the `MP_MIN_STKS_BP_WORLD = 10` and `MP_MIN_OBS_PF_WORLD = 3` gates. So `compare_factor_parquet` compares an empty actual against an empty expected and reports success.

The net effect is that the primary deliverable of issues #165 and #166 — the `smb_mispricing`, `mispricing_mgmt`, and `mispricing_perf` factor return series — has no regression coverage from the golden suite. The `mp_characteristics.parquet` fixture is substantive (7,489 rows) and does exercise the per-stock characteristic path, so that part of the test works.

Two ways to address this: (a) enrich `generate_synthetic_wrds.py` so the mispricing portfolios actually form on the synthetic data — probably means more firms per country and enough variance across the anomaly inputs to populate the percentile ranks; or (b) add a one-line `assert df.height > 0` guard inside `compare_factor_parquet` in `_golden_helpers.py` so the test fails until the fixture becomes meaningful, deferring the synthetic-data work to follow-up #177 explicitly. The first is the real fix; the second is the minimum viable acknowledgement that the gap exists.

tests/unit/test_gen_mispricing_data.py
```diff
@@ -0,0 +1,2083 @@
...
+
+from __future__ import annotations
+
+import math
+from datetime import date
+
+import numpy as np
+import polars as pl
+import pytest
+from polars import col
```
capellini: Small one — `from polars import col` here brings the bare `col` name into scope, and the rest of the file uses `col(...)` rather than `pl.col(...)`. CLAUDE.md asks new code to use the namespaced `pl.col(...)` form (the bare-`col` import only persists in existing legacy code for backward compatibility). The sibling test files added in this PR (`test_gen_ff_data.py`, `test_gen_hxz_data.py`) already use `pl.col`, so this is the odd one out.

Removing the import and replacing every `col(...)` call in this file with `pl.col(...)` would bring it into line with the convention and the rest of the new test suite.

src/jkp/data/aux_functions.py
```diff
@@ -7632,1334 +10068,3656 @@ def acc_chars_list():
...
+        "inv_ff": pl.col("inv"),
+    }
+    if mom_signal is not None:
+        cols["umd_ff"] = pl.col("mom_2_12")
+    return p.select("date", "excntry", "id", **cols).sort("date", "excntry", "id")
 
 
-def prc_to_high(df, sfx, __min):
+@measure_time
+def gen_ff_data(
```
capellini, line 13446: A consistency question about the API surface for the four new top-level functions added in this PR (`gen_ff_data`, `gen_hxz_data`, `gen_mispricing_data`, `ap_factor_model_data`).

The post-PR-#132 convention in `main.py` is for the caller to compose absolute `Path` objects against `paths.interim_dir` and pass them explicitly to each pipeline function. For example:

```python
market_returns(paths, interim / "world_dsf.parquet", "d", ...)
create_world_data_prelim(paths, interim / "world_msf.parquet", ...)
firm_age(paths, interim / "world_msf.parquet")
```

The new functions added by this PR take bare string filenames instead:

```python
gen_ff_data(paths, "ff_factors_monthly.parquet", "ff_factors_daily.parquet", ...)
```

Internally the function does `interim_dir = paths.interim_dir` and then composes `interim_dir / monthly_factors_path`, so it isn't actually depending on the process cwd — but the API surface looks like it is, and it doesn't match the dominant idiom in `main.py`. A future reader is likely to assume the bare-string parameters are resolved against cwd (the way `gen_ff_data`'s docstring just above explicitly says: "Runs from the pipeline cwd (= interim/)"), and write new code that depends on that assumption.

Would it be reasonable to switch these four signatures to take `Path` arguments and let `main.py` compose them at the call site, the way every other pipeline function works? The change is mechanical:

```python
def gen_ff_data(
    paths: DataPaths,
    monthly_factors_path: Path,
    daily_factors_path: Path,
    chars_path: Path,
    freqs: tuple[str, ...] = ("monthly", "daily"),
) -> None: ...
```

and at the call site:

```python
gen_ff_data(
    paths,
    interim / "ff_factors_monthly.parquet",
    interim / "ff_factors_daily.parquet",
    interim / "ff_characteristics.parquet",
)
```

Same idea for `gen_hxz_data` and `ap_factor_model_data` (its `*_inputs: list[str]` and bare-string `out_*` parameters). For `gen_mispricing_data` the change is bigger because the function currently takes no path parameters at all (see separate comment).

src/jkp/data/aux_functions.py
```diff
@@ -6971,320 +6892,2835 @@ def gen_misp_exp(var_list, min_fcts):
...
+    if sc_world is not None:
+        sc = pl.concat(
+            [sc, sc_world.select("id", "eom", "mispricing_mgmt", "mispricing_perf")],
+            how="vertical_relaxed",
+        )
+    _mp_write(sc.sort("id", "eom"), "mp_characteristics")
+
+
+@measure_time
+def gen_mispricing_data(
```
capellini, line 9899: `gen_mispricing_data` is the only one of the four new pipeline functions that takes no path parameters at all — its docstring says it "writes 3 outputs to CWD." This is the strongest signal that the function still assumes the pre-PR-#132 cwd-anchored model.

The downstream evidence is in `tests/golden/test_gen_mispricing_data_golden.py` at the test body: the test has to wrap the call in `with cwd(test_paths.interim_dir):` to make the file writes land where the test expects them. The fact that the test needs a `chdir` context manager just to invoke the function is a sign the API is fighting the convention. (And note that this chdir-in-test pattern is itself something PR #132 was trying to eliminate — it's only acceptable here because it's scoped to a single test.)

The natural shape is to give `gen_mispricing_data` the same kind of explicit `Path` output parameters that `gen_ff_data` and `gen_hxz_data` have:

```python
def gen_mispricing_data(
    paths: DataPaths,
    monthly_factors_path: Path,
    daily_factors_path: Path,
    chars_path: Path,
    min_stks: int = 30,
    min_fcts: int = 3,
    min_obs: int = 10,
    min_stks_world: int = MP_MIN_STKS_BP_WORLD,
    min_obs_world: int = MP_MIN_OBS_PF_WORLD,
) -> None: ...
```

Once the three output paths are explicit parameters, the `with cwd(...)` workaround in the golden test can go away, and the function no longer has any reason to mutate the module-level globals discussed in the comments above.

src/jkp/data/aux_functions.py
```diff
@@ -6971,320 +6892,2835 @@ def gen_misp_exp(var_list, min_fcts):
...
+
+# ============================================================================
+# STAGE 2 — LOAD + FILTER FUNDAMENTALS (Compustat US + acc_chars_world + g_fundq)
+# ============================================================================
 
-    pi = 3.141592653589793
-    k2 = sqrt(8 / pi)
-    const = 3 - 2 * sqrt(2)
+
+@functools.cache
```
capellini, line 7719: `_mp_load_funda` and `_mp_load_fundq` (immediately below) are both decorated with `@functools.cache`, and both read the module-level `_MP_RAW` global to find the parquet files. Neither function takes any arguments, so the cache key is just `()` — the cached result is determined entirely by the value of `_MP_RAW` at the time of the first call.

This is fine when there's a single `gen_mispricing_data` invocation per process, which is the production case. It's a latent hazard in any scenario where a process makes multiple `gen_mispricing_data` calls with different `paths` arguments — for example, a notebook or a future test that exercises multiple pipeline configurations in one Python session. The first call writes its `paths.raw_tables_dir` into `_MP_RAW` and the load functions cache the resulting DataFrame keyed by `()`. The second call writes a different value into `_MP_RAW`, but the load functions return the cached DataFrame from the first call — silently reading from the wrong directory.

Two ways to fix this cleanly: pass `raw_dir: Path` as an explicit argument to both loaders so it becomes part of the cache key, or call `_mp_load_funda.cache_clear()` and `_mp_load_fundq.cache_clear()` at the top of `gen_mispricing_data` before the new paths are set. The first is the cleaner long-term fix and falls out naturally if the module-globals-mutation pattern is removed as suggested in the earlier comments.

src/jkp/data/aux_functions.py
```diff
@@ -7632,1334 +10068,3656 @@ def acc_chars_list():
...
+        )
+        .filter(
+            (pl.col("indfmt") == "INDL")
+            & (pl.col("datafmt") == "STD")
+            & (pl.col("popsrc") == "D")
+            & (pl.col("consol") == "C")
+        )
+        .with_columns(
+            ps=pl.coalesce("pstkrv", "pstkl", "pstk", pl.lit(0.0)),
+            txditc=pl.when(pl.col("datadate").dt.year() < 1993)
```
capellini, line 12013: Just want to flag this for explicit confirmation rather than treat it as a defect. The conditional here zeros out `txditc` for any observation with `datadate.year() >= 1993`, only adding the deferred-tax balance to book equity for pre-1993 fiscal years. The variable name `txditc_FASB109` in the docstring at line 11468 and the commit message on `ad4e64c` ("FASB 109 TXDITC cutoff (Fama-French 2016 convention)") make it clear this is intentional — it matches Fama-French's own published convention from 2016 onward, which dropped the deferred-tax add-back for post-FASB-109 fiscal years.

The reason it's worth flagging: the JKP pipeline's own book-equity formula in `be_x` (around line 4016) adds `txditc` (via `pl.coalesce(["txditc", 0.0])`) **unconditionally**, with no 1993 gate. And `hxz_load_funda` at line 12610 explicitly applies `txditc.fill_null(0.0)` unconditionally too, with a docstring at line 12594 that flags the contrast ("HXZ uses TXDITC unconditionally, in contrast to ff_load_compustat_us").

So the FF replication's BE is now structurally different from the JKP characteristic BE for every observation after fiscal year 1992 — by design, but worth confirming with @theisij that this is the methodological choice we want. If the answer is "yes, FF replication should match FF," no code change is needed; if the answer is "we want JKP-consistent BE everywhere," the gate should be dropped here too.

src/jkp/data/aux_functions.py
```diff
@@ -7632,1334 +10068,3656 @@ def acc_chars_list():
...
+            formation_max=pl.col("datadate").dt.offset_by("18mo").dt.month_end(),
         )
-        .select(["id_int", "group_number", f"zero_trades{sfx}"])
+        .join(fmtimes, how="cross")
+        .filter(
+            (pl.col("date") >= pl.col("formation_min"))
+            & (pl.col("date") <= pl.col("formation_max"))
+        )
+        # Tie-break by roe_a so polars threading can't shuffle tied
+        # (gvkey, date, datadate) rows. Larger |roe_a| (= more material
```
capellini: The comment on the previous two lines claims that "Larger |roe_a| (= more material filing) wins via roe_a DESC after datadate ASC." Reading the sort + unique together: the sort key is `(gvkey ASC, date ASC, datadate ASC, roe_a DESC)` with `nulls_last=True`, and the dedup uses `keep="last"` over `(gvkey, date)`. Within a `(gvkey, date)` group, keeping the last row means keeping the maximum `datadate` (most recent fiscal date — correct, and presumably the dominant intent), and **among rows tied on datadate**, keeping the last row after sorting `roe_a DESC` means keeping the **smallest** `roe_a`, not the largest.

The numerical impact is negligible because identical `(gvkey, date, datadate)` triples only show up as rare linker artifacts where the choice between duplicates doesn't really matter. But the comment as written describes the opposite behavior to what the code actually does. The simplest fix is to rewrite the comment to describe the actual tie-break ("ties on datadate are broken arbitrarily but deterministically"), or to flip `keep="last"` to `keep="first"` (which would make the comment's stated intent true) — either is fine.

The same pattern appears at lines 13258–13266 in `_hxz_us_roe_monthly` for the US path, where the comment is more careful (it explicitly says the tie-break is for determinism only) but the underlying mechanic is the same.

src/jkp/data/aux_functions.py
```diff
@@ -6971,320 +6892,2835 @@ def gen_misp_exp(var_list, min_fcts):
...
+        .with_columns(gp_adj=(col("revt") - col("cogs")) / col("at"))
+        .filter(col("gp_adj").is_not_null())
+        .select("permno", "datadate", "gp_adj")
+        .sort(["permno", "datadate", "gp_adj"], nulls_last=True)
+        .unique(subset=["permno", "datadate"], keep="first", maintain_order=True)
+    )
+    return _mp_write(_mp_expand_to_crsp_window(ca, "gp_adj"), "gp_adj")
+
+
+def _mp_compute_oscore():
```
capellini: Convention nit that applies across several of the new mispricing, FF, and HXZ helpers: CLAUDE.md asks division on Polars expressions to go through the `safe_div()` helper rather than the raw `/` operator (the idea being that null and zero handling becomes uniform across the codebase). For example, `_mp_compute_oscore` has five raw divisions in its body (`col("lt") / col("at")`, `(col("act") - col("lct")) / col("at")`, etc.), each of which would become `safe_div(col("lt"), col("at"))` and so on.

Most of these are inside `pl.when()` guards that prevent zero-division at runtime, so this isn't a correctness issue — it's purely about convention consistency. The new sites I noticed are:

- in `_mp_*` helpers: `_mp_vw_return`, `_mp_add_leg_score`, `_mp_compute_gross_profit`, `_mp_compute_oscore`, `_mp_compute_nsi_ag_inv_noa`, `_mp_compute_roa`, `_mp_distress_book_equity_mb` (US + world), `_mp_distress_quarterly_inputs` (US + world), `_mp_distress_sigma` (US + world)
- in `ff_*` helpers: `ff_load_compustat_us` (OP and INV), `ff_prepare_chars_weights_rets` (BE/ME), `ff_load_world_compustat` (OP and INV), `_ff_vw_leg`
- in `hxz_*` helpers: `hxz_supplement_q4_shares`, `hxz_compute_factors` (the VW return)

The migration is mechanical. Pre-existing functions in the file (`z_ranks`, `res_mom`, `dolvol`, `turnover`, etc.) also use raw `/` and aren't introduced by this PR, so they're out of scope here.

capellini (changes requested):
Thanks for the substantial work on this PR — the factor-builder code is well-decomposed, the methodology is clearly grounded in the source papers, and the unit-test coverage on the helper functions is excellent. Before merge I have a number of inline comments below, plus a few documentation gaps.

---

A few documentation gaps I noticed while reviewing this PR — none of them are in the diff, so flagging them here rather than inline:

1. `documentation/documentation.tex` lines 2213–2239 currently describe `mispricing_mgmt` and `mispricing_perf` using the **old** JKP rank-average formula (averaging O_SCORE, RET_12_1, GP_AT, NIQ_AT for PERF; CHCSHO_12M, EQNPO_12M, OACCRUALS_AT, NOA_AT, AT_GR1, PPEINV_GR1A for MGMT). This PR replaces that construction with the Stambaugh-Yuan original (6-anomaly MGMT composite, 5-anomaly PERF composite), so the published documentation now actively misdescribes the code. This one feels important enough to land in this PR rather than #178, even if the broader rewrite waits.

2. Nine new stock-level characteristics (`me_ff`, `beme_ff`, `op_ff`, `inv_ff`, `umd_ff`, `me_hxz`, `ia_hxz`, `roe_hxz`, `smb_mispricing`) and three new public output files (`ap_factors_monthly.parquet`, `ap_factors_daily.parquet`, `ap_factors_characteristics.parquet`) have no entries anywhere in `documentation.tex`. Probably fits under #178, but worth tracking.

3. The changelog table near line 116 of `documentation.tex` has no entry for this PR — given the scope (new factor builders, behaviour-changing replacement of `mispricing_mgmt` / `mispricing_perf`, new output parquets), a single bullet would be appropriate.

4. The new `documentation/ap_factor_replication/jkp_factor_replication.qmd` imports `plot_cumsum_ff_usa` and references `factor_comparison/Reference_data/`, neither of which is committed (the `.gitignore` in that directory excludes everything but `*.qmd`, `*.html`, and `*_files/`). The pre-rendered `.html` is the only artifact a reader can see. Either commit the missing helper script and reference data, or add a note at the top of the `.qmd` explaining the file can't be re-rendered from a fresh clone.

---

Separately, there are three concrete things I'd be happy to help with — just say the word on any or all of them.

**Rebasing against current `origin/main`.** GitHub is showing the PR as `CONFLICTING` right now. The merge-base of your branch is `0e91cfb`, which is actually PR #132 itself (so your branch diverged at the moment #132 landed — more on that below), and `origin/main` is exactly one commit ahead of that: `7080186` "Consolidate portfolio.py helpers into aux_functions.py" (#146). That single commit touched `src/jkp/data/aux_functions.py`, which is the same file you've modified most heavily, so the conflict surface is concentrated entirely there. I'm happy to do the rebase myself, resolve the `aux_functions.py` conflicts on a working branch, and push it up for you to inspect before you replace your current head — whichever workflow you prefer. The portfolio consolidation is mechanical and shouldn't interact with anything in your new code.

**Conforming to the PR #132 conventions.** Several of the inline comments below — the `_MP_RAW` / `_MP_INTERIM` module-global mutation, the bare-string path arguments on the four new top-level functions, the stale "CWD per jkp convention" comment — all trace back to the same root cause: the conventions PR #132 established for explicit `Path` threading and the elimination of process-global state were brand-new when you branched (the merge-base really is #132's merge commit), and the new code in this PR adopts them partway but not all the way through. I can prepare a follow-up patch on top of your branch that handles the mechanical refactor: turn the bare-string path parameters on `gen_ff_data`, `gen_hxz_data`, `gen_mispricing_data`, and `ap_factor_model_data` into explicit `Path` parameters; eliminate the `_MP_RAW` / `_MP_INTERIM` globals by threading `raw_dir` and `interim_dir` through the `_mp_*` helpers as arguments; and update the golden-test wiring (which currently needs the `with cwd(...)` workaround) to match. You'd keep ownership of all the methodology decisions; I'd handle the plumbing.

**Removing WRDS-derived data from branch history.** The golden-fixture parquets under `tests/golden/fixtures/{ff,hxz,mispricing}/` initially contained real WRDS-derived per-security data — real CRSP PERMNOs, real characteristic values for around 16,000 securities across 2015–2022 in the `7d65c5b` version, and a smaller US-only subset of around 3,600 securities in the `a6f1498` version. Both were replaced with the synthetic versions at commit `d5432d0`, and your current branch tip is clean. Four `.dta` files committed at `bee6ccf` under `ff_data_comp/` are similarly removed from the final tree but still reachable through the PR's history. None of this affects `main` — a squash-merge would keep `main`'s permanent history clean either way — but the PR-ref view on GitHub still surfaces those intermediate commits, and a `gh pr checkout 171` continues to fetch them. I've put together a `git filter-repo` script that strips the nineteen real-data fixture blobs (preserving the final synthetic versions, which live at the same paths but with different blob SHAs) plus the five scaffolding directories that were added and removed in the same branch. It runs against a fresh throwaway clone, verifies that the rewritten tip is byte-identical to the current tip before letting me push anything, and the force-push would be scoped to the `factors_rep` branch only — `main` is untouched throughout. Nothing about the code or the synthetic fixtures changes; CI should pass on the rewritten branch unchanged. Happy to coordinate timing on this so it doesn't clash with anything else you're doing on the branch.

---

A note on how these work together: rebasing first is the simplest order, because once the branch is rebased onto current `origin/main` and the convention refactor can happen on top of a clean base.

capellini:
@theisij and @FernandoRDLL - I want to call out explicitly in a separate comment that we should *squash* commit this PR. That will prevent any WRDS data from making it into the main branch.

src/jkp/data/aux_functions.py
FernandoRDLL: Fixed — removed `_MP_RAW`/`_MP_INTERIM`. `gen_mispricing_data` now derives `raw_dir`/`interim_dir` from `paths` and threads them explicitly through every `_mp_*` helper and stage function, mirroring `gen_ff_data`/`gen_hxz_data`. A grep for `_MP_RAW|_MP_INTERIM` now returns 0.

src/jkp/data/aux_functions.py
FernandoRDLL: Fixed — the cwd-based globals are gone, so this stale comment was removed with them.

tests/golden/test_gen_mispricing_data_golden.py
FernandoRDLL: Fixed for real, and it surfaced a deeper bug. First, the fixtures were 0-row because synthetic prices (`me/shrout`, median ~$0.42) failed the `lag_prc > 5` mispricing floor; `generate_synthetic_wrds.py` now sets `shrout = me / target_price` ($10-100) so portfolios form. Then the regenerated mp_factors golden turned out NONDETERMINISTIC (~29% of US perf/SMB rows flipped run-to-run): the synthetic `crsp_a_indexes_msp500` fixture was generated DAILY (~22 rows/month), but real WRDS msp500 is monthly and `_mp_compute_distress` joins it on `caldt.dt.month_end()` assuming one row per eom — so the join fanned out ~22x and `unique(keep="first")` picked a nondeterministic survivor under DuckDB's parallel join order (only visible at high core counts; ruled out threads/float/sample-size). `gen_crsp_a_indexes_msp500` now emits one row per month-end; verified DISTINCT=1 across 8 runs on a 32-core host. mp_factors_monthly/daily are now substantive and reproducible.

tests/unit/test_gen_mispricing_data.py
FernandoRDLL: Fixed — dropped `from polars import col`; the file uses `pl.col(...)` throughout now.

src/jkp/data/aux_functions.py
FernandoRDLL: Fixed — `gen_ff_data`, `gen_hxz_data`, `gen_mispricing_data`, and `ap_factor_model_data` now take explicit `Path` params (and `ap_factor_model_data` takes `list[Path]` inputs / `Path` outputs); `main.py` composes them against `paths.interim_dir` at the call sites.

src/jkp/data/aux_functions.py
FernandoRDLL: Fixed — `gen_mispricing_data` now takes explicit `monthly_factors_path`/`daily_factors_path`/`chars_path` (`Path`) plus the formation-gate params. The `with cwd(...)` workaround in the golden test is gone, and the function no longer mutates module globals.

src/jkp/data/aux_functions.py
FernandoRDLL: Fixed — `_mp_load_funda`/`_mp_load_fundq` take an explicit `raw_dir: Path` that joins the `@functools.cache` key, and `gen_mispricing_data` calls `.cache_clear()` on both at entry. The stale-on-global-change hazard is gone.

src/jkp/data/aux_functions.py
FernandoRDLL: Confirmed intentional — this matches the FASB-109 adoption convention from the FF replication, and @theisij has reviewed it as correct (FF replication should match FF, not JKP-consistent BE). Left as-is.

src/jkp/data/aux_functions.py
FernandoRDLL: Fixed — comment rewritten: `keep="last"` after `datadate` ASC selects the most recent fiscal date; the trailing `roe_a` DESC only breaks remaining ties deterministically. No behavior change.

src/jkp/data/aux_functions.py
FernandoRDLL: Fixed — switched to `safe_div()` at these sites across the new `_mp_*`/`ff_*`/`hxz_*` helpers. `safe_div` was extended to accept Polars expressions/literals (not just column-name strings) so composite numerators/denominators work.

FernandoRDLL (commented):
@capellini 
Changes done!

FernandoRDLL:
Also fixes #182 — slurm script now uses `set -e` so a failed `jkp build` aborts the job instead of cascading into a misleading FileNotFoundError from `jkp portfolio`.

capellini:
> @capellini Changes done!

Great! Thank you, Fernando! What an amazing job. Our test coverage is much more complete now.

This PR LGTM.

> Also fixes #182 — slurm script now uses `set -e` so a failed `jkp build` aborts the job instead of cascading into a misleading FileNotFoundError from `jkp portfolio`.

Thanks, I've been meaning to take care of that. 

But could we do it in a separate PR, given that it's not really related to this PR and this one is already so large? It's best to keep PRs small and logically separated, if possible.

capellini (dismissed):

FernandoRDLL:
@claude please review the PR and make suggestions if appropriate

FernandoRDLL:
Thanks for the review — addressed in the latest commits.

**🔴 DHS international coverage never fires in production — fixed.** Confirmed exactly as described: `gen_dhs_data` runs inside `generate_factor_models` (main.py:187), before `merge_qmj_to_world_data` writes `world_data.parquet` (main.py:237), so the ROW branch's `.exists()` gate was always `False` on a fresh build and FIN/PEAD shipped US-only. Switched the ROW gate + all four ROW scans (`_dhs_load_row_panel`, `_dhs_row_chars`, `_dhs_ibes_announcements`, the ROW NYSE-median panel) to read `world_data_prelim.parquet` — the same fix `gen_mispricing_data` already uses. Verified the prelim carries every column the ROW path needs (`id, gvkey, excntry, eom, me, ret, shares, adjfct, size_grp, be_me`) and that `_dhs_row_universe` only filters columns the downstream QMJ merge never modifies, so US output is byte-unchanged.

**🟢 Nits — done:**
- `safe_div()` — routed the four hand-rolled div guards in the DHS FIN builders through `safe_div()`; the two ROW sites keep their extra calendar-gap condition as an outer `when()`. Validated null-and-value identical to the originals on edge-case data (den=0, den=null, gap mismatch, gap null, num null, negatives).
- `pl.col()` for the `mktrf` select in `_ap_join_factor_inputs`; added a docstring to `_dhs_wd_idx`.
- `ap_factor_model_data` — documented why it takes explicit `Path` lists rather than `paths: DataPaths` (it's a pure combinator; keeping the model set defined by the caller/`generators` mapping is deliberate).
- `documentation.tex` — corrected the `mispricing_mgmt`/`mispricing_perf` characteristic formulas from the old JKP 4/6-anomaly rank average to the Stambaugh-Yuan cluster percentile-rank averages (management: 6 anomalies, performance: 5), and updated the footnote.
- `jkp_factor_replication.qmd` — added a reproduction-note comment documenting the local, un-committed supporting files (`plot_cumsum_usa.py`, `styles.scss`, `Reference_data/`) and the required prior `jkp build`.

**🟡 DHS test coverage.** Deferring the unit + golden tests to the follow-up issues (#241, #242) — they'll be done in a separate PR rather than growing this one further.

FernandoRDLL:
@claude please review the PR and make suggestions if appropriate

FernandoRDLL:
@claude please review the code and make suggestions if appropriate

FernandoRDLL:
Addressed the review findings from both Claude passes. Status:

**Round 2 (latest)**
- 🔴 **ROW FF book equity double-FX-converted** — fixed in `a23e215`. `acc_std_ann` is built `convert_to_usd=1`, so `_jkp_annual_acc`'s `be_x`/`ope_x`/`at_x`/`ni_x` are already USD (verified on real data: `curcd == "USD"` for every row; GBP firm `at` 11,555 → `at_x` 18,632 USD). Dropped the `join_asof(fx)` + the now-dead `curcd` plumbing and set `be = be_x` directly. Verified `be == acc be_x` exactly (max|d| = 0).
- 🟡 **Outer-join coverage narrowing** — checked on real data: ROW `be` non-null 1.30M (vs ~1.26M before), `op` 1.20M, `inv` 1.25M. Not shrunk.
- 🟢 **HXZ `be_a` "local currency" docstring** — corrected in `a23e215` (also `_jkp_annual_acc` + FF docstrings). `be_a` is USD but only gates `be_a > 0` and is the same-currency `roe_a` denominator, so FX-invariant.
- 🟢 **`at_gr1`/`inv` currency noise** — intentional; matches JKP's own `at_gr1` methodology (computed on the USD panel). No change.

**Round 1**
- 🔴 **DHS international gate never fires** — fixed (`464842e`): `_dhs_load_row_panel`/`_dhs_row_chars`/`_dhs_ibes_announcements` and the row gate all read `world_data_prelim.parquet` (exists before `generate_factor_models`), mirroring `gen_mispricing_data`.
- 🟡 **DHS test coverage** — still tracked in #241/#242.
- 🟢 **`safe_div()` in `_dhs_row_chars`/`_dhs_issuance_chars`** — divisions now route through `safe_div` (`me/me_lag5`, `adj_csho/lag`); the remaining `when(x>0).then(x.log())` are log-domain guards, not divisions.
- 🟢 **bare `col()` at `_ap_join_factor_inputs`** — uses `pl.col("mkt_vw_exc")`.
- 🟢 **`_dhs_wd_idx` docstring** — added (Description of the weekday-index arithmetic).
- 🟢 **`ap_factor_model_data` no `paths` param** — fixed in `9aff23b`: now takes `paths: DataPaths` first (fixed I/O paths derived from `interim_dir`; only the per-model input lists stay caller-supplied).
- 🟢 **Doc gaps** — `documentation.tex` mispricing footnote now describes the Stambaugh-Yuan (2017) percentile-rank construction (cites `stambaugh2017`); the `.qmd` supporting files (`plot_cumsum_usa.py`, `styles.scss`, `Reference_data/`) are now committed (`5cf0026`), so a clone re-renders after a local `jkp build`.

FernandoRDLL:
@theisij 
Can you please run Copilot review here please? I think it's ready to go, including the documentation but I want to make sure I didn't miss anything.

FernandoRDLL:
@theisj this is how we build the chars for ROW countries:

## US vs. ROW characteristics

**US:** each factor replicated exactly per its source paper, from raw WRDS.

**ROW:** same construction and timing, but accounting characteristics come from JKP's standardized cross-country panel (`acc_std_ann`) instead of raw Compustat Global — for comparability. Only the raw inputs change, not the algorithms.

**ROW chars from JKP:**

| Model | Char | Definition |
|---|---|---|
| FF | `beme` | `be_x` / Dec ME |
| FF | `op` | `ope_x` / `be_x` |
| FF | `inv` | `at_gr1` |
| HXZ | `be_a` | `be_x` |
| HXZ | `inv` | `at_gr1` |
| HXZ | `roe_a` | `ni_x` / prior-yr `be_x` |
| DHS | `ns` | `log(1 + chcsho_12m)` |

**Timing preserved:** values sit at fiscal-year-end, June(t) formation on FY t−1 chars.

**One deliberate difference:** US HXZ ROE is quarterly (`IBQ`, RDQ-gated); ROW ROE stays **annual** — Global quarterly coverage too thin (~56% vs ~77% of firm-months) and a single-quarter ROE is ~¼ the annual magnitude, so mixing would misrank firms.

src/jkp/data/config.py
```diff
@@ -209,12 +209,165 @@
...
 # Variables can repeat across windows (e.g. zero_trades runs at 21d/126d/252d).
 ROLLING_DAILY_SPECS: list[tuple[str, int, list[str]]] = [
-    ("_21d", 15, ["rvol", "rmax", "skew", "capm_ext", "ff3", "hxz4", "dimsonbeta", "zero_trades"]),
-    ("_126d", 60, ["zero_trades", "turnover", "dolvol", "ami"]),
-    ("_252d", 120, ["rvol", "capm", "downbeta", "zero_trades", "prc_to_high", "mktvol"]),
-    ("_1260d", 750, ["mktcorr"]),
+    ("_21d", 15, ["ff3", "hxz4"]),
+    # ("_126d", 60, ["zero_trades", "turnover", "dolvol", "ami"]),
+    # ("_252d", 120, ["rvol", "capm", "downbeta", "zero_trades", "prc_to_high", "mktvol"]),
+    # ("_1260d", 750, ["mktcorr"]),
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in 7e12637 — the full ROLLING_DAILY_SPECS is restored (rvol/mktvol at 252d, mktcorr at 1260d, rmax at 21d alongside the ff3/hxz4 additions), so finish_daily_chars gets corr_1260d, rvol_252d, __mktvol_252d, and rmax5_21d again. Also added tests/unit/test_rolling_daily_specs_coverage.py so every spec var must dispatch in process_window.

tests/unit/test_rolling_daily_specs_coverage.py
```diff
@@ -0,0 +1,66 @@
...
+            and node.targets[0].id == "funcs"
+            and isinstance(node.value, ast.Dict)
+        ):
+            return {k.value for k in node.value.keys if isinstance(k, ast.Constant)}
+    raise AssertionError("funcs dict not found in process_window source")
+
+
+SPEC_PAIRS = [(sfx, var) for sfx, _, vars_ in ROLLING_DAILY_SPECS for var in vars_]
+SPEC_SUFFIXES = [sfx for sfx, _, _ in ROLLING_DAILY_SPECS]
+
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in 0c27523 — dropped the unused SPEC_SUFFIXES.

tests/unit/test_factor_model_data.py
```diff
@@ -0,0 +1,66 @@
...
+    assert list(outputs) == PREFIXES
+    assert outputs == {prefix: main.factor_paths(tmp_path, prefix) for prefix in PREFIXES}
+
+
+def test_combine_factor_models_feeds_ap_in_order(monkeypatch, tmp_path):
+    ap_kwargs: dict = {}
+    monkeypatch.setattr(main, "ap_factor_model_data", lambda *a, **kw: ap_kwargs.update(kw))
+
+    factor_outputs = {prefix: main.factor_paths(tmp_path, prefix) for prefix in PREFIXES}
+    main.combine_factor_models(tmp_path, factor_outputs)
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in 0c27523 — the test now passes DataPaths(base_dir=tmp_path) to combine_factor_models, matching the production call signature.

tests/unit/test_ff_dff_splice.py
```diff
@@ -0,0 +1,169 @@
+"""
+Unit tests for the DFF (Davis-Fama-French) hand-collected BE splice in
+ff_load_compustat_us.
+
+DFF (2000) collected Moody's BE for NYSE firms absent from Compustat; the
+splice unions the permno-keyed DFF rows with the CCM-linked Compustat rows,
+Compustat-first on any (permno, year) collision. DFF be(t) is publicly
+available by June 30 of year t and pairs with Dec(t-1) ME, so the unioned
+row carries year = t - 1 (the June-frame join uses cyp1 = year + 1).
```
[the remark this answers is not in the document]
FernandoRDLL: Fixed in 0c27523 — docstring corrected to DFF-first (be = coalesce(dff, comp), and the two sources agree where both exist).

theisij:
> @TheisJ this is how we build the chars for ROW countries:
> 
> ## US vs. ROW characteristics
> **US:** each factor replicated exactly per its source paper, from raw WRDS.
> 
> **ROW:** same construction and timing, but accounting characteristics come from JKP's standardized cross-country panel (`acc_std_ann`) instead of raw Compustat Global — for comparability. Only the raw inputs change, not the algorithms.
> 
> **ROW chars from JKP:**
> 
> Model	Char	Definition
> FF	`beme`	`be_x` / Dec ME
> FF	`op`	`ope_x` / `be_x`
> FF	`inv`	`at_gr1`
> HXZ	`be_a`	`be_x`
> HXZ	`inv`	`at_gr1`
> HXZ	`roe_a`	`ni_x` / prior-yr `be_x`
> DHS	`ns`	`log(1 + chcsho_12m)`
> **Timing preserved:** values sit at fiscal-year-end, June(t) formation on FY t−1 chars.
> 
> **One deliberate difference:** US HXZ ROE is quarterly (`IBQ`, RDQ-gated); ROW ROE stays **annual** — Global quarterly coverage too thin (~56% vs ~77% of firm-months) and a single-quarter ROE is ~¼ the annual magnitude, so mixing would misrank firms.

I think using the standardized accounting panel makes a lot of sense. But aren't we missing a bunch from the ROW chars from JKP overview? Like what about the Stambaugh-Yuan ones, and the DHS ones? We really should be in a position where we can generate all factors both within an outside of the U.S. (or if we can't I'd love to know why)

theisij:
Overall, the documentation looks amazing. Great work!!

FernandoRDLL:
> > @TheisJ this is how we build the chars for ROW countries:
> 
> > 
> 
> > ## US vs. ROW characteristics
> 
> > **US:** each factor replicated exactly per its source paper, from raw WRDS.
> 
> > 
> 
> > **ROW:** same construction and timing, but accounting characteristics come from JKP's standardized cross-country panel (`acc_std_ann`) instead of raw Compustat Global — for comparability. Only the raw inputs change, not the algorithms.
> 
> > 
> 
> > **ROW chars from JKP:**
> 
> > 
> 
> > Model	Char	Definition
> 
> > FF	`beme`	`be_x` / Dec ME
> 
> > FF	`op`	`ope_x` / `be_x`
> 
> > FF	`inv`	`at_gr1`
> 
> > HXZ	`be_a`	`be_x`
> 
> > HXZ	`inv`	`at_gr1`
> 
> > HXZ	`roe_a`	`ni_x` / prior-yr `be_x`
> 
> > DHS	`ns`	`log(1 + chcsho_12m)`
> 
> > **Timing preserved:** values sit at fiscal-year-end, June(t) formation on FY t−1 chars.
> 
> > 
> 
> > **One deliberate difference:** US HXZ ROE is quarterly (`IBQ`, RDQ-gated); ROW ROE stays **annual** — Global quarterly coverage too thin (~56% vs ~77% of firm-months) and a single-quarter ROE is ~¼ the annual magnitude, so mixing would misrank firms.
> 
> 
> 
> I think using the standardized accounting panel makes a lot of sense. But aren't we missing a bunch from the ROW chars from JKP overview? Like what about the Stambaugh-Yuan ones, and the DHS ones? We really should be in a position where we can generate all factors both within an outside of the U.S. (or if we can't I'd love to know why)

I completely overlooked it in the message. We are indeed able to generate the portfolios for the rest of the world. I will correct the note. Also, please don't merge yet. I'm making a few last changes

PR #185, opened 2026-06-03
155 add unit golden tests for industry classification helpers

### LINKED ISSUE #155: Add unit + golden tests for industry classification helpers
## Gap

`tests/unit/` covers `ff_ind_class` (`test_ff_ind_class.py`) and `comp_hgics` (`test_comp_hgics.py`), but the rest of the industry-classification surface in `src/jkp/data/aux_functions.py` has no unit or golden coverage.

## Functions needing tests

- `crsp_industry` — daily panel of CRSP SIC/NAICS per permno from name-date spans
- `hgics_join` — merges national + global Compustat GICS daily panels (local preferred)
- `comp_sic_naics` — combines/reconciles Compustat SIC/NAICS (NA + Global) into daily series per gvkey
- `comp_industry` — merges daily GICS and SIC/NAICS into single daily Compustat industry file (gap-filled)

> Note: FF49 bucketing itself lives in `ff_ind_class` / `_parse_siccodes_file` and is already covered. These four functions only build/reconcile the raw SIC/NAICS/GICS daily panels that feed into it.

## Scope

For each function:

1. Unit tests in `tests/unit/test_<name>.py` covering:
   - Date-range expansion (`secinfostartdt`→`secinfoenddt`, `indfrom`→`indthru`) produces correct daily rows
   - Null / sentinel handling (`sic == 0` → null in CRSP; `gics is null` → `-999` in Compustat)
   - Coalesce precedence on full/outer joins (local GICS preferred over global; non-null SIC preferred in `comp_sic_naics`)
   - Gap-fill continuity in `comp_industry` (`aux_date = LEAD(date) − 1 day` logic)
   - Dedup + sort invariants on `(gvkey, date)` / `(permno, date)`

2. Golden parquet at `tests/golden/fixtures/<name>/` where the function produces a stable panel — generator script at `tests/golden/generate_<name>_golden.py` (see #140 for convention).

## Conventions

- Polars lazy throughout, `pl.col(...)` form — applies to `crsp_industry`, `hgics_join`
- `comp_sic_naics` and `comp_industry` are DuckDB/ibis SQL; test at parquet-in → run → parquet-out level. `test` group already pulls duckdb/ibis transitively.
- Use `make_dataframe`, `assert_series_equal`, `tolerance`, `test_paths` fixtures (see `tests/README.md`); `test_paths` keeps the `.ddb` file in the temp interim dir
- Markers: `unit` for fast tests, `regression` for golden comparisons

## Assignee

@miahdy

miahdy:
## Summary

Adds focused unit tests and golden regression fixtures for the four industry-related helpers in `aux_functions.py` that build daily SIC/NAICS/GICS panels:

| Function | Test module | Golden fixture |
|----------|-------------|----------------|
| `comp_sic_naics` | `tests/unit/test_comp_sic_naics.py` | `tests/golden/fixtures/comp_sic_naics/comp_other.parquet` |
| `hgics_join` | `tests/unit/test_hgics_join.py` | `tests/golden/fixtures/hgics_join/comp_hgics.parquet` |
| `comp_industry` | `tests/unit/test_comp_industry.py` | `tests/golden/fixtures/comp_industry/comp_ind.parquet` |
| `crsp_industry` | `tests/unit/test_crsp_industry.py` | `tests/golden/fixtures/crsp_industry/crsp_ind.parquet` |

Adds four golden generator scripts under `tests/golden/` plus shared input/stub modules so generators and tests use the same deterministic fixtures:

| Module | Role |
|--------|------|
| `generate_comp_sic_naics_golden.py` | Writes `comp_other.parquet`; imports `build_sic_naics_inputs` from `comp_sic_naics_inputs.py` |
| `generate_hgics_join_golden.py` | Writes `comp_hgics.parquet`; defines `build_hgics_inputs()` (golden test imports it) |
| `generate_comp_industry_golden.py` | Writes `comp_ind.parquet`; defines `build_comp_industry_inputs()`; uses `comp_industry_stubs.comp_industry_upstream_stubs` |
| `generate_crsp_industry_golden.py` | Writes `crsp_ind.parquet`; imports `build_permno0_input` from `crsp_industry_inputs.py` |
| `comp_sic_naics_inputs.py` | `sic_naics_frame`, `empty_sic_naics_frame`, `build_sic_naics_inputs` — shared by `test_comp_sic_naics` and its generator |
| `crsp_industry_inputs.py` | `permno0_frame`, `build_permno0_input` — shared by `test_crsp_industry` and its generator |
| `comp_industry_stubs.py` | `patch_comp_industry_upstream_stubs` (pytest) and `comp_industry_upstream_stubs` (generators) — no-op `comp_sic_naics` / `hgics_join` so only `comp_industry` SQL runs |

**No production code changes** — tests and fixtures only.

## Tests

**39 unit tests** across four modules (four golden `@pytest.mark.regression` tests; `test_hgics_join` also marks wall-clock independence as regression):

- **`test_comp_sic_naics.py`** (10 tests): NA/GL full outer join, SIC coalesce when NA is null (GL wins), NA precedence when both non-null, hard-coded `175650` row drop, `LPAD(gvkey, 6)`, half-open daily expansion (`closed="left"`), `DISTINCT ON` lowest-SIC tie-break, dedup/sort invariants, golden fixture.
- **`test_hgics_join.py`** (9 tests): NA-only / GL-only gvkeys (with `_FILLER_GVKEY` so empty sides do not break `comp_hgics`), national-wins coalesce, `-999` null-GICS sentinel passthrough, date-range expansion via `comp_hgics`, dedup/sort, dedup of deliberately duplicated upstream rows (monkeypatched `comp_hgics`), wall-clock independence (regression of #131 via `comp_hgics` sub-calls), golden fixture.
- **`test_comp_industry.py`** (12 tests): DuckDB gap-fill continuity (single/multi-span, per-`gvkey` partition; output schema checked in continuity test), full outer join + same-date coalesce, terminal-row / `COALESCE(LEAD…, date)` handling, dedup/sort invariants, stale `aux_comp_ind.ddb` cleanup, golden fixture. Upstream `comp_sic_naics` / `hgics_join` stubbed via `comp_industry_stubs`.
- **`test_crsp_industry.py`** (8 tests): `date_ranges` expansion, multi-span and overlapping spans, `sic == 0` → null, null `naics` preservation, dedup/sort, `permco` passthrough, golden fixture.

**Golden fixtures** (deterministic, small):

| Path | Rows | Columns |
|------|------|---------|
| `fixtures/comp_sic_naics/comp_other.parquet` | 9 | `gvkey`, `date`, `sic`, `naics` |
| `fixtures/hgics_join/comp_hgics.parquet` | 9 | `gvkey`, `date`, `gics` |
| `fixtures/comp_industry/comp_ind.parquet` | 8 | `gvkey`, `date`, `gics`, `sic`, `naics` |
| `fixtures/crsp_industry/crsp_ind.parquet` | 14 | `permno`, `permco`, `date`, `sic`, `naics` |

Regenerate a fixture after intentional output changes:

```bash
uv run python -m tests.golden.generate_comp_sic_naics_golden
uv run python -m tests.golden.generate_hgics_join_golden
uv run python -m tests.golden.generate_comp_industry_golden
uv run python -m tests.golden.generate_crsp_industry_golden
```

## Test plan

- [x] All 39 unit tests pass locally
- [x] Ruff lint and format checks pass
- [x] CI workflow runs successfully


FernandoRDLL:
Hey @miahdy can you please add the PR description? That will help me review and merge

miahdy:
> Hey @miahdy can you please add the PR description? That will help me review and merge

@FernandoRDLL Sorry about that. Description added.

tests/golden/generate_comp_sic_naics_golden.py
```diff
@@ -0,0 +1,120 @@
...
+
+def build_sic_naics_inputs(seed: int = 42) -> tuple[pl.DataFrame, pl.DataFrame]:
+    """Build deterministic NA and GL SIC/NAICS fixtures.
+
+    Scenarios covered by the returned (sic_naics_na, sic_naics_gl) pair:
+        001000 — NA-only, two datadates (2020-01-01 → 2020-01-04) producing 3 daily
+                 fill rows plus a trailing single-date row.
+        002000 — GL-only, one datadate (2020-06-15).
+        003000 — Both NA and GL on same datadate with different SICs; non-null
+                 higher SIC wins via DISTINCT ON (gvkey, date) ORDER BY sic.
```
FernandoRDLL: **Docstring contradicts production behavior (mechanism and direction).** Scenario `003000` says "higher SIC wins via DISTINCT ON (gvkey, date) ORDER BY sic", but in `comp_sic_naics` the NA and GL rows for the same `(gvkey, datadate)` meet in a single `FULL OUTER JOIN` row, so the survivor is decided by `COALESCE(sica, sicb)` — NA wins — and the result is sic=6020, the *lower* value. `DISTINCT ON` never sees two rows for this scenario. (Separately, the inner CTE's `ORDER BY ... sic DESC` is dead — DuckDB resolves `DISTINCT ON` with the outer `ORDER BY t.sic` ASC, which is what `test_distinct_on_picks_lowest_sic` correctly asserts.)

Risk: a maintainer who trusts this docstring could "fix" the SQL to make the higher SIC win and regenerate the golden — the regression test would move with the code and silently flip industry classifications. Please reword to: "both sources on same datadate → single joined row; COALESCE prefers the NA value (6020)."

tests/unit/test_comp_sic_naics.py
```diff
@@ -0,0 +1,308 @@
...
+
+        comp_sic_naics(self.paths)
+
+        result = pl.read_parquet(self.output_path)
+        assert result.height == 1
+        row = result.row(0, named=True)
+        assert row["sic"] == 2834
+        assert row["naics"] == 325412
+
+    def test_coalesce_prefers_non_null_sic(self) -> None:
```
FernandoRDLL, line 74: **This test cannot detect a flipped COALESCE precedence.** With NA sic=NULL and GL sic=4813, both `COALESCE(sica, sicb)` and a hypothetically reversed `COALESCE(sicb, sica)` return 4813 — the assertion passes either way. The NA-over-GL precedence when *both* sides are non-null is only pinned indirectly by the golden fixture (scenario 003000), where a failure is opaque.

Suggest adding a companion case with both sides non-null and different (e.g. NA sic=6020, GL sic=6021 on the same datadate) asserting the NA value survives, so precedence is covered by a named unit test rather than only the golden diff.

tests/unit/test_comp_industry.py
```diff
@@ -0,0 +1,535 @@
...
+            [date(2020, 1, 1), date(2020, 1, 5), date(2020, 6, 15)],
+            [10101010, 10101010, 20202020],
+        )
+        result = self._run(comp_other, comp_hgics)
+        assert result.unique(["gvkey", "date"]).height == result.height, (
+            f"Found duplicate (gvkey, date) rows: {result.height} total vs "
+            f"{result.unique(['gvkey', 'date']).height} unique"
+        )
+
+    def test_duplicate_input_rows_collapse(self) -> None:
```
FernandoRDLL: **Test exercises a path its own docstring calls unreachable.** The docstring states the `DISTINCT ON` dedup branch is "structurally unreachable in DuckDB 1.x … can never appear", and the assertion then only re-checks the generic `(gvkey, date)` uniqueness contract already covered by `test_unique_gvkey_date_invariant`. As written it can never catch a regression the sibling test misses.

Suggest either deleting it or folding the duplicated-input row into `test_unique_gvkey_date_invariant` — ~22 lines of maintenance weight otherwise.

tests/unit/test_comp_industry.py
```diff
@@ -0,0 +1,535 @@
...
+        assert result.height == 1, (
+            f"Expected 1 output row after stale-ddb recovery, got {result.height}"
+        )
+        row = result.row(0, named=True)
+        assert (row["sic"], row["naics"], row["gics"]) == (7372, 511210, 10101010), (
+            f"Expected codes (7372, 511210, 10101010), "
+            f"got {(row['sic'], row['naics'], row['gics'])}"
+        )
+        # The stale bytes were replaced by a real DuckDB database.
+        if self.ddb_path.exists():
```
FernandoRDLL: **The substantive assertion is silently skippable.** The check that the planted invalid bytes were replaced by a real DuckDB file is gated behind `if self.ddb_path.exists():`. If a duckdb/ibis version ever removes the on-disk DB at disconnect, the meaningful half of this test is skipped without failing, leaving only the trivial `height == 1` check.

Suggest asserting unconditionally on the recovery behavior, e.g.:

```python
if self.ddb_path.exists():
    assert self.ddb_path.read_bytes()[:23] != b"not a valid duckdb file", ...
```

i.e. drop the guard and instead assert `read_bytes()[:23] != b"not a valid duckdb file"` only after asserting the run succeeded — or at minimum note in the test why a missing file is also an acceptable outcome.

tests/unit/test_hgics_join.py
```diff
@@ -0,0 +1,309 @@
...
+        },
+    )
+
+
+# Filler gvkey used as a non-interfering "other-side" row in tests that want to
+# isolate behavior of one source. ``comp_hgics`` does not handle empty inputs
+# (it calls ``data[["indfrom"]].max()[0, 0] > END_DATE``, which raises on None),
+# so each side needs at least one row. The filler is at an unused gvkey/date so
+# it can be excluded from assertions via a ``.filter(gvkey != _FILLER_GVKEY)``.
+_FILLER_GVKEY = "999999"
```
FernandoRDLL, line 71: **Filler workaround masks a real production gap worth filing.** `_FILLER_GVKEY` exists solely because `comp_hgics` crashes on an empty input (`data[["indfrom"]].max()[0, 0] > END_DATE` raises on `None`), so every single-source test must carry a filler row and filter it back out of assertions.

The deeper fix is a guard for empty input in `aux_functions.comp_hgics` — then the filler dance disappears from 5 tests. Since this PR is tests-only, suggest filing a follow-up issue for the production guard and referencing it in this comment block so the workaround is traceable.

tests/unit/test_hgics_join.py
```diff
@@ -0,0 +1,309 @@
...
+        hgics_join(self.paths)
+
+        result = pl.read_parquet(self.output_path)
+        assert result.unique(["gvkey", "date"]).height == result.height, (
+            f"Found duplicate (gvkey, date) rows: {result.height} total vs "
+            f"{result.unique(['gvkey', 'date']).height} unique"
+        )
+        gvkeys = result["gvkey"].to_list()
+        dates = result["date"].to_list()
+        assert gvkeys == sorted(gvkeys), f"gvkeys not sorted ascending: {gvkeys}"
```
FernandoRDLL: **Copy-pasted invariant checks across modules — extract shared helpers.** This exact uniqueness check (`result.unique([...]).height == result.height` + formatted message) and the manual sorted-within-key loop below appear near-verbatim in `test_hgics_join.py`, `test_comp_industry.py`, and `test_crsp_industry.py` (uniqueness in 5 places, the sort loop in 3). A fix to the invariant logic must currently be replicated in every copy.

Suggest two small helpers in `tests/conftest.py`:

```python
def assert_unique_keys(df: pl.DataFrame, keys: list[str]) -> None: ...
def assert_sorted_by_keys(df: pl.DataFrame, group_col: str, date_col: str) -> None: ...
```

and call them from all four modules.

tests/unit/test_comp_sic_naics.py
```diff
@@ -0,0 +1,308 @@
...
+
+class TestCompSicNaics:
+    """Tests for ``comp_sic_naics``."""
+
+    @pytest.fixture(autouse=True)
+    def _setup(self, test_paths: DataPaths) -> None:
+        self.paths = test_paths
+        self.output_path = self.paths.interim_dir / "comp_other.parquet"
+
+    def test_na_only_gvkey(self) -> None:
```
FernandoRDLL, line 47: **Input schema is re-declared many times and can drift from the generator.** Each test in this module inlines the full `{gvkey: Utf8, datadate: Date, sic: Int64, naics: Int64}` schema dict (9+ repetitions; `test_crsp_industry.py` has the same pattern 8×), while the paired golden generator already encodes the same schema in `build_sic_naics_inputs`. Two definitions of the input contract can drift apart, so golden fixtures and unit tests could end up exercising differently-typed inputs.

Suggest a small frame-builder per module (like `_other_frame`/`_gics_frame` in `test_comp_industry.py`) — or better, one shared builder imported by both the test module and the generator — so the schema lives in exactly one place.

tests/unit/test_comp_industry.py
```diff
@@ -0,0 +1,535 @@
...
+        dates = result["date"].to_list()
+        assert gvkeys == sorted(gvkeys), f"gvkeys not sorted ascending: {gvkeys}"
+        for i in range(1, len(result)):
+            same_gvkey = gvkeys[i] == gvkeys[i - 1]
+            assert (not same_gvkey) or dates[i] >= dates[i - 1], (
+                f"Dates not sorted within gvkey {gvkeys[i]}: "
+                f"{dates[i - 1]} > {dates[i]} at rows {i - 1},{i}"
+            )
+
+    def test_output_schema(self) -> None:
```
FernandoRDLL: **Redundant full pipeline run + redundant assertion.** `test_output_schema` feeds inputs byte-identical to `test_gap_fill_continuity` and re-runs the entire DuckDB pipeline (fresh connection, `create_table` ×2, multi-statement SQL) only to assert columns/schema. Also, `dict(result.schema) == EXPECTED_SCHEMA` already implies `result.columns == list(EXPECTED_SCHEMA)`, so the first assertion is a strict subset of the second.

Suggest folding the schema-dict assertion into `test_gap_fill_continuity` (or sharing one class-scoped run) and dropping the redundant columns check.

tests/golden/generate_comp_sic_naics_golden.py
```diff
@@ -0,0 +1,120 @@
...
+        004000 — NA row with sic=NULL on one datadate; GL row with sic non-null
+                 on the same datadate → coalesce keeps non-null SIC.
+        175650 — Hard-coded dropped row (datadate=2005-12-31, naics IS NULL); a
+                 separate datadate (2006-06-30) for the same gvkey is retained.
+        500    — gvkey not zero-padded in input; output must be LPAD to '000500'.
+    """
+    del seed
```
FernandoRDLL: **Dead `seed` parameter in all four generators.** `build_*_inputs(seed=42)` immediately does `del seed` and builds hand-crafted literals — unlike `generate_rolling_golden.py`, where the seed actually drives `np.random.default_rng(seed)`. Keeping the parameter (and the tests calling with `seed=42`) suggests randomness that does not exist; a maintainer debugging a failure may waste time hunting for the RNG or trying "another seed" on fully deterministic fixtures.

Suggest dropping the parameter from all four `build_*_inputs` signatures and call sites (applies to `generate_hgics_join_golden.py`, `generate_comp_industry_golden.py`, `generate_crsp_industry_golden.py` too).

tests/unit/test_comp_industry.py
```diff
@@ -0,0 +1,535 @@
...
+class TestCompIndustry:
+    """Tests for ``comp_industry``."""
+
+    @pytest.fixture(autouse=True)
+    def _setup(self, test_paths: DataPaths, monkeypatch: pytest.MonkeyPatch) -> None:
+        """Bind paths and monkeypatch sub-calls so only ``comp_industry``'s SQL runs."""
+        self.paths = test_paths
+        self.output_path = self.paths.interim_dir / "comp_ind.parquet"
+        self.ddb_path = self.paths.interim_dir / "aux_comp_ind.ddb"
+        monkeypatch.setattr(aux_functions, "comp_sic_naics", lambda _paths: None)
```
FernandoRDLL: **No-op patches are duplicated between test and generator with no sync guarantee.** This `_setup` fixture stubs `comp_sic_naics`/`hgics_join`, and `generate_comp_industry_golden.py` independently re-implements the same stubbing in its `main()`. If the stub signatures or the patch targets ever diverge (e.g. one side updated after a production refactor), the golden would be generated by a different code path than the test exercises — and since the patches are never asserted to have been hit, a refactor that stops calling these functions would make the patches silently dead.

Suggest a tiny shared helper (e.g. `stub_upstream(monkeypatch_or_module)` importable by both) and/or a sentinel assertion that the stubs were actually invoked.

FernandoRDLL:
Hey @miahdy, I asked Claude to review the code. Please address the changes, then I'll make a final review and merge. Thank you for the PR!

miahdy:
Hi @FernandoRDLL, I have addressed and resolved all of Claude’s comments except for “Filler workaround masks a real production gap worth filing.” The PR description has also been updated. I’m not sure whether it would be best to open a separate issue for this. Could you please take a look at the rest of the generator and unit test code and let me know if there is anything else I should modify. Thanks!

tests/unit/test_comp_sic_naics.py
```diff
@@ -0,0 +1,215 @@
...
+        gl = sic_naics_frame(["004000"], [date(2019, 3, 1)], [4813], [517110])
+        _write_inputs(self.paths, na, gl)
+
+        comp_sic_naics(self.paths)
+
+        result = pl.read_parquet(self.output_path)
+        assert result.height == 1
+        row = result.row(0, named=True)
+        # Non-null SIC from GL wins because DISTINCT ON sorts sic DESC and
+        # COALESCE(sica, sicb) brings the GL value in when NA is null.
```
FernandoRDLL: This comment looks like a leftover from the earlier docstring round — it doesn't match what the SQL does here. In this scenario the NA and GL rows meet in a single `FULL OUTER JOIN` row, so `DISTINCT ON` never has a duplicate to pick from; `COALESCE(sica, sicb)` alone brings in the GL value. And the effective sort for `DISTINCT ON` is the outer `ORDER BY t.sic` (ascending) — the inner CTE's `ORDER BY sic DESC` is dead, which your own `test_distinct_on_picks_lowest_sic` demonstrates. Suggest trimming the comment to just the coalesce:

```suggestion
        # COALESCE(sica, sicb) brings the GL value in when NA is null.
```

tests/golden/comp_industry_stubs.py
```diff
@@ -0,0 +1,77 @@
...
+from unittest.mock import patch
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+    import pytest
+
+    from jkp.data.paths import DataPaths
+
+_PATCH_TARGET = "jkp.data.aux_functions"
```
FernandoRDLL: `_PATCH_TARGET` is never used — both `patch_comp_industry_upstream_stubs` and `comp_industry_upstream_stubs` import `aux_functions` directly and patch via `setattr`/`patch.object`. Suggest deleting it.

```suggestion
```

tests/unit/test_comp_industry.py
```diff
@@ -0,0 +1,506 @@
...
+        assert intermediates["naics"].null_count() == intermediates.height, (
+            f"Intermediate rows should have null naics, got "
+            f"{intermediates.height - intermediates['naics'].null_count()} non-null"
+        )
+        assert intermediates["gics"].null_count() == intermediates.height, (
+            f"Intermediate rows should have null gics, got "
+            f"{intermediates.height - intermediates['gics'].null_count()} non-null"
+        )
+        assert dict(result.schema) == EXPECTED_SCHEMA, (
+            f"Schema mismatch: expected {EXPECTED_SCHEMA}, got {dict(result.schema)}"
```
FernandoRDLL, line 200: The comment above `EXPECTED_SCHEMA` promises "column order and dtypes", but Python dict equality is order-insensitive, so a reordering of the output columns would still pass here. Simplest fix is to pin the order explicitly alongside the dtype check:

```suggestion
        assert result.columns == list(EXPECTED_SCHEMA), (
            f"Column order mismatch: expected {list(EXPECTED_SCHEMA)}, got {result.columns}"
        )
        assert dict(result.schema) == EXPECTED_SCHEMA, (
            f"Schema mismatch: expected {EXPECTED_SCHEMA}, got {dict(result.schema)}"
        )
```

FernandoRDLL (commented):
Thanks @miahdy — the previous round looks well addressed. I re-ran the 39 tests, lint/format, and regenerated all four golden fixtures locally from your generators; everything passes and the fixtures come out bit-identical. Three small leftovers inline below.

On the filler question: yes, I'd open a separate issue (answered in the open thread).

One optional note, no change requested: the shared input builders live in two places — dedicated `*_inputs.py` modules for `comp_sic_naics`/`crsp_industry`, but inside the generator scripts for `hgics_join`/`comp_industry`. Both layouts work; mentioning it only in case you'd like them uniform.

tests/unit/test_hgics_join.py
FernandoRDLL: Good question — this deserves its own issue rather than a fix in this tests-only PR. I've opened #192 for the underlying gap (`comp_hgics` crashing on an empty input frame) and assigned it to you. The `_FILLER_GVKEY` workaround is fine to keep here for now; #192's acceptance criteria include removing it once the guard is in. Maybe add a brief pointer to #192 next to the `_FILLER_GVKEY` comment so the link is discoverable from the test.

FernandoRDLL:
I've opened #192 for the underlying `comp_hgics` empty-input gap and assigned it to you — no change needed in this PR beyond optionally linking the issue from the test comment.

miahdy:
> Thanks @miahdy — the previous round looks well addressed. I re-ran the 39 tests, lint/format, and regenerated all four golden fixtures locally from your generators; everything passes and the fixtures come out bit-identical. Three small leftovers inline below.
> 
> On the filler question: yes, I'd open a separate issue (answered in the open thread).
> 
> One optional note, no change requested: the shared input builders live in two places — dedicated `*_inputs.py` modules for `comp_sic_naics`/`crsp_industry`, but inside the generator scripts for `hgics_join`/`comp_industry`. Both layouts work; mentioning it only in case you'd like them uniform.

Thank you @FernandoRDLL . I've applied the suggested changes and adjusted the code for lint/format. I will start working on the other issue.

FernandoRDLL (approved):

PR #188, opened 2026-06-03
Add unit, golden, and integration tests for quality_minus_junk

### LINKED ISSUE #187: [Feature]: Test coverage for quality_minus_junk / z_ranks / merge_qmj_to_world_data
### Feature Description

`quality_minus_junk` (aux_functions.py, ~114 lines of core QMJ scoring logic), its helper `z_ranks`, and `merge_qmj_to_world_data` have no test coverage. These implement rank-based within-(excntry, eom) z-scoring with subtle semantics that are easy to break silently in a refactor:

- min_stks threshold counts only finite values (nulls AND NaNs dropped by `is_not_nan`)
- sample std (ddof=1) on average-method ranks; all-equal or single-stock groups drop out
- `__evol = coalesce(2 * roeq_be_std, roe_be_std)` — 2x multiplier on the quarterly term only
- composites via `mean_horizontal` (skips nulls) but `__qmj` via raw `+` (null propagates)
- below-min_stks groups are retained in qmj.parquet with all-null qmj columns (benign due to downstream left join, but behavior worth locking)
- input must be unique on (excntry, eom, id): duplicates fan out across 16+3 full joins and panic at the Polars frame-length limit

### Use Case

Unit tests for the math and edge cases, a golden regression test on synthetic data calibrated to real-data summary statistics (null rates, gate pass rates, group sizes — no WRDS rows committed), and an integration test of the quality_minus_junk -> merge_qmj_to_world_data chain, so refactors of this code can be verified against locked behavior.

FernandoRDLL:
Closes #187.

## Simple explanation

`quality_minus_junk` (the QMJ score construction), its `z_ranks` helper, and `merge_qmj_to_world_data` had no tests. This PR adds 41 tests across three layers — unit tests for the math and corner cases, a golden regression test that locks the exact output on synthetic data, and an integration test of the qmj → world_data merge chain — so future refactors of this code can be verified against current behavior. No pipeline behavior changes; the only `src/` change is a comment.

## Technical details

- **Unit** (`tests/unit/test_quality_minus_junk.py`, 34 tests): `z_ranks` math parity vs hand-computed `(rank − mean)/std(ddof=1)` z-scores, ascending/descending direction, average-rank ties, min_stks threshold counting only finite values, NaN/null filtering, per-(excntry, eom) independence; the `c1` filter gate; `__evol = coalesce(2 * roeq_be_std, roe_be_std)` (2× on the quarterly term only); `mean_horizontal` composites skipping nulls vs raw-`+` null propagation in the final qmj; output schema and idempotency. Tests also lock that below-min_stks groups remain in `qmj.parquet` with all-null qmj columns (benign: downstream left join yields identical world_data either way).
- **Golden regression** (`tests/golden/test_quality_minus_junk_golden.py` + `generate_qmj_golden.py` + committed fixture, 231 rows): synthetic input calibrated to summary statistics of the real interim `world_data_-1.parquet` — per-variable medians/IQR scales, null rates (16–54%), gate pass rates (~82–88%), `ret_exc` null rate, Int32 flag dtypes, and a deliberately below-min_stks country (real q5 group size ≈ 5) locking thin-group behaviour. **Only aggregates were used for calibration; no WRDS rows are embedded in the repo.** Since qmj is rank-based, calibration matters mainly through null patterns, group sizes, and gate mix, which now mirror production. Supports `--regen-golden`.
- **Integration** (`tests/integration/test_qmj_pipeline.py`, 6 tests): `quality_minus_junk` → `merge_qmj_to_world_data` chain; left-join semantics (gate-failing rows kept with null qmj), height/dedup on `(id, eom)`, sort order, parity of merged qmj columns vs standalone `qmj.parquet`.
- **src change (comment only)**: note above the join loop in `quality_minus_junk` documenting the `(excntry, eom, id)` uniqueness assumption — duplicates would fan out multiplicatively across the 16+3 full joins and panic at the Polars frame-length limit; uniqueness holds upstream by construction.

@capellini — please take a look at the synthetic-data calibration in `tests/golden/generate_qmj_golden.py` (sampling scheme, null rates, and the no-WRDS-data constraint).

## Test plan

- [x] `PYTHONPATH=src uv run --no-sync python -m pytest tests/unit/test_quality_minus_junk.py` — 34 passed
- [x] `PYTHONPATH=src uv run --no-sync python -m pytest tests/golden/test_quality_minus_junk_golden.py` — passes vs committed fixture
- [x] `PYTHONPATH=src uv run --no-sync python -m pytest tests/integration/test_qmj_pipeline.py` — 6 passed
- [x] Full unit suite: 514 passed
- [x] `ruff check` + `ruff format --check` clean
- [x] Standalone generator (`python -m tests.golden.generate_qmj_golden`) reproduces the committed fixture


FernandoRDLL:
@capellini Can you please review that I created synthetic data correctly?

tests/golden/generate_qmj_golden.py
```diff
@@ -0,0 +1,298 @@
...
+                    "exch_main": int(flags["exch_main"][i]),
+                    "ret_exc": None if ret_null[i] else float(ret_exc[i]),
+                    "me": float(me[i]),
+                }
+                for var in _FLOAT_VARS:
+                    row[var] = None if nulls[var][i] else float(values[var][i])
+                rows.append(row)
+
+    # --- Corner case: a few NaN values in single z-vars ---
+    # Use id=100 (USA) / id=300 (FRA); they already appear above — we
```
capellini: This comment mentions that the USA ID is 100, but 150 is used below.

tests/golden/test_quality_minus_junk_golden.py
```diff
@@ -0,0 +1,67 @@
...
+import pytest
+
+from jkp.data.aux_functions import quality_minus_junk
+from tests.golden.generate_qmj_golden import build_qmj_world_data_input
+
+FIXTURE_PATH = Path(__file__).parent / "fixtures" / "qmj" / "qmj.parquet"
+
+# rtol=1e-9 for floats: z-ranks involve rank/mean/std, all exact floating-point
+# ops on the same seed; cross-platform IEEE-754 guarantees this tight tolerance.
+_FLOAT_RTOL = 1e-9
```
capellini: [This is more for recording purposes - feel free to mark resolved immediately with no action]
While technically correct for a given Polars version, since rank/mean/std reductions can reorder across Polars versions, we may want to reduce to 1e-6. But let's keep it as 1e-9 for now. We can always loosen the restriction if it becomes a problem.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+
+    def test_ascending_hand_computed(self, tolerance: Any) -> None:
+        """z-scores match hand-computed (rank - mean)/std ddof=1 for ascending sort."""
+        vals = [1.0, 3.0, 5.0, 7.0, 9.0]
+        df = _z_frame(vals)
+        result = z_ranks(df, "x", 2, "ascending")
+        result = result.sort("id")
+
+        expected = _hand_z(vals)
+        import numpy as np
```
capellini: Move/consolidate import to the top of the file.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+        np.testing.assert_allclose(result["z_x"].to_list(), expected, **tolerance.STANDARD)
+
+    def test_descending_reverses_sign(self, tolerance: Any) -> None:
+        """Descending z-scores are the negation of ascending z-scores."""
+        vals = [1.0, 3.0, 5.0, 7.0, 9.0]
+        df = _z_frame(vals)
+        asc = z_ranks(df, "x", 2, "ascending").sort("id")
+        desc = z_ranks(df, "x", 2, "descending").sort("id")
+
+        import numpy as np
```
capellini: Move/consolidate import to the top of the file.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+
+    def test_ties_get_equal_rank_and_z(self, tolerance: Any) -> None:
+        """Tied values share average rank and identical z-score."""
+        vals = [1.0, 2.0, 2.0, 4.0]
+        df = _z_frame(vals)
+        result = z_ranks(df, "x", 2, "ascending").sort("id")
+
+        z_list = result["z_x"].to_list()
+        # ids 2 and 3 are tied at 2.0 → equal z
+        import numpy as np
```
capellini: Move/consolidate import to the top of the file.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+    def test_null_row_filtered_same_as_nan(self) -> None:
+        """None (null) row is filtered identically to NaN."""
+        vals = [1.0, 3.0, None]
+        df = _z_frame(vals)
+        result = z_ranks(df, "x", 2, "ascending")
+        assert result.height == 2
+
+    def test_two_countries_independent(self, tolerance: Any) -> None:
+        """Two countries with identical values produce identical z-scores per country."""
+        import numpy as np
```
capellini: Move/consolidate import to the top of the file.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+        qmj_growth = z_ranks_py(growth_vals)
+        qmj_safety = z_ranks_py(safety_vals)
+
+        qmj_raw = [
+            (p + g + s) / 3.0 for p, g, s in zip(qmj_prof, qmj_growth, qmj_safety, strict=True)
+        ]
+        qmj_expected = z_ranks_py(qmj_raw)
+
+        result_sorted = result.sort("id")
+        import numpy as np
```
capellini: Move/consolidate import to the top of the file.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+
+        result_a = self._run(df_a_only).sort("id")
+        # Re-run with both countries (need fresh paths → write again)
+        p_ab = self.paths.interim_dir / "test_msf.parquet"
+        df_ab.write_parquet(p_ab)
+        quality_minus_junk(self.paths, p_ab, 3)
+        result_ab = pl.read_parquet(self.paths.interim_dir / "qmj.parquet")
+        result_ab_a = result_ab.filter(pl.col("excntry") == "US").sort("id")
+
+        import numpy as np
```
capellini: Move/consolidate import to the top of the file.

tests/integration/test_qmj_pipeline.py
```diff
@@ -0,0 +1,195 @@
...
+
+    return {
+        "df_input": df_input,
+        "world_data": world_data,
+        "qmj": qmj,
+        "paths": paths,
+    }
+
+
+def _setup_dirs(base: Path) -> None:
```
capellini: This reimplements directory setup by hand. `_setup_dirs()` duplicates the body of the `temp_data_dir` fixture verbatim, and `DataPaths(base_dir=...)` is constructed manually and in `test_duplicate_id_eom_collapses`. Per `tests/README.md` (the "Testing Pipeline Functions That Take paths: DataPaths" section), tests should request the `test_paths` fixture rather than constructing `DataPaths` manually. The manual construction pattern is being phased out, and the hand-rolled copy will silently drift from `DataPaths/temp_data_dir` if the layout ever changes.

The blocker is scope: `pipeline_outputs` is `scope="class"` and `test_paths` is function-scoped. I suggest making `pipeline_outputs` function-scoped (the chain runs in ~0.06s, so re-running per test is negligible) and requesting `test_paths`. That lets you delete `_setup_dirs` entirely. `test_duplicate_id_eom_collapses` is already function-scoped and can just take `test_paths` directly.

tests/golden/generate_qmj_golden.py
```diff
@@ -0,0 +1,298 @@
...
+            "exch_main": pl.Int32,
+            "ret_exc": pl.Float64,
+            "me": pl.Float64,
+            **dict.fromkeys(_FLOAT_VARS, pl.Float64),
+        },
+    )
+    return df
+
+
+def _make_paths(base_dir: str) -> DataPaths:
```
capellini, line 267: This hardcodes the pipeline layout instead of leaning on `DataPaths`.

As a standalone script it can't use the `test_paths` fixture, which is fine. But it only ever writes to `interim_dir`, so the `raw/` and `processed/*` dirs created here are unused.

I suggest trimming to just what's needed:

```python
def _make_paths(base_dir: str) -> DataPaths:
    paths = DataPaths(base_dir=Path(base_dir))
    paths.interim_dir.mkdir(parents=True, exist_ok=True)
    return paths
```

That keeps the layout knowledge in `DataPaths` rather than duplicating it here.

tests/integration/test_qmj_pipeline.py
```diff
@@ -0,0 +1,195 @@
...
+import pytest
+
+from jkp.data.aux_functions import merge_qmj_to_world_data, quality_minus_junk
+from jkp.data.paths import DataPaths
+from tests.golden.generate_qmj_golden import build_qmj_world_data_input
+
+# ---------------------------------------------------------------------------
+# Tolerance (mirrors ToleranceSpec.STANDARD)
+# ---------------------------------------------------------------------------
+_RTOL = 1e-6
```
capellini: These module-level tolerance constants duplicate ToleranceSpec (the comment on line 22 even says they "mirror ToleranceSpec.STANDARD"). Per the "Using the ToleranceSpec Fixture" section of `tests/README.md`, tests should request the tolerance fixture and use `tolerance.STANDARD` rather than hand-copying its values (which silently drift if the spec changes).

tests/golden/test_quality_minus_junk_golden.py
capellini: Same as the integration test: this hand-rolls a tolerance constant instead of using the `tolerance` fixture. Tying the `1e-9` vs `1e-6` discussion (other comment) to `ToleranceSpec` would keep the choice in one place. Not blocking, but worth aligning with the README's tolerance-fixture guidance.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+
+        np.testing.assert_allclose(
+            result_a["qmj"].to_list(),
+            result_ab_a["qmj"].to_list(),
+            rtol=1e-6,
```
capellini: Inconsistent with the rest of this file. Every other comparison in this file uses `**tolerance.STANDARD`, but this one hardcodes `rtol=1e-6, atol=1e-10` (the same numbers). Use `**tolerance.STANDARD` here too for consistency.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+        for eom in eoms:
+            grp = _group_frame(excntry=excntry, eom=eom, n=n_per_group, overrides=overrides)
+            # shift ids to be globally unique
+            grp = grp.with_columns(pl.col("id") + id_offset)
+            id_offset += n_per_group
+            frames.append(grp)
+    return pl.concat(frames)
+
+
+def _write_world(paths: DataPaths, df: pl.DataFrame) -> pl.Path:  # type: ignore[name-defined]
```
capellini: `pl.Path` isn't a real type. Polars has no Path, which is why the `# type: ignore[name-defined]` is needed to silence it. It should be `pathlib.Path` (add `from pathlib import Path` and `annotate -> Path`), which lets you drop the `type: ignore`.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+    @pytest.fixture(autouse=True)
+    def _setup(self, test_paths: DataPaths) -> None:
+        self.paths = test_paths
+
+    def _run(self, df: pl.DataFrame, min_stks: int = 3) -> pl.DataFrame:
+        p = _write_world(self.paths, df)
+        quality_minus_junk(self.paths, p, min_stks)
+        return pl.read_parquet(self.paths.interim_dir / "qmj.parquet")
+
+    def test_evol_uses_2x_roeq_multiplier(self) -> None:
```
capellini: This test doesn't actually exercise the 2× multiplier. I verified by mutating the source `coalesce(2 * roeq_be_std, roe_be_std)` to `coalesce(1 * ...` and this test still passes (only the golden fixture catches it) because the assertion checks ordering only (safety[0] > safety[-1]), and z-ranking is invariant to positive scaling, so a uniform 2x across all stocks in the group produces identical ranks to 1x. The same applies to `test_evol_fallback_to_roe_be_std` below: its `null_count < n` assertion would pass against a swapped coalesce priority too (mutating to `coalesce(roe_be_std, 2 * roeq_be_std)` is also caught only by the golden).

So both the multiplier and the fallback priority are currently protected only by the golden tests. To unit test the 2x (and the coalesce order), build a single (excntry, eom) group that mixes roeq-null stocks (which fall back to `roe_be_std`) with roeq-non-null stocks (which use `2 * roeq_be_std`), and assert the cross-ordering between the two subsets. That ordering is the only thing the factor of 2 and the coalesce priority actually change.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+            # prof and growth can still be non-null
+            assert row1["qmj_prof"][0] is not None or row1["qmj_growth"][0] is not None
+
+    def test_hand_computed_pipeline_single_group(self, tolerance: Any) -> None:
+        """Hand-compute full pipeline for a small group and compare qmj.
+
+        Uses 4 stocks with distinct values, computes z-ranks per var,
+        composites, z-rank composites, mean, final z-rank in plain Python.
+        """
+        import math as _math
```
capellini: We're already importing math in this file, so we can remove this line.

tests/unit/test_quality_minus_junk.py
```diff
@@ -0,0 +1,760 @@
...
+    def test_output_dtypes(self) -> None:
+        """eom is Date; qmj_{prof,growth,safety} and qmj are Float64."""
+        result = self._run(_group_frame(n=5))
+        assert result.schema["eom"] == pl.Date
+        for col in ["qmj_prof", "qmj_growth", "qmj_safety", "qmj"]:
+            assert result.schema[col] == pl.Float64, f"{col} should be Float64"
+
+    def test_idempotency(self) -> None:
+        """Running quality_minus_junk twice produces identical output."""
+        from polars.testing import assert_frame_equal as pl_assert_frame_equal
```
capellini: Consolidate imports at the top of the file.

capellini (commented):
Thanks Fernando -- this is a thorough, well-reasoned addition. I verified the suite end-to-end (ran all three files in a clean worktree, regenerated the fixture from scratch): 41 tests pass, the golden fixture reproduces byte-identically, and all three files are picked  up by CI (pytest tests/, no marker/path filter).

On your specific question — yes, the synthetic data is calibrated correctly. The things I checked:

- No WRDS data embedded. The generator holds only aggregate statistics (medians, IQR scales, null/gate rates). The committed qmj.parquet contains only synthetic sequential ids, USA/FRA/ITA, 2020 dates, and computed z-scores. No raw rows, no real identifiers.
- Fully deterministic & reproducible. Seeded default_rng(42): regenerating the fixture independently matched the committed one exactly.

Most inline comments are minor and don't block correctness:
- A stale comment (id 100 vs 150) and a tolerance note (recording-only).
- Import consolidation cleanup.

The two worth addressing are the DataPaths convention and the __evol test gap. The rest are fine as follow-ups. Happy to approve once those two are sorted.

tests/golden/generate_qmj_golden.py
FernandoRDLL: Good catch — comment was stale. Fixed in 4255cfc to say id=150/350 and explain they sit outside the base id ranges.

tests/golden/test_quality_minus_junk_golden.py
FernandoRDLL: Took this a step further in 4255cfc: the 1e-9 now lives as `ToleranceSpec.GOLDEN` in conftest with a note to loosen it there (toward STANDARD) if Polars reductions ever reorder, so the choice is in one place.

tests/integration/test_qmj_pipeline.py
FernandoRDLL: Done in 4255cfc — `pipeline_outputs` is now function-scoped on `test_paths`, `_setup_dirs` is gone, and `test_duplicate_id_eom_collapses` takes `test_paths` directly.

tests/golden/generate_qmj_golden.py
FernandoRDLL: Trimmed as suggested in 4255cfc — `_make_paths` now only creates `interim_dir` via `DataPaths`.

tests/integration/test_qmj_pipeline.py
FernandoRDLL: Done in 4255cfc — constants removed, `test_qmj_values_match_standalone` now uses `**tolerance.STANDARD`.

tests/golden/test_quality_minus_junk_golden.py
FernandoRDLL: Done in 4255cfc — added `ToleranceSpec.GOLDEN` (rtol=1e-9, atol=0) to conftest and the golden test now uses `**tolerance.GOLDEN`, so the 1e-9 vs 1e-6 choice lives in ToleranceSpec.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Fixed in 4255cfc — uses `**tolerance.STANDARD` now.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Fixed in 4255cfc — `pathlib.Path` with the import at the top, `type: ignore` dropped.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Great analysis, thanks. Rewrote both tests in 4255cfc around a single mixed group (roeq-non-null and roeq-null stocks) and asserted the cross-ordering: stock 1 (2*0.6=1.2) vs stock 2 (fallback 1.0) catches the 2x, and stock 4 (roeq=0.2, roe=5.0 → evol 0.4) vs stock 3 (fallback 0.5) catches the coalesce priority. Verified by mutating the source both ways — `1 * roeq` fails the multiplier test, swapped coalesce fails both.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Removed in 4255cfc.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Consolidated to the top of the file in 4255cfc.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Consolidated to the top of the file in 4255cfc.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Consolidated to the top of the file in 4255cfc.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Consolidated to the top of the file in 4255cfc.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Consolidated to the top of the file in 4255cfc.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Consolidated to the top of the file in 4255cfc.

tests/unit/test_quality_minus_junk.py
FernandoRDLL: Consolidated to the top of the file in 4255cfc.

FernandoRDLL:
@theisij I think we're good to merge

theisij (approved):

PR #196, opened 2026-06-05
perf: rewrite comp_industry chain in lazy Polars (2.5x, output-identical)

### LINKED ISSUE #195: [Feature]: comp_industry spends ~9 min materializing daily industry panels through disk-backed DuckDB
## Problem

`comp_industry` (with its helpers `comp_hgics`, `hgics_join`, `comp_sic_naics`) is one of the largest compute hotspots in `jkp build` (~8m47s in production run, ~3% of build time), and its implementation carries structural overhead unrelated to the result:

- The final merge runs in a **disk-backed** DuckDB database (`interim/aux_comp_ind.ddb`, ~16 GB on GPFS), materializing six intermediate tables for a single linear query.
- The gap-fill machinery (`CROSS JOIN LATERAL generate_series` + `UNION` + `DISTINCT ON`) is provably redundant: the `UNION` dedup and `DISTINCT ON` never collapse anything (key sets are disjoint), and the gap left-join only ever attaches values to the gap-start day, which already exists in the joined panel — interior gap days always come out null.
- The helper stages (`comp_hgics` ×2, `hgics_join`, `comp_sic_naics`) eagerly `collect()` multi-hundred-million-row daily panels in RAM before writing.

The daily output contract must stay: downstream/external users may rely on the daily `comp_ind.parquet`, so output values cannot change.

## Proposal

Rewrite the chain as a single lazy Polars plan with streaming `sink_parquet` writes, preserving the exact output (including legacy edge-case behavior: null-date rows dropped, null codes on gap-interior days), and guard equivalence with a test that runs the legacy SQL verbatim against the rewrite.

FernandoRDLL:
Closes #195

## Simple explanation

`comp_industry` built the daily Compustat industry panel through a disk-backed DuckDB pipeline that materialized six intermediate tables (a ~16 GB scratch `.ddb` on disk) and ran redundant dedup/gap-fill steps. This PR rewrites the chain as a single lazy Polars plan with streaming writes. Same output, 2.5x faster, no scratch file.

## Technical details

- `comp_industry`: DuckDB SQL → lazy Polars. Two provable no-ops removed: the `UNION` dedup + `DISTINCT ON` (key sets of continuous and gap rows are disjoint) and the gap-fill value join (only the gap-start day — already present in the joined panel — ever carried values; interior days are null). The rewrite emits joined rows plus null-coded interior gap days directly.
- Legacy SQL silently dropped null-date rows (from GICS records with null `indfrom`): both `WHERE date <> aux_date` and `WHERE date = aux_date` evaluate to NULL for them. Replicated with an explicit `.filter(pl.col("date").is_not_null())` and a comment.
- `comp_hgics`, `hgics_join`, `comp_sic_naics`: eager `collect().write_parquet()` → streaming `sink_parquet`; `comp_hgics` also goes fully lazy (`scan_parquet`, `pl.max_horizontal` for the terminal-date cap, same semantics).
- Polars CSE caches the shared join subplan (verified via `explain()`), so the panel join executes once despite two consumers.

Validated on full production data (Yale, 128-core node, job 1016196):

| | before (run 817992) | after |
|---|---|---|
| `COMP_INDUSTRY` chain | 8m 47s | 3m 31s |
| `comp_ind.parquet` | 634,366,645 rows | row-identical |
| scratch `aux_comp_ind.ddb` | 16 GB | none |

All five stage outputs (`g_hgics`, `na_hgics`, `comp_hgics`, `comp_other`, `comp_ind`) compared row-identical against the production run.

## Test plan

- New `tests/unit/test_comp_industry.py`: runs the **legacy SQL verbatim** (DuckDB) against the rewrite on synthetic fixtures covering gaps, open-ended intervals, null `indfrom`, single-source gvkeys, and misaligned GICS/SIC dates; asserts row-identical output plus a no-duplicate-keys guard.
- Full unit suite passes (605 tests); ruff check + format clean.

### DIFF
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -2901,7 +2901,7 @@ def comp_hgics(paths: DataPaths, lib):
         },
     }
     data = pl.scan_parquet(file_paths["raw data"][lib])  # .sort(['gvkey', 'indfrom'])
-    if data.select(pl.len()).collect().item() == 0:
+    if data.limit(1).collect().is_empty():
         warnings.warn(
             f"comp_hgics: {lib} GICS input is empty "
             f"({file_paths['raw data'][lib]}); "
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -2900,8 +2900,8 @@ def comp_hgics(paths: DataPaths, lib):
             "global": paths.interim_dir / "g_hgics.parquet",
         },
     }
-    data = pl.read_parquet(file_paths["raw data"][lib])  # .sort(['gvkey', 'indfrom'])
-    if data.height == 0:
+    data = pl.scan_parquet(file_paths["raw data"][lib])  # .sort(['gvkey', 'indfrom'])
+    if data.select(pl.len()).collect().item() == 0:
         warnings.warn(
             f"comp_hgics: {lib} GICS input is empty "
             f"({file_paths['raw data'][lib]}); "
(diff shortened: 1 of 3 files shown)


FernandoRDLL:
@claude please review the diff and make suggestions if necessary

FernandoRDLL:
Applied the review suggestions in 77e4b7e: added an equivalence fixture where gvkey 000007 mixes a null-indfrom GICS record with real-dated records (pinning the early null-date filter vs the legacy LEAD-over-unfiltered-partition behavior), and the comp_hgics emptiness check now uses `data.limit(1).collect().is_empty()` instead of a `len()` collect. Full unit suite (696) + ruff green.

theisij (approved):

PR #204, opened 2026-06-10
Fix comp_hgics crash on empty input (#192)

### LINKED ISSUE #192: comp_hgics crashes on empty input frame (missing empty-input guard in indthru fill)
### Which script does this issue concern?

`src/jkp/data/aux_functions.py` (`comp_hgics`)

### Error category

Pipeline crash on a degenerate-but-valid input

### Description

`comp_hgics` raises a `TypeError` when its raw input parquet (`comp_hgics_na.parquet` or `comp_hgics_gl.parquet`) contains zero rows. The crash comes from the open-ended `indthru` fill:

```python
indthru_date = (
    pl.lit(data[["indfrom"]].max()[0, 0])
    if data[["indfrom"]].max()[0, 0] > END_DATE
    else pl.lit(END_DATE)
)
```

On an empty frame, `data[["indfrom"]].max()[0, 0]` is `None`, and `None > END_DATE` raises:

```
TypeError: '>' not supported between instances of 'NoneType' and 'datetime.date'
```

The root cause is a missing empty-input guard: the max-vs-`END_DATE` comparison assumes at least one row. If either the NA or GL GICS download ever comes back empty (e.g. a WRDS subscription that lacks one region, or a filtered test/debug run), `hgics_join` — which calls `comp_hgics(paths, "global")` and `comp_hgics(paths, "national")` unconditionally — crashes the pipeline instead of producing an empty (but correctly typed) daily panel for that side.

**Downstream symptom:** the unit tests in PR #185 must inject a `_FILLER_GVKEY` row into whichever side would otherwise be empty, and filter it back out of every assertion, purely to keep `comp_hgics` from crashing. Once this is fixed, that workaround can be removed (see the open review thread on `tests/unit/test_hgics_join.py`).

### Steps to reproduce

```python
import polars as pl
from jkp.data.aux_functions import comp_hgics
from jkp.data.paths import DataPaths

paths = DataPaths(base_dir=...)  # any temp dir
empty = pl.DataFrame(
    schema={"gvkey": pl.Utf8, "indfrom": pl.Date, "indthru": pl.Date, "gics": pl.Int64}
)
(paths.interim_dir / "raw_data_dfs").mkdir(parents=True, exist_ok=True)
empty.write_parquet(paths.interim_dir / "raw_data_dfs" / "comp_hgics_na.parquet")
comp_hgics(paths, "national")  # TypeError
```

### Expected behavior

An empty input should produce an empty output parquet with the expected `{gvkey, date, gics}` schema, so `hgics_join` can full-join it against the non-empty side. Concretely: guard the `indthru_date` computation (e.g. treat a `None` max as "use `END_DATE`") or short-circuit on `data.height == 0`.

### Acceptance criteria

- `comp_hgics` handles a 0-row input without raising and writes a typed empty daily panel.
- `hgics_join` works when exactly one of NA/GL is empty.
- The `_FILLER_GVKEY` workaround in `tests/unit/test_hgics_join.py` is removed; single-source tests pass an empty frame for the other side directly.

miahdy:
## Summary

Fixes #192 by adding an empty-input guard to `comp_hgics` so a 0-row NA or GL GICS parquet no longer crashes the industry pipeline. When input is empty, the function now writes a correctly typed empty daily panel (`{gvkey, date, gics}`) and returns early, allowing `hgics_join` to proceed with the non-empty side.

Also removes the `_FILLER_GVKEY` test workaround introduced in #185 and adds direct coverage for empty-input and single-side-empty join behavior.

## Change Type

- [ ] Documentation only

## Methodological Impact

None. This is a robustness fix for degenerate-but-valid inputs. Normal non-empty GICS processing is unchanged.

## Data Impact

None for standard WRDS runs with both NA and GL GICS data present. The fix only affects cases where one side would previously have crashed the pipeline (e.g. missing regional subscription, filtered/debug runs, or empty upstream downloads).

## Tests

- Added `test_empty_input_produces_typed_empty_output` in `tests/unit/test_comp_hgics.py`.
- Removed `_FILLER_GVKEY` / `_filler_frame()` workaround from `tests/unit/test_hgics_join.py`; single-source tests now pass a typed empty frame for the unused side.
- Added `test_empty_na_side` and `test_empty_gl_side` in `tests/unit/test_hgics_join.py` to verify `hgics_join` works when exactly one of NA/GL is empty.

Run locally:

```bash
uv run pytest tests/unit/test_comp_hgics.py tests/unit/test_hgics_join.py -v -o "addopts="
uv run ruff check src/jkp/data/aux_functions.py tests/unit/test_comp_hgics.py tests/unit/test_hgics_join.py
uv run ruff format --check src/jkp/data/aux_functions.py tests/unit/test_comp_hgics.py tests/unit/test_hgics_join.py
```

## Checklist

## Additional Notes

**Root cause:** `comp_hgics` compared `data[["indfrom"]].max()[0, 0] > END_DATE` before checking for empty input. On a 0-row frame, the max is `None`, which raised `TypeError: '>' not supported between instances of 'NoneType' and 'datetime.date'`.

**Fix:** After reading the input parquet, short-circuit when `data.height == 0` and write an empty output parquet with schema `{gvkey: Utf8, date: Date, gics: Int64}`.

**Files changed:**
- `src/jkp/data/aux_functions.py` — empty-input guard in `comp_hgics`
- `tests/unit/test_comp_hgics.py` — empty-input unit test
- `tests/unit/test_hgics_join.py` — remove filler workaround; add single-side-empty tests

### DIFF
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -6,6 +6,7 @@ import os
 import re
 import shutil
 import time
+import warnings
 from datetime import date
 from math import exp, sqrt
 from pathlib import Path
@@ -2378,18 +2379,22 @@ def comp_hgics(paths: DataPaths, lib):
     }
     data = pl.read_parquet(file_paths["raw data"][lib])  # .sort(['gvkey', 'indfrom'])
     if data.height == 0:
-        pl.DataFrame(schema={"gvkey": pl.Utf8, "date": pl.Date, "gics": pl.Int64}).write_parquet(
-            file_paths["output"][lib]
+        warnings.warn(
+            f"comp_hgics: {lib} GICS input is empty "
+            f"({file_paths['raw data'][lib]}); "
+            "downstream comp_ind.parquet will have null GICS for all firms "
+            "unless the other side (national/global) provides coverage.",
+            stacklevel=2,
         )
-        return
     data = data.with_columns(
         gics=pl.when(col("gics").is_null()).then(-999).otherwise(col("gics")),
         n=pl.len().over("gvkey"),
         n_aux=pl.cum_count("gvkey").over("gvkey"),
     )
+    max_indfrom = data[["indfrom"]].max()[0, 0]
     indthru_date = (
-        pl.lit(data[["indfrom"]].max()[0, 0])
-        if data[["indfrom"]].max()[0, 0] > END_DATE
+        pl.lit(max_indfrom)
+        if max_indfrom is not None and max_indfrom > END_DATE
         else pl.lit(END_DATE)
     )
     c1 = col("n") == col("n_aux")
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -2377,6 +2377,11 @@ def comp_hgics(paths: DataPaths, lib):
         },
     }
     data = pl.read_parquet(file_paths["raw data"][lib])  # .sort(['gvkey', 'indfrom'])
+    if data.height == 0:
+        pl.DataFrame(schema={"gvkey": pl.Utf8, "date": pl.Date, "gics": pl.Int64}).write_parquet(
+            file_paths["output"][lib]
+        )
+        return
     data = data.with_columns(
         gics=pl.when(col("gics").is_null()).then(-999).otherwise(col("gics")),
         n=pl.len().over("gvkey"),
--- a/tests/unit/test_comp_hgics.py
+++ b/tests/unit/test_comp_hgics.py
@@ -64,18 +64,45 @@ class TestCompHgics:
     """Tests for `comp_hgics`."""
 
     def test_empty_input_produces_typed_empty_output(self, test_paths: DataPaths) -> None:
-        """A 0-row input must not crash and must write a typed empty parquet."""
+        """A 0-row input must not crash, must warn, and must write a typed empty parquet."""
         raw_data_dfs = test_paths.interim_dir / "raw_data_dfs"
         raw_data_dfs.mkdir(parents=True, exist_ok=True)
         pl.DataFrame(
             schema={"gvkey": pl.Utf8, "indfrom": pl.Date, "indthru": pl.Date, "gics": pl.Int64}
         ).write_parquet(raw_data_dfs / "comp_hgics_na.parquet")
 
-        comp_hgics(test_paths, "national")
+        with pytest.warns(UserWarning, match="national GICS input is empty"):
+            comp_hgics(test_paths, "national")
 
         output = pl.read_parquet(test_paths.interim_dir / "na_hgics.parquet")
         assert output.height == 0, f"Expected 0 rows, got {output.height}"
-        assert output.schema == {"gvkey": pl.Utf8, "date": pl.Date, "gics": pl.Int64}
+        assert set(output.columns) == {"gvkey", "date", "gics"}, (
+            f"Expected columns {{gvkey, date, gics}}, got schema {output.schema}"
+        )
+
+    def test_all_null_indfrom_does_not_crash(self, test_paths: DataPaths) -> None:
+        """Non-empty input with all-null indfrom must not raise TypeError."""
+        raw_data_dfs = test_paths.interim_dir / "raw_data_dfs"
+        raw_data_dfs.mkdir(parents=True, exist_ok=True)
+        pl.DataFrame(
+            {
+                "gvkey": ["001000"],
+                "indfrom": [None],
+                "indthru": [None],
+                "gics": [10101010],
+            },
+            schema={
+                "gvkey": pl.Utf8,
+                "indfrom": pl.Date,
+                "indthru": pl.Date,
+                "gics": pl.Int64,
+            },
+        ).write_parquet(raw_data_dfs / "comp_hgics_na.parquet")
+
+        comp_hgics(test_paths, "national")
+
+        output = pl.read_parquet(test_paths.interim_dir / "na_hgics.parquet")
+        assert set(output.columns) == {"gvkey", "date", "gics"}
 
     @pytest.mark.regression
     def test_independent_of_wall_clock(
--- a/tests/unit/test_comp_hgics.py
+++ b/tests/unit/test_comp_hgics.py
@@ -63,6 +63,20 @@ def _date_subclass_returning(today_value: _dt.date) -> type:
 class TestCompHgics:
     """Tests for `comp_hgics`."""
 
+    def test_empty_input_produces_typed_empty_output(self, test_paths: DataPaths) -> None:
+        """A 0-row input must not crash and must write a typed empty parquet."""
+        raw_data_dfs = test_paths.interim_dir / "raw_data_dfs"
+        raw_data_dfs.mkdir(parents=True, exist_ok=True)
+        pl.DataFrame(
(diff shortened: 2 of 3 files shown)


src/jkp/data/aux_functions.py
```diff
@@ -2377,6 +2377,11 @@ def comp_hgics(paths: DataPaths, lib):
         },
     }
     data = pl.read_parquet(file_paths["raw data"][lib])  # .sort(['gvkey', 'indfrom'])
+    if data.height == 0:
+        pl.DataFrame(schema={"gvkey": pl.Utf8, "date": pl.Date, "gics": pl.Int64}).write_parquet(
```
FernandoRDLL: **Necessary — fix the crash at its source instead of hardcoding the output schema.**

This hardcodes `{gvkey: Utf8, date: Date, gics: Int64}`, but the normal path never casts those dtypes: `gics` is inherited from raw `gsubind` (the `-999` literal supertypes to that column's dtype, it doesn't force `Int64`) and `gvkey` comes straight from raw. If production `gsubind` isn't `Int64`, empty and non-empty runs write *different* schemas, and any future column change to the normal path silently diverges from this literal. The golden fixture is `Int64`, so tests won't catch the drift.

The actual crash is just the eager compare on line ~2392, `data[["indfrom"]].max()[0, 0] > END_DATE`, blowing up when `.max()` is `None`. On a 0-row frame the rest of the pipeline (`over`/`date_ranges`/`explode`/`unique`/`sort`) all no-op cleanly — so making only that comparison null-safe lets the normal code path emit a correctly-typed empty output on its own, with no second hardcoded schema to keep in sync.

That same fix also closes a crash this guard misses: a **non-empty** input whose `indfrom` is entirely null also makes `.max()[0,0]` return `None`, so `None > END_DATE` raises the same `TypeError` as #192. Nothing upstream guarantees `indfrom` non-null (`comp.co_hgic`/`g_co_hgic` download unfiltered; `comp_hgics_aux` only filters on `gvkey`). The `height == 0` guard doesn't cover it; a null-safe compare does.

---

**Separate follow-up (not this PR):** the `data[["indfrom"]].max()[0, 0] > END_DATE` idiom — eager `.max()[0,0]` then a raw Python compare that dies on `None` — is a fragile pattern worth a repo-wide sweep. Please open a new issue/PR for that rather than expanding scope here.

src/jkp/data/aux_functions.py
```diff
@@ -2377,6 +2377,11 @@ def comp_hgics(paths: DataPaths, lib):
...
+            file_paths["output"][lib]
+        )
+        return
```
FernandoRDLL: **Necessary — both-sides-empty now silently zeroes GICS instead of crashing.**

Worth a conscious decision. Pre-PR, both inputs empty crashed loudly. Now two empty parquets flow through `hgics_join` to a 0-row `comp_hgics.parquet`, and `comp_industry`'s FULL OUTER JOIN keeps the `comp_other` rows with `gics = NULL` everywhere — `comp_ind.parquet` ships with null GICS, no log, no error. A broken WRDS download of both `co_hgic` tables would turn a loud crash into silent total loss of GICS classification.

A `log`/warning when a side comes back empty would make that visible rather than trading one failure mode for a quieter one.

tests/unit/test_hgics_join.py
```diff
@@ -193,6 +170,32 @@ def test_date_range_expansion(self) -> None:
             f"Expected contiguous dates {expected_dates}, got {result['date'].to_list()}"
         )
 
+    def test_empty_na_side(self) -> None:
```
FernandoRDLL: **Fold in — these duplicate the tests just above.** `test_empty_na_side` / `test_empty_gl_side` mostly repeat the now-modified `test_global_only_gvkey_preserved` / `test_national_only_gvkey_preserved` (same gvkeys/gics/dates, other side `_empty_frame()`). The only new bit is a 3-day vs 2-day span, and span expansion is already covered by `test_date_range_expansion`. Fold the extra `gvkey.unique()` assert into the originals or drop this pair.

tests/unit/test_hgics_join.py
```diff
@@ -193,6 +170,32 @@ def test_date_range_expansion(self) -> None:
...
+        """NA side is empty; output should contain only GL rows."""
+        na = _empty_frame()
+        gl = _hgics_frame(["200000"], [date(2020, 2, 1)], [date(2020, 2, 3)], [20202020])
+        _write_inputs(self.paths, na, gl)
+
+        hgics_join(self.paths)
+
+        result = pl.read_parquet(self.output_path).sort(["gvkey", "date"])
+        assert result.height == 3, f"Expected 3 rows (3-day GL span), got {result.height}"
+        assert result["gvkey"].unique().to_list() == ["200000"]
```
FernandoRDLL: **Fold in — missing assert messages.** These (and 196-197) lack diagnostic messages; tests/README.md:182 asks for one on every assert, and the neighboring `height` assert has one. Add `f"...got {...}"` messages.

tests/unit/test_comp_hgics.py
```diff
@@ -63,6 +63,20 @@ def today(cls) -> _dt.date:
...
+        raw_data_dfs.mkdir(parents=True, exist_ok=True)
+        pl.DataFrame(
+            schema={"gvkey": pl.Utf8, "indfrom": pl.Date, "indthru": pl.Date, "gics": pl.Int64}
+        ).write_parquet(raw_data_dfs / "comp_hgics_na.parquet")
+
+        comp_hgics(test_paths, "national")
+
+        output = pl.read_parquet(test_paths.interim_dir / "na_hgics.parquet")
+        assert output.height == 0, f"Expected 0 rows, got {output.height}"
+        assert output.schema == {"gvkey": pl.Utf8, "date": pl.Date, "gics": pl.Int64}
```
FernandoRDLL: **Fold in — missing assert message.** Same as the `hgics_join` asserts: no diagnostic message here, though the `height` assert on the line above has one (tests/README.md:182). Add one showing `output.schema`.

FernandoRDLL (changes requested):
Nice, tight fix — the guard works and dropping the `_FILLER_GVKEY` scaffolding is a real cleanup. Left inline comments. Two I'd want addressed at the root, plus two test cleanups I'd like folded in before merge rather than left as optional. There's also a separate follow-up (the `.max()[0, 0]` pattern) that shouldn't be done here — flagged inline with a suggestion to open its own issue/PR.

**Necessary:** 1 (schema/root fix, which also closes the all-null-`indfrom` crash), 2 (both-empty silent degradation).
**Fold in too:** 3 (duplicate tests), 4 (missing assert messages).
**Separate PR/issue:** the `.max()[0, 0]` pattern — see comment 1.

src/jkp/data/aux_functions.py
FernandoRDLL: We want to change `data[["indfrom"]].max()[0, 0] > END_DATE` to another pattern in the actual codebase (aux_functions.py)

src/jkp/data/aux_functions.py
FernandoRDLL: Is this ready? I saw you resolved the previous ones @miahdy

FernandoRDLL (approved):

FernandoRDLL:
@miahdy Great job! Just merged

FernandoRDLL (approved):
Re-reviewed: all four points addressed — null-safe compare (drops hardcoded schema, closes null-indfrom variant), empty-side warning, dup tests removed, assert messages added. CI green. LGTM.

Issue #194, opened 2026-06-05
[Bug]: WRDS raw table download in HPC slowed from ~30 min to 60–120 min


FernandoRDLL:
### Which script were you running?

main.py (data generation)

### Error Category

Data download failure

### Bug Description

The raw data download step — `download_raw_data_tables()` in `src/jkp/data/aux_functions.py` — has been slowing down significantly on the HPC cluster. It used to complete in roughly **30 minutes**; over the last 3–4 pipeline runs it has been taking **60–120 minutes** instead.

No errors are raised — the step completes successfully, just much more slowly. This affects both wall-clock time and scheduler allocations for full pipeline runs.

### Steps to Reproduce

1. Run `jkp build data/` on the HPC cluster
2. Observe the timing reported by `@measure_time` for `download_raw_data_tables`
3. Compare against earlier runs (~30 min): the last 3–4 runs have taken 60–120 min

### Additional context

Possible suspects worth investigating:
- WRDS server-side throttling or load
- Network path between the cluster and WRDS
- DuckDB `postgres_scan()` per-query connection overhead vs. the persistent `ATTACH` path (`persistent_connection=True`)
- Table growth on the WRDS side

@capellini can you please help with this one?

capellini:
I've been digging into this on the Yale SOM HPC with an instrumented benchmark and resource monitor. Here's a summary of what I found:

## The bottleneck is single-threaded transfer

During every table download, one CPU core sits pinned at 100% while ~25 cores are idle, and:

- the network link bursts to 932 Mbit/s but the download sustains a small fraction of that, so we're not link-bound
- node-local `/tmp` is no faster than GPFS scratch, disk_busy is always around 0%, so we're not disk I/O bound
- `COUNT(*)` over 7.3 M WRDS table rows returns in ~3–4s, so it also does not appear to be a WRDS server-side problem, nor a DuckDB `postgres_scan` query time issue
- 1.5 TB RAM free, so not a memory issue

Each table is pulled over a single-threaded, single connection and the limiting factor is the
client-side network ingesting of that one stream (~7 MB/s when the link is healthy).

## What actually changed ~2 weeks ago

When I conducted my tests, throughput from WRDS to the cluster seemed to be degraded and highly variable.
On the same node, same table, same night I measured it swing ~5–10x (e.g. `dsf_v2` 2023+ took ~100 s early in the
night and 568 s a few hours later). That lines up with the WRDS + HPC maintenance windows and is
[...]

PR #209, opened 2026-06-20
perf: parallelize WRDS raw-table downloads (#194)

capellini:
## Summary

Addresses #194 (raw-table download slowed from ~30 min to 60–120 min). Diagnosis on the Yale SOM
HPC showed the download is single-threaded and transfer-bound, and that WRDS↔cluster throughput
became degraded/variable after the maintenance windows (environmental, not our code). A single
sequential stream is maximally exposed to that, so this PR **parallelizes the download** to recover
the time.

`download_raw_data_tables()` gains a `max_workers` parameter (CLI: `jkp build --download-workers/-j`,
default `1` = current sequential behavior, fully backward-compatible):

- **Across-table parallelism** — the ~26 tables download concurrently over a pool of persistent
  `ATTACH` connections (one per worker), drained from a shared queue for load balancing. Clamped to
  the WRDS per-account connection limit (`WRDS_MAX_CONNECTIONS`, pool ≤ 6).
- **Date-range splitting** — the three giant daily tables (`crsp.dsf_v2`, `comp.secd`,
  `comp.g_secd`; 110–300 M rows) are split into `max_workers` row-balanced date-range chunks
  (boundaries from concurrent per-year row histograms), downloaded as a mix of whole-table and chunk
  tasks, then concatenated back into one parquet per table.
- `--persistent-connection` takes precedence and forces a single sequential connection (preserving
  its single-MFA-prompt behavior on NAT-rotated networks); a notice goes to stderr if both are set.

## Change Type

- [ ] Documentation only
- [x] Performance-impacting change

## Methodological Impact

None. No factor definitions or calculations are touched — only how the raw WRDS tables are fetched.

## Data Impact

**None of practical significance**, confirmed by a full 3-way determinism experiment on Yale — three
complete `jkp build` runs (`main`, this branch at `--download-workers 1`, and at
`--download-workers 4`), with an order-independent diff of every `interim/` and `processed/` output.

- **Raw is reproduced exactly:** all 26 raw tables are identical (order-independent, by row count and
  per-column fingerprint), including the 110–300 M-row giants — the split/parallel download yields the
  same raw row *set* as sequential.
- **Final outputs:** the parallel run matches the sequential runs on every output file and column
  except one — `other_output/market_returns.parquet`'s `dolvol_lag1` (a dollar-volume sum over
  securities), which differs by at most **5.9e-16 relative (machine epsilon / last-bit float)**,
  because a different security order changes the order of a floating-point summation.

The experiment also surfaced **potential pre-existing determinism problems in the pipeline that are
unrelated to this change** — two *sequential* runs with byte-identical raw input already differ on a
number of downstream files (float-reduction noise, plus an order-sensitivity in the industry-history
expansion). The parallel download stays within that existing run-to-run variation. These are **not
introduced by this PR and should be investigated/fixed separately, in their own PR** — flagged here
only so the reviewer knows the observed downstream differences don't come from this change.

## Breaking Changes

None. `max_workers` defaults to 1, so existing invocations are unchanged.

## Tests

679 unit tests pass. New tests in `tests/unit/test_wrds_download.py` cover:
- worker-count clamping to the WRDS connection limit; parallel path covers the same tables; one
  connection per worker (ATTACH/DETACH/close); error aggregation; password redaction;
  `--persistent-connection` forcing sequential.
- date-range splitting: row-balanced chunk boundaries (coverage, balance, no overlap, degenerate
  cases), task building (split vs whole-table), chunk concatenation, the `start_date` date-range
  clause, and an end-to-end (mocked) check that giants expand into `max_workers` chunks.

Connection safety and end-to-end correctness/speedup were verified on live WRDS (see notes).

## Checklist

(`jkp build` was run end-to-end three times on Yale for the determinism experiment — see Data
Impact. `jkp portfolio` was not separately re-run; the one observed difference is machine-epsilon
and within the pipeline's existing run-to-run variation, so re-running it would not change the
conclusion.)

## Additional Notes

Verified on Yale (full 26-table history, `max_workers=6`):

| | sequential | parallel | speedup |
|---|---|---|---|
| across-table only | 85 min | 53 min | 1.6× |
| + date-range splitting | 64 min | 25 min | **2.6×** |

- **Connection safety** (measured client-side with `psutil`): `max_workers=N` opens exactly N client
  connections — the boundary phase uses ≤3, the workers 6, non-overlapping — so the peak stays at 6,
  one under the WRDS limit of 7. (Server-side `pg_stat_activity` over-reports because WRDS fronts
  PostgreSQL with pgpool; the client socket count is the real number.)
- No `CHANGELOG_DATA.md` entry: this changes *how* tables are downloaded, not *what* is written.

### DIFF
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -1003,12 +1003,13 @@ def _attach_download_worker(
     task_errors: list[str],
     attach_errors: list[str],
     errors_lock: threading.Lock,
+    stop_event: threading.Event,
 ) -> None:
     """One parallel worker: open a private ATTACH connection and drain the task queue.
 
     Each task is a whole table or one date-range chunk. The worker holds exactly one WRDS
     connection (``threads=1`` keeps DuckDB from opening extra connections for a parallel scan)
-    and reuses it.
+    and reuses it. It stops pulling new tasks once ``stop_event`` is set (used to unwind on Ctrl-C).
 
     Failures are recorded (password-redacted) instead of raised, so one bad task doesn't abandon
     the rest. Attach failures and download (task) failures are kept in separate lists: because the
@@ -1028,14 +1029,14 @@ def _attach_download_worker(
             with errors_lock:
                 attach_errors.append(str(e))
                 print(
-                    f"Warning: a WRDS download worker failed to attach ({e}); "
-                    "continuing with the remaining worker(s).",
+                    f"Warning: a WRDS download worker failed to attach ({e}); continuing with the "
+                    "remaining worker(s). Press Ctrl-C to cancel and re-run for full parallelism.",
                     file=sys.stderr,
                     flush=True,
                 )
             return
         try:
-            while True:
+            while not stop_event.is_set():
                 try:
                     task = task_queue.get_nowait()
                 except queue.Empty:
@@ -1105,18 +1106,48 @@ def _download_tables_parallel(
     task_errors: list[str] = []
     attach_errors: list[str] = []
     errors_lock = threading.Lock()
+    stop_event = threading.Event()
     threads = [
         threading.Thread(
             target=_attach_download_worker,
-            args=(task_queue, conninfo, password, task_errors, attach_errors, errors_lock),
+            args=(
+                task_queue,
+                conninfo,
+                password,
+                task_errors,
+                attach_errors,
+                errors_lock,
+                stop_event,
+            ),
             name=f"wrds-download-{i}",
         )
         for i in range(workers)
     ]
     for t in threads:
         t.start()
-    for t in threads:
-        t.join()
+    # Join with a timeout so the main thread stays responsive to Ctrl-C (a plain join() would defer
+    # the KeyboardInterrupt until the whole download finished). On Ctrl-C, ask the workers to stop
+    # pulling new tasks; they finish the download already in flight and exit.
+    #
+    # We deliberately do NOT call DuckDB's con.interrupt() to abort an in-flight COPY mid-transfer:
+    # it would require sharing every worker's connection with this thread plus telling a user-cancel
+    # apart from a real failure, and its effect on a COPY blocked in the postgres extension's network
+    # read is unverified. Consequently a Ctrl-C takes effect only after the currently-downloading
+    # task(s) finish -- bounded by one chunk, not the entire remaining download.
+    try:
+        for t in threads:
+            while t.is_alive():
+                t.join(timeout=0.5)
+    except KeyboardInterrupt:
+        print(
+            "\nInterrupted: stopping after the in-flight download(s) finish...",
+            file=sys.stderr,
+            flush=True,
+        )
+        stop_event.set()
+        for t in threads:
+            t.join()
+        raise
 
     # A download failure means a table is missing -> always fatal.
     if task_errors:
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -989,7 +989,9 @@ def _remove_chunk_parts(filename: str) -> None:
     ones. The final (non-part) file itself is left untouched.
     """
     p = Path(filename)
-    for f in p.parent.glob(f"{p.stem}.part*{p.suffix}"):
+    # Match exactly what _chunk_path writes (.part00..part99) rather than `part*`, which would also
+    # sweep an unrelated file like `<stem>.partial.parquet`.
+    for f in p.parent.glob(f"{p.stem}.part[0-9][0-9]{p.suffix}"):
         with contextlib.suppress(FileNotFoundError):
             f.unlink()
 
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -1,10 +1,14 @@
 from __future__ import annotations
 
+import contextlib
 import functools
 import operator
 import os
+import queue
 import re
 import shutil
+import sys
+import threading
 import time
 import warnings
 from datetime import date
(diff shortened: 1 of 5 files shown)


FernandoRDLL:
@capellini Do you want me to review this?

capellini:
Yes, please @FernandoRDLL -- thank you.

src/jkp/data/aux_functions.py
```diff
@@ -765,13 +787,357 @@ def download_wrds_table(
     """)
 
 
+# WRDS enforces a per-role PostgreSQL CONNECTION LIMIT (currently 7 for our account; see
```
FernandoRDLL, line 793: We've loosely followed an "engine-level parallelism only" rule in pipeline code, and I'm happy to treat this as the legitimate exception — N concurrent WRDS connections can't come from the query engine. Could you add a line here saying so explicitly, so this doesn't become precedent for thread pools in compute code?

src/jkp/data/aux_functions.py
```diff
@@ -765,13 +787,357 @@ def download_wrds_table(
...
+
+def _remove_chunk_parts(filename: str) -> None:
+    """Delete any existing chunk-part files for ``filename`` (see _chunk_path).
+
+    A prior interrupted run may have written chunk parts that were never concatenated — possibly
+    with a different worker count, hence different indices — so clear them all before writing fresh
+    ones. The final (non-part) file itself is left untouched.
+    """
+    p = Path(filename)
+    for f in p.parent.glob(f"{p.stem}.part*{p.suffix}"):
```
FernandoRDLL: `part*` would also match a file like `crsp_dsf_v2.partial.parquet`. Theoretical, but `part[0-9][0-9]` matches exactly what `_chunk_path` writes.

src/jkp/data/aux_functions.py
```diff
@@ -765,13 +787,357 @@ def download_wrds_table(
...
+    """
+    # `with duckdb.connect(...)` closes the connection on every exit path (return / exception).
+    with duckdb.connect(":memory:") as con:
+        con.execute("SET threads TO 1")
+        con.execute("INSTALL postgres; LOAD postgres;")
+        try:
+            _attach_wrds(con, conninfo, password)
+        except Exception as e:  # noqa: BLE001
+            with errors_lock:
+                errors.append(f"<attach>: {e}")
```
FernandoRDLL: If one worker fails to ATTACH but the surviving workers drain the queue, this still fails the whole run at the end — after all 26 tables downloaded successfully. Maybe track attach failures separately and only raise when tasks were actually left undone or failed; if everything completed, a notice to stderr seems enough. As is, a transient MFA hiccup on one worker throws away a 25-minute download.

src/jkp/data/aux_functions.py
```diff
@@ -765,13 +787,357 @@ def download_wrds_table(
...
+                        task.table,
+                        task.out,
+                        date_column=task.date_column,
+                        end_date=task.end_date,
+                        start_date=task.start_date,
+                    )
+                except Exception as e:  # noqa: BLE001
+                    msg = str(e)
+                    if password and password in msg:
+                        msg = "<redacted>"
```
FernandoRDLL: Replacing the whole message loses the diagnostic context. `msg.replace(password, "***")` redacts the credential and keeps the rest of the error.

FernandoRDLL (changes requested):
A few small suggestions inline.

One ask before merge: `main` has moved on since this branch was cut , so please rebase onto current `main`.

src/jkp/data/aux_functions.py
capellini: Fixed: the sweep now globs `part[0-9][0-9]`, matching exactly what `_chunk_path` writes, so an unrelated `partial.parquet` is left untouched. Added a test asserting that. 77423b9

src/jkp/data/aux_functions.py
capellini: Done. Now `str(e).replace(password, "***")`, so the credential is scrubbed but the rest of the error survives for diagnostics. Test asserts the password is gone and the error text is preserved. a15064b

src/jkp/data/aux_functions.py
capellini: Reworked so download (task) failures and attach failures are tracked separately: we raise only if a table actually failed to download, or if the queue never drained (every worker failed to attach and nothing downloaded).

If the surviving workers completed every table, we warn to stderr instead of failing so a transient one-worker hiccup no longer discards a finished download. I also added an immediate stderr warning the moment a worker fails to attach, so the run doesn't silently proceed under-provisioned and so the user has a chance to bail on the current download and retry.

Tests cover both the tolerated-one-failure and all-failures-raise cases. e56b1a0

src/jkp/data/aux_functions.py
capellini: Added a comment here spelling out that the thread pool is a deliberate, narrow exception (the parallelism is N concurrent WRDS connections, which the query engine can't provide), and explicitly not a precedent for thread pools in compute code. ebd017e

I'd push back slightly on treating "engine-level parallelism only" as absolute. The engine is the right tool for within-operation data parallelism, but it can't touch coarse-grained, independent units at the orchestration level, and we have some: the `for ex in countries:` loop in portfolio.py builds ~90 independent countries serially and comp_hgics runs global-then-national sequentially.

However some of that underutilization may be due to memory-constraints where more threads won't help and could OOM. So we should profile first. Worth a proper look as its own investigation outside of this PR.

capellini:
Thanks for the review.

All four inline points addressed in separate commits (ebd017e, 77423b9, a15064b, e56b1a0), plus a feature commit cd5a8f7 making Ctrl-C actually interrupt the parallel download.

Rebased onto current main. 

I'm going to put this back in draft and do some smoke tests on the HPC for added verification over and above the unit tests. It'll be ready for another look when it comes out of draft again.

capellini:
Out of draft again and ready for another look.

FernandoRDLL (approved):
All four points addressed and verified in code + tests. Thanks.

PR #229, opened 2026-07-08
test: unit tests for firm_age

FernandoRDLL:
## Simple explanation

Focused unit tests for `firm_age`. Stacked on the golden-firm_age PR; adds only `tests/unit/test_firm_age.py`. Merge the golden PR first; this retargets to main.

## Technical details

Asserts real semantics on small synthetic frames: single-source firms, all-three-source firms, mismatched earliest dates, monthend filter, per-id partition, null-permco, null-first_obs fallback, empty input. Notably proves the `monthend==1` filter matters via an A/B flip and that first-date is a per-id partition. Reuses the golden PR's `firm_age_inputs.py` builders. No `src/` changes, no new fixture data.

## Test plan

`PYTHONPATH=src uv run --no-sync python -m pytest tests/unit/test_firm_age.py --no-cov` passes; ruff clean.

Closes #219

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #230, opened 2026-07-09
fix: make residual_momentum byte-deterministic (fixes CI on main)

FernandoRDLL:
## Simple explanation

CI on `main` went red right after #229 merged. The failing test is `tests/unit/test_residual_momentum.py::test_determinism_rerun`, which reruns `residual_momentum` on identical inputs and asserts the two outputs are byte-identical. They weren't. The failure is unrelated to #229's firm_age tests — #229 just happened to be the push that ran the suite.

## Technical details

`res_mom` gets its rows from `prep_data_factor_regs`, which builds `__msf2` in DuckDB with `threads=os.cpu_count()` and hands it back via `to_polars()`. That scan's **row order is nondeterministic** — the row *set* is identical every run (sorted frames compare equal), but the order isn't.

Downstream, `res_mom` runs a `polars-ols` least-squares residual solve `.over(["id_int","group_number"])` and then a `mean/std` reduction. Both are float-summation-order sensitive, so a different input order yields residuals that differ at ~1e-15. That's enough to break the exact `.equals()` in `test_determinism_rerun`.

The golden regression test didn't catch it because it compares the value column with `STANDARD` tolerance (rtol 1e-6), which absorbs the noise.

**Fix:** sort by `[id_int, group_number, aux_date]` before the OLS in `res_mom`, so the solver and the reduction always see a fixed row order. Output values are unchanged within tolerance; the committed golden fixture still matches.

## Test plan

- `test_determinism_rerun` failed 12/12 runs before, passes 10/10 after.
- Full `tests/unit/test_residual_momentum.py` + `tests/golden/test_residual_momentum_golden.py`: 26 passed.
- `ruff check` / `ruff format --check` clean.

### DIFF
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -8925,7 +8925,12 @@ def res_mom(df, sfx, __min, incl, skip):
         .over(["id_int", "group_number"])
     )
     df = (
-        df.filter(col("hml").is_not_null() & col("smb_ff").is_not_null())
+        # Fix within-group row order before the OLS: base_data comes from a
+        # multithreaded DuckDB scan whose row order is nondeterministic, and the
+        # least-squares solve is float-order-sensitive, so an unsorted input
+        # yields byte-different residuals across runs.
+        df.sort(["id_int", "group_number", "aux_date"])
+        .filter(col("hml").is_not_null() & col("smb_ff").is_not_null())
         .with_columns(
             res=res_exp.alias("res"),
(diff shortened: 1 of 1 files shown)


FernandoRDLL:
@theisij — CI-fix for the red run on main after #229. Root cause is nondeterministic DuckDB row order feeding the OLS in `res_mom`, not the firm_age tests. Details in the PR body.

theisij (approved):

PR #231, opened 2026-07-09
fix: make market_beta byte-deterministic across runs

FernandoRDLL:
## Simple explanation

Companion to #230. That PR fixed `residual_momentum`; this fixes the identical latent bug in `market_beta`. `market_beta` has no determinism test, so CI never went red on it, but the same non-reproducibility is present.

## Technical details

`capm` (called by `market_beta`) gets its rows from `prep_data_factor_regs`, which builds `__msf2` in DuckDB with `threads=os.cpu_count()` and returns it via `to_polars()`. That scan's **row order is nondeterministic** — the row set is identical every run, but the order isn't.

`capm` then computes `cov(ret_exc, mktrf) / var(mktrf)` and the residual `std` in a `group_by().agg()`. Those reductions are float-summation-order sensitive, so a different input order yields betas / ivol that differ at ~1e-15 run to run.

**Fix:** sort by `[id_int, group_number, aux_date]` before the `group_by`, mirroring the `res_mom` fix in #230. Output values are unchanged within tolerance; the committed golden fixture still matches.

## Test plan

- Added a local reproducer: `market_beta` output was byte-different across two runs before, byte-identical after.
- `tests/unit/test_market_beta.py` + `tests/golden/test_market_beta_golden.py`: 14 passed.
- `ruff check` / `ruff format --check` clean.

### DIFF
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -9050,13 +9050,21 @@ def capm(df, sfx, __min):
     Output:
         LazyFrame with f'beta{sfx}' and f'ivol_capm{sfx}'.
     """
-    df = df.group_by(["id_int", "group_number"]).agg(
-        [
-            (pl.cov("ret_exc", "mktrf") / pl.var("mktrf")).alias(f"beta{sfx}"),
-            (col("ret_exc") - col("mktrf") * (pl.cov("ret_exc", "mktrf") / pl.var("mktrf")))
-            .std()
-            .alias(f"ivol_capm{sfx}"),
-        ]
+    # Fix within-group row order before the reductions: base_data comes from a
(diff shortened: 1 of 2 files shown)


FernandoRDLL:
@theisij — follow-up to #230. Same DuckDB-order → float-sensitive-reduction bug, this time in `market_beta`/`capm`. No determinism test guarding it, but the fix keeps it byte-reproducible. Golden unchanged.

theisij (dismissed):

theisij:
It looked good to me, but I see now that the checks failed. Please take a look @FernandoRDLL

FernandoRDLL:
Ready @theisij

theisij (approved):

PR #234, opened 2026-07-11
Apply Bessembinder et al. (2023) data corrections to Compustat

FernandoRDLL:
## Simple explanation

Applies the Bessembinder et al. (2023) data correction to the Compustat daily security files: it repairs decimal-shift price errors and drops unreliable observations before returns are computed. Adds two related returns fixes (never-dividend `trfd` recovery and a >1000% return screen) and a documentation report quantifying the impact.

The effect is concentrated outside the US — US factor series draw primarily on CRSP, so their long-short returns are essentially unchanged (median correlation 0.999–1.000). Non-US factors shift more (value-weighted median 0.998, equal-weighted 0.955), driven by cleanup of small-cap Compustat errors.

## Technical details

- Section 6 (decimal-error detection + rescaling) and Section 8 filters 8a–8h (drops) run as numba kernels over per-security date-sorted arrays, with a memory-bounded slim path (DuckDB external sort → spill → kernels → lazy reattach). Applied to `comp_secd` (NA) and `comp_g_secd` (Global) inside `gen_comp_dsf`.
- Multi-period detection adapted for daily data: symmetric windows over a priority set (1, 2, 3, 5, 10, 21 trading days) instead of Bessembinder's monthly asymmetric all-pairs. Single-period detection, rescale factors, and drop criteria are unchanged from the paper.
- `trfd = 1` recovery for never-dividend securities (per WRDS Compustat guidance) so their price returns survive instead of being dropped.
- Post-correction screen: Compustat `ret > 10` (>1000%) set to null in `gen_returns_df`.
- Report under `documentation/bessembinder_correction/` (scope-of-change tables + factor-impact figures).
- `CHANGELOG_DATA.md` entry.

## Test plan

- Unit tests: `tests/unit/test_bessembinder.py` (slim-path Section 6/8), `tests/unit/test_deterministic_dedup.py` (ret>10 nulling).
- Full-scale validation on Yale HPC: matched baseline vs corrected build on an identical raw download. US factor correlations 0.999–1.000; non-US VW 0.998 / EW 0.955. Scope tables (Total Current / Corrections / Removals / Post-Correction by size × region) tallied from the corrected run's Compustat DSF.


FernandoRDLL:
@capellini can you review the separation of the code in modules please?
I changed the code since you handed it off. It takes like 5 min to make the correction. Let me know what you think of the code
@theisij can you please read the release note? documentation/bessembinder_correction/bessembinder_correction.html

FernandoRDLL:
@claude please review and make suggestions if necessary

FernandoRDLL:
Addressed the review findings:

1. **Inverted inequality in the report** — fixed in d12099d. The qmd now reads `ret > 10`, matching the code and `CHANGELOG_DATA.md`, and the html was re-rendered.
2. **`@measure_time` removal from `aug_msf_v2` / `build_mcti`** — intentional (4c28730): both run in well under a second, so their timing blocks were noise in the pipeline log. Leaving as is.

capellini:
@FernandoRDLL before I start my review, I wanted to quickly check on the file deletions in this PR. Commit bc744890 ("Drop Bessembinder report artifacts and validation scripts") removes the Bessembinder scratch outputs and build helpers, which makes sense. But it also deletes three files that don't look like regenerable scratch artifacts:

- documentation/documentation.tex (−3,185 lines): the main documentation source
- documentation/documentation.pdf: its compiled output
- documentation/is_there_a_replication_crisis_in_finance_JF.pdf: the JKP paper PDF

Was removing these intentional, or did they get bundled into the cleanup commit by accident?

FernandoRDLL:
Good catch, thanks — those three were bundled in by accident. bc744890 was meant to remove only the Bessembinder scratch outputs; `documentation.tex`, its compiled PDF, and the JKP paper PDF are standing repo documentation and shouldn't have been touched. Restored all three from `main` in 5cb8e0c.

src/jkp/data/bessembinder_kernels.py
theisij, line 1: Do we need two files for everything related to the Bessembinder adjustment? Feels like they are conceptually similar enough that we'd want just one. Alternatively, we could have them in a folder under src/jkp/data. 

Also, let's not call the modules 'Bessembinder.' First, there are multiple co-authors on this paper, and second the generic thing we're trying to do is to adjust compustat returns. So maybe something like compustat_return_correction.py

CHANGELOG_DATA.md
theisij, line 1: This should have a link to the release notes. Something like, to see the impact, look at the release notes (link)

theisij (commented):
Do we need two files for everything related to the Bessembinder adjustment? Feels like they are conceptually similar enough that we'd want just one. Alternatively, we could have them in a folder under src/jkp/data.

Also, let's not call the modules 'Bessembinder.' First, there are multiple co-authors on this paper, and second the generic thing we're trying to do is to adjust compustat returns. So maybe something like compustat_return_correction.py. Wantever name we pick, we should be consistent across the codebase. Remove the Bessembinder name for all file names, and then we can mention the paper in code comments and of course in the release notes

FernandoRDLL:
Thanks Theis — addressed both points.

**One file.** `bessembinder_kernels.py` is folded into the main module, so the whole correction now lives in a single file: the Polars/DuckDB orchestration and the numba per-security kernels together, kernels at the bottom under a comment banner.

**Dropped the "Bessembinder" and Section 6/8 naming** and made it consistent across code, tests, and docs. The generic thing is correcting Compustat returns, so:

- `src/jkp/data/bessembinder.py` → `compustat_correction.py`
- `apply_bessembinder_section6` → `correct_decimal_errors`, `apply_bessembinder_section8` → `drop_unreliable_observations`
- kernels/config/params renamed to say what they do rather than which paper section they come from (`section8_all` → `run_all_filters`, `_filter_8a_low_volume` → `_filter_low_volume`, `BESS_*`/`SECTION*` → `CORRECTION_*`/`DECIMAL_*`/`FILTER_*`, `correction_method="bessembinder"` → `"multiplier"`, `gen_comp_dsf(apply_bessembinder=...)` → `apply_correction`, spill files `__bess_*` → `__corr_*`)
- report: `documentation/bessembinder_correction/` → `compustat_correction/`, `bess_compare.py` → `correction_compare.py`, and the Bessembinder/Section labels scrubbed from the report + `CHANGELOG_DATA.md`

The paper is now cited only where it's provenance: the two main-caller docstrings (which note they implement Section 6 / Section 8 of Bessembinder et al. 2023), a few code comments, and the release notes.

FernandoRDLL:
I think it's now ready @theisij

theisij (approved):
Beautiful. Great work!!

PR #245, opened 2026-07-13
feat: support ~/.pgpass for WRDS credentials, replacing the plaintext keyring

### LINKED ISSUE #232: [Feature]: Support ~/.pgpass for WRDS auth
## Feature Description

On headless environments (HPC compute nodes), the pipeline authenticates to WRDS by reading the password from a keyring and injecting it into the libpq connection string as an explicit `password=` field (`src/jkp/data/aux_functions.py:684`, via `get_wrds_credentials()`). Because compute nodes have no system keyring daemon, the only headless option today is `JKP_ALLOW_PLAINTEXT_KEYRING=1`, which writes the WRDS password **unencrypted** to a mode-600 file under `~/.local/share/python_keyring/` and emits a warning on every run.

The root cause is not the warning — it's that the pipeline *requires* jkp to hold the password at all. libpq already falls back to `~/.pgpass` when the connection string omits `password=`, and WRDS's own tooling assumes a `~/.pgpass` entry. Many nodes (ours included) already have a valid mode-600 `~/.pgpass`. The pipeline ignores it and forces a second copy of the secret into a jkp-managed store.

Proposal: make the password optional. When no password is supplied (no `WRDS_PASSWORD`, no keyring entry) but a `~/.pgpass` / `PGPASSFILE` exists, omit `password=` from the connection string and let libpq authenticate from `.pgpass`. jkp then never handles the password on headless nodes: no keyring, no plaintext-keyring file, no warning.

## Use Case / Motivation

Every HPC pipeline run currently either (a) stores the WRDS password in plaintext in a jkp-specific file and prints a security warning, or (b) requires exporting `WRDS_USERNAME`/`WRDS_PASSWORD` into the environment. Both are secret-at-rest copies that duplicate the `~/.pgpass` the node already has. Supporting `.pgpass` removes password handling from jkp on headless nodes entirely and matches the standard postgres/WRDS auth pattern.

## Proposed Implementation

- `src/jkp/data/wrds_credentials.py`: allow a username-only credential (password optional) when `~/.pgpass` or `PGPASSFILE` is present; only fall back to the keyring / plaintext-keyring path when no `.pgpass` is available.
- `src/jkp/data/aux_functions.py` (~line 684): build the libpq conninfo without `password=` when no password is available, so libpq reads `~/.pgpass`.
- `slurm/submit_job_som_hpc.slurm`: document that a `~/.pgpass` entry is the preferred headless path; `JKP_ALLOW_PLAINTEXT_KEYRING=1` becomes the last-resort fallback rather than the default.
- Keep env-var and keyring precedence unchanged for interactive/desktop use.

Scope: credential handling only. No change to downloaded data or pipeline outputs.

## Warning observed (headless run)

```
src/jkp/data/wrds_credentials.py:128: UserWarning: JKP_ALLOW_PLAINTEXT_KEYRING=1 is set: switching the keyring backend to keyrings.alt.file.PlaintextKeyring for this process. The WRDS password will be read from / written to a mode-600 file under ~/.local/share/python_keyring/.
  _ensure_keyring_backend()
```

This fires on every headless run that sets `JKP_ALLOW_PLAINTEXT_KEYRING=1` (the current default in `slurm/submit_job_som_hpc.slurm`). Under the `.pgpass` path above, the plaintext-keyring backend is never selected, so the warning disappears.

capellini:
## Summary

Rework WRDS credential handling to use libpq's `~/.pgpass` / `$PGPASSFILE` as the headless credential store, and remove the file-backed plaintext keyring (the `JKP_ALLOW_PLAINTEXT_KEYRING` opt-in and the `keyrings-alt` dependency).

- **Decoupled resolution.** Username: `WRDS_USERNAME` → saved state file → prompt. Password: `WRDS_PASSWORD` → system keyring → `~/.pgpass` / `$PGPASSFILE`.
- **libpq handles the secret on headless nodes.** When the password comes from a pgpass file, jkp omits `password=` from the connection string and lets libpq read the file; the resolved source is printed to stderr.
- **`jkp connect` stays the single interface.** It stores in the system keyring on desktops, or writes `~/.pgpass` (mode 600) on a headless login node — `$HOME` is shared, so batch jobs read it with no further setup.
- **Automatic, non-interactive migration.** An existing plaintext-keyring entry is moved to `~/.pgpass` (removing only jkp's own `[WRDS]` section, preserving any other tool's), and the username cache moves from `~/.wrds_user` to a per-OS state directory.

Closes #232

## Change Type

Dependency delta: drop `keyrings-alt` and `pycryptodomex`, add `platformdirs`.

## Methodological Impact

None.

## Data Impact

None — credential handling only. No change to downloaded data, characteristics, or factor returns.

## Breaking Changes

**Yes — `JKP_ALLOW_PLAINTEXT_KEYRING` and the file-backed plaintext keyring are removed.** Existing headless setups are migrated automatically on the next run: jkp reads its entry from the legacy `keyring_pass.cfg`, writes `~/.pgpass`, and removes only its own section. No user action is required; provisioning going forward is `~/.pgpass` (via `jkp connect` on a login node) or `WRDS_USERNAME`/`WRDS_PASSWORD`.

This is a pre-release (pre-PyPI) breaking change, so no released dataset is affected. Release notes should mention the removed env var and the automatic migration.

## Tests

- Rewrote `tests/unit/test_wrds_credentials.py` for the new model: env/keyring/`.pgpass` precedence and the full credential-source interaction matrix; both migrations; surgical section removal that preserves other tools' entries; `.pgpass` escaping, wildcard matching, and 0600-permission handling; the on-disk keyring format; and the interactive prompt paths.
- Added a `password=None` case for `gen_wrds_connection_info` in `tests/unit/test_wrds_download.py`.
- Full unit suite passes (838).

## Checklist

## Additional Notes

- Validated end-to-end on an HPC cluster: the migration ran against real credential files on a login node, and a full pipeline run on a compute node authenticated to live WRDS via the migrated `~/.pgpass` and downloaded all raw tables.
- Cross-platform: the `.pgpass` and legacy-keyring paths are derived from libpq's and keyring's own per-OS conventions (`keyring.util.platform_.data_root()` for the legacy file). macOS/Windows path behavior is reasoned from those libraries' source; CI executes on Linux only.
- `slurm/submit_job_som_hpc.slurm` and `README.md` are updated to document the login-node `~/.pgpass` workflow.

### DIFF
--- a/slurm/submit_job_som_hpc.slurm
+++ b/slurm/submit_job_som_hpc.slurm
@@ -41,9 +41,7 @@ source .venv/bin/activate
 # separate copy. Point PGPASSFILE at a non-default location if you prefer.
 #
 # Fully non-interactive alternative (no login-node terminal): export
-# WRDS_USERNAME / WRDS_PASSWORD, or create ~/.pgpass yourself (chmod 600):
-#
-#     wrds-pgdata.wharton.upenn.edu:9737:wrds:<user>:<password>
+# WRDS_USERNAME / WRDS_PASSWORD.
 
 # Create char data
 # Set PERSISTENT_WRDS_CONNECTION=1 to use a single WRDS connection (reduces MFA on NAT-rotated networks)
--- a/slurm/submit_job_som_hpc.slurm
+++ b/slurm/submit_job_som_hpc.slurm
@@ -30,17 +30,20 @@ export NUMEXPR_NUM_THREADS=${SLURM_CPUS_PER_TASK}
 # Activate the project's virtual environment so the jkp CLI is on PATH
 source .venv/bin/activate
 
-# WRDS credentials. Compute nodes don't have a system keyring daemon, so the
-# pipeline needs an alternate source. The default below opts in to the
-# file-backed keyring (a mode-600 file under ~/.local/share/python_keyring/);
-# populate it once on the login node by running:
+# WRDS credentials. Compute nodes have no system keyring daemon, so provision
+# credentials ONCE on a login node (which has an interactive terminal) BEFORE
+# submitting this job -- $HOME is shared with the compute nodes, so the batch
+# job sees the file:
 #
-#     JKP_ALLOW_PLAINTEXT_KEYRING=1 jkp connect
+#     jkp connect          # prompts, writes ~/.pgpass (mode 600)
 #
-# Alternative (for shared service accounts or container deployments): unset
-# JKP_ALLOW_PLAINTEXT_KEYRING and export WRDS_USERNAME / WRDS_PASSWORD instead,
-# e.g. by sourcing a mode-600 ~/.wrds_secrets file with `export WRDS_USERNAME=...`.
-export JKP_ALLOW_PLAINTEXT_KEYRING=1
+# The pipeline reads the password from ~/.pgpass via libpq; jkp never stores a
+# separate copy. Point PGPASSFILE at a non-default location if you prefer.
+#
+# Fully non-interactive alternative (no login-node terminal): export
+# WRDS_USERNAME / WRDS_PASSWORD, or create ~/.pgpass yourself (chmod 600):
+#
+#     wrds-pgdata.wharton.upenn.edu:9737:wrds:<user>:<password>
 
 # Create char data
 # Set PERSISTENT_WRDS_CONNECTION=1 to use a single WRDS connection (reduces MFA on NAT-rotated networks)
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -675,6 +675,28 @@ def gen_crsp_sf(paths: DataPaths, freq):
     return result
 
 
+def _pg_escape_value(value: str) -> str:
+    """libpq-escape a conninfo value (user or password) for a single-quoted field.
+
+    libpq accepts single-quoted values with backslash-escaped ``\\`` and ``'``,
+    so quoting lets a value hold spaces or special characters without breaking the
+    conninfo. For the password this is also the form that appears in any error text
+    echoing the connection string, so the credential-masking checks reuse it rather
+    than matching the raw password.
+    """
+    return value.replace("\\", "\\\\").replace("'", "\\'")
+
+
+def _sql_literal(value: str) -> str:
+    """Escape a string for embedding inside a single-quoted DuckDB SQL literal.
+
+    The conninfo is interpolated into ``ATTACH '...'`` / ``postgres_scan('...')``
+    SQL, so any single quote it contains (e.g. around a libpq-quoted password)
+    must be doubled or it terminates the SQL string literal.
+    """
+    return value.replace("'", "''")
+
+
 def gen_wrds_connection_info(user, password: str | None = None) -> str:
     """Build a libpq conninfo for WRDS.
 
@@ -685,18 +707,49 @@ def gen_wrds_connection_info(user, password: str | None = None) -> str:
         f"host={WRDS_HOST}",
         f"port={WRDS_PORT}",
         f"dbname={WRDS_DB}",
-        f"user={user}",
+        # Quote the username too: a space or quote in it would otherwise break
+        # libpq's conninfo parsing exactly as an unquoted password would.
+        f"user='{_pg_escape_value(user)}'",
     ]
     if password is not None:
-        parts.append(f"password={password}")
+        # Single-quote and escape so a password containing spaces or special
+        # characters can't break the conninfo (or split the value, which would
+        # defeat the password-masking check in _attach_wrds). The conninfo is
+        # itself embedded in a single-quoted SQL literal at each use site, so
+        # callers must additionally pass it through _sql_literal.
+        parts.append(f"password='{_pg_escape_value(password)}'")
     parts.append("sslmode=require")
     return " ".join(parts)
+
+
+def _password_forms(password: str) -> tuple[str, str, str]:
+    """The forms the password can take on its way into an error message: raw, the
+    libpq-escaped conninfo form (echoed by a connection IOException), and the
+    SQL-escaped-then-libpq-escaped form (echoed from the raw statement text by a
+    parser error). Ordered most-escaped first so redaction replaces the longest
+    match before its shorter substrings."""
+    escaped = _pg_escape_value(password)
+    return (_sql_literal(escaped), escaped, password)
+
+
+def _password_in_error(text: str, password: str) -> bool:
+    """True if the password appears in ``text`` in any of the forms it can take in
+    an error message (see :func:`_password_forms`)."""
+    return any(form in text for form in _password_forms(password))
+
+
+def _redact_password(text: str, password: str) -> str:
+    """Replace the password with ``***`` in every form it can take in an error
+    message (see :func:`_password_forms`)."""
+    for form in _password_forms(password):
+        text = text.replace(form, "***")
+    return text
 
 
 def get_columns(conn, conninfo, lib, table):
     cols = conn.execute(f"""
         SELECT *
-        FROM postgres_scan('{conninfo}', '{lib}', '{table}')
+        FROM postgres_scan('{_sql_literal(conninfo)}', '{lib}', '{table}')
         LIMIT 0
     """).description
     return [c[0] for c in cols]
@@ -791,7 +844,7 @@ def download_wrds_table(
     duckdb_conn.execute(f"""
         COPY (
           SELECT {projection}
-          FROM postgres_scan('{conninfo}', '{lib}', '{table}')
+          FROM postgres_scan('{_sql_literal(conninfo)}', '{lib}', '{table}')
           {where_clause}
         )
         TO '{filename}' (FORMAT PARQUET);
@@ -859,9 +912,9 @@ def _attach_wrds(con: duckdb.DuckDBPyConnection, conninfo: str, password: str | 
     password-free error. Errors that don't contain the password propagate unchanged.
     """
     try:
-        con.execute(f"ATTACH '{conninfo}' AS wrds (TYPE postgres, READ_ONLY)")
+        con.execute(f"ATTACH '{_sql_literal(conninfo)}' AS wrds (TYPE postgres, READ_ONLY)")
     except Exception as e:
-        if password and password in str(e):
+        if password and _password_in_error(str(e), password):
             raise RuntimeError(
                 "Failed to attach WRDS connection. Check credentials and MFA approval."
             ) from None
@@ -1114,7 +1167,7 @@ def _attach_download_worker(
                     )
                 except Exception as e:  # noqa: BLE001
                     # Redact only the credential, keeping the rest of the error for diagnostics.
-                    msg = str(e).replace(password, "***") if password else str(e)
+                    msg = _redact_password(str(e), password) if password else str(e)
                     with errors_lock:
                         task_errors.append(f"{task.table} ({Path(task.out).name}): {msg}")
         finally:
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -34,6 +34,7 @@ from .compustat_correction import (
 from .config import COLLECT_CHUNK_SIZE, CORRECTION_SPILL_COMPRESSION, END_DATE, MAIN_FILTERS
 from .output_writer import write_dataframe
 from .paths import DataPaths
+from .wrds_credentials import WRDS_DB, WRDS_HOST, WRDS_PORT
 
 
 def fl_none():
@@ -674,12 +675,22 @@ def gen_crsp_sf(paths: DataPaths, freq):
     return result
 
 
-def gen_wrds_connection_info(user, password):
-    return (
-        f"host=wrds-pgdata.wharton.upenn.edu "
-        f"port=9737 dbname=wrds "
-        f"user={user} password={password} sslmode=require"
-    )
+def gen_wrds_connection_info(user, password: str | None = None) -> str:
+    """Build a libpq conninfo for WRDS.
+
+    When ``password`` is ``None`` the ``password=`` field is omitted, so libpq
+    authenticates from ``$PGPASSFILE`` / ``~/.pgpass`` instead.
+    """
+    parts = [
+        f"host={WRDS_HOST}",
+        f"port={WRDS_PORT}",
+        f"dbname={WRDS_DB}",
+        f"user={user}",
+    ]
+    if password is not None:
+        parts.append(f"password={password}")
+    parts.append("sslmode=require")
+    return " ".join(parts)
 
 
 def get_columns(conn, conninfo, lib, table):
@@ -840,7 +851,7 @@ def _chunk_path(filename: str, i: int) -> str:
     return str(p.with_name(f"{p.stem}.part{i:02d}{p.suffix}"))
 
 
-def _attach_wrds(con: duckdb.DuckDBPyConnection, conninfo: str, password: str) -> None:
+def _attach_wrds(con: duckdb.DuckDBPyConnection, conninfo: str, password: str | None) -> None:
     """ATTACH the WRDS Postgres database read-only on an existing DuckDB connection.
 
     DuckDB's postgres extension embeds the full connection string (including the password) in
@@ -953,7 +964,7 @@ def _compute_histograms(
     date_columns: dict[str, str],
     end_date: date | None,
     max_conns: int,
-    password: str,
+    password: str | None,
 ) -> dict[str, list[tuple[int, int]]]:
     """Concurrently compute per-year row histograms for ``tables`` (each over its own connection).
 
@@ -1048,7 +1059,7 @@ def _remove_chunk_parts(filename: str) -> None:
 def _attach_download_worker(
     task_queue: queue.Queue[_DownloadTask],
     conninfo: str,
-    password: str,
+    password: str | None,
     task_errors: list[str],
     startup_errors: list[str],
     errors_lock: threading.Lock,
@@ -1117,7 +1128,7 @@ def _download_tables_parallel(
     conninfo: str,
     date_columns: dict[str, str],
     end_date: date | None,
-    password: str,
+    password: str | None,
     workers: int,
     split_tables: frozenset[str] = frozenset(),
     n_chunks: int = 1,
@@ -1252,7 +1263,7 @@ def _download_tables_parallel(
 def download_raw_data_tables(
     paths: DataPaths,
     username: str,
-    password: str,
+    password: str | None,
     end_date: date | None = None,
     persistent_connection: bool = False,
     max_workers: int = 1,
--- a/src/jkp/data/wrds_credentials.py
+++ b/src/jkp/data/wrds_credentials.py
@@ -135,13 +135,27 @@ def _keyring_set(username: str, password: str) -> bool:
         return False
 
 
-def _keyring_delete(username: str) -> bool:
-    """Delete the keyring entry; return False if absent or no backend."""
+def _keyring_delete(username: str) -> bool | None:
+    """Delete the keyring entry.
+
+    Returns True if one was removed, False when there is definitively nothing to
+    delete (no entry, or no keyring backend at all — in both cases nothing
+    survives), and None when the backend is present but unreachable (locked): the
+    entry may still exist, so we warn and let the caller avoid claiming a clean
+    success it can't guarantee.
+    """
     try:
         keyring.delete_password(SERVICE_NAME, username)
         return True
     except (keyring.errors.NoKeyringError, keyring.errors.PasswordDeleteError):
-        return False
+        return False  # no keyring, or nothing to delete
+    except keyring.errors.KeyringError as exc:
+        print(
+            f"Warning: could not reach the system keyring ({exc}); a stored entry may still exist.",
+            file=sys.stderr,
+            flush=True,
+        )
+        return None
 
 
 # --------------------------------------------------------------------------- #
@@ -562,7 +576,10 @@ def reset_credentials(full_reset: bool = False) -> None:
             print(f"Deleted password for '{username}' from the system keyring")
         if removed_pgpass:
             print(f"Removed WRDS entry for '{username}' from {_pgpass_path()}")
-        if not (removed_keyring or removed_pgpass):
+        # Only claim nothing was found when the keyring answer was definitive
+        # (False, not None): an unreachable backend already warned that an entry
+        # may survive, so a "nothing found" line here would contradict it.
+        if removed_keyring is False and not removed_pgpass:
             print(f"No stored password found for '{username}'")
 
 
--- a/src/jkp/data/wrds_credentials.py
+++ b/src/jkp/data/wrds_credentials.py
@@ -425,18 +425,23 @@ def _migrate_legacy_keyring(username: str) -> None:
 
 
 def _resolve_username(env_user: str | None = None) -> str:
-    if env_user:
+    if env_user and env_user.strip():  # ignore a whitespace-only WRDS_USERNAME
         return env_user.strip()
     if LAST_USER_FILE.exists():
-        return LAST_USER_FILE.read_text().strip()
+        cached = LAST_USER_FILE.read_text().strip()
+        if cached:  # treat an empty cache as missing rather than user ''
+            return cached
     if _LEGACY_USER_FILE.exists():  # migrate ~/.wrds_user -> state dir
         username = _LEGACY_USER_FILE.read_text().strip()
-        _persist_username(username)
-        with contextlib.suppress(OSError):
-            _LEGACY_USER_FILE.unlink()
-        return username
+        if username:  # an empty legacy file is treated as missing, not cached as ''
+            _persist_username(username)
+            with contextlib.suppress(OSError):
+                _LEGACY_USER_FILE.unlink()
+            return username
     if _interactive():
         username = input(f"Username for {SERVICE_NAME}: ").strip()
+        if not username:  # don't cache an empty username entered by mistake
+            raise RuntimeError(f"Empty username entered for {SERVICE_NAME}.")
         _persist_username(username)
         return username
     raise RuntimeError(
@@ -448,6 +453,15 @@ def _resolve_username(env_user: str | None = None) -> str:
 def _prompt_and_store(username: str) -> Credentials:
     """Interactively obtain a password and persist it to the best available store."""
     password = getpass.getpass(f"Password or token for {username} at {SERVICE_NAME}: ")
+    if not password:
+        # An empty password would be stored as a junk ``…:username:`` pgpass line
+        # that later "resolves" while auth silently fails — reject it, mirroring
+        # the empty-username guard.
+        raise RuntimeError(f"Empty password entered for {username}; nothing stored.")
+    # Cache the username alongside the stored password, so a credential provisioned
+    # for an env-provided WRDS_USERNAME (which _resolve_username does not persist)
+    # is still revocable via `jkp connect --reset` rather than orphaned.
+    _persist_username(username)
     if _keyring_set(username, password):
         print(
             f"Stored WRDS password for '{username}' in the system keyring.",
@@ -485,14 +499,16 @@ def get_wrds_credentials() -> Credentials:
       4. If nothing is found and a terminal is available, prompt and store;
          otherwise raise with actionable guidance.
     """
-    env_user = os.environ.get(ENV_USERNAME)
+    # Strip before the truthiness check so WRDS_USERNAME="   " is treated as
+    # unset rather than accepted as username="".
+    env_user = (os.environ.get(ENV_USERNAME) or "").strip() or None
     env_pw = os.environ.get(ENV_PASSWORD)
     if env_user and env_pw:
         # Literal env-var names in the log message (not the ENV_* constants) so no
         # "password"-named identifier flows into a logging sink — the value is
         # never logged, only the source label.
         _log_source("the WRDS_USERNAME/WRDS_PASSWORD environment variables")
-        return Credentials(env_user.strip(), env_pw)
+        return Credentials(env_user, env_pw)  # env_user already stripped/non-empty
 
     username = _resolve_username(env_user)
 
--- a/src/jkp/data/wrds_credentials.py
+++ b/src/jkp/data/wrds_credentials.py
@@ -387,19 +387,15 @@ def _migrate_legacy_keyring(username: str) -> None:
         )
         return
 
-    value = None
+    # Only migrate jkp's own exact-key entry. Since jkp always stored the entry
+    # via keyring.set_password (the same escaping _escape_keyring_option
+    # reproduces), the exact-key lookup covers everything jkp ever wrote; a
+    # fallback for a mismatched key would only serve a hypothetical rename and
+    # risks migrating another user's password under the resolved username.
     option = _escape_keyring_option(username)
-    if cp.has_option(SERVICE_NAME, option):
-        value = cp.get(SERVICE_NAME, option)
-    else:
-        items = cp.items(SERVICE_NAME)
-        if len(items) == 1:
-            # Exactly one stored user: take it even if the key doesn't match the
-            # resolved username. jkp only ever stored the current user's entry,
-            # so this rescues a renamed/re-escaped username. Risk if the file
-            # somehow holds one *other* user's entry: its password migrates under
-            # the resolved username and auth fails (recoverable via `jkp connect`).
-            value = items[0][1]
+    if not cp.has_option(SERVICE_NAME, option):
+        return
+    value = cp.get(SERVICE_NAME, option)
     if not value:
         return
     try:
--- a/src/jkp/data/wrds_credentials.py
+++ b/src/jkp/data/wrds_credentials.py
@@ -1,207 +1,564 @@
 """WRDS credential resolution.
 
-Credentials are resolved in a defined precedence order: environment variables,
-then the system keyring, then an opt-in file-backed keyring. Run
-``jkp connect --help`` for the full order and the ``JKP_ALLOW_PLAINTEXT_KEYRING``
-opt-in details.
+Username and password are resolved independently:
 
-The file-backed keyring is never selected automatically: the user opts in with
-``JKP_ALLOW_PLAINTEXT_KEYRING=1``, and the swap happens only inside this module's
-credential-resolution functions — importing ``jkp.data`` never changes the
-process-wide keyring backend.
+* **Username** — ``WRDS_USERNAME`` → saved state file → interactive prompt.
+* **Password** — ``WRDS_PASSWORD`` → system keyring → ``$PGPASSFILE`` / ``~/.pgpass``.
+
+When the password comes from a libpq password file, jkp never handles it: the
+connection string omits ``password=`` and libpq reads the file itself
+(:attr:`Credentials.password` is ``None``). The resolved source is printed to
+stderr on every run so it is always clear which one was used.
+
+On a headless node with no system keyring daemon, provision credentials by
+running ``jkp connect`` on a login node (it writes ``~/.pgpass``), by exporting
+``WRDS_USERNAME``/``WRDS_PASSWORD``, or by creating ``~/.pgpass`` directly — it
+is the standard Postgres/WRDS mechanism and any tool can populate it.
+
+Migration (one-time, automatic, non-interactive): a legacy plaintext-keyring
+entry (``~/.local/share/python_keyring/keyring_pass.cfg``, written by the removed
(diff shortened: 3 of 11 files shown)


src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,564 @@
...
-class _WrdsKeyring:
-    """WRDS-scoped wrapper over the keyring library.
 
-    The one place that speaks keyring's vocabulary for credential storage:
-    callers get/set/delete the WRDS password without naming ``SERVICE_NAME`` or
-    handling keyring's exception types. Handed out only via ``_keyring_backend``.
-    """
+def _log_source(desc: str) -> None:
+    """Announce which credential source authenticated (never the secret itself)."""
+    print(f"Using WRDS credentials from {desc}.", file=sys.stderr, flush=True)
```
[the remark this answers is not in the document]
capellini: False positive: `_log_source` only ever receives a source label, never the secret. CodeQL's name-based taint flagged the `ENV_PASSWORD` constant (whose value is the env-var name `WRDS_PASSWORD`, not a password) flowing into the log. I replaced that interpolation with a literal so no password-named identifier reaches a log sink. (The remaining `.pgpass` log interpolates only a file path.)

capellini:
@claude please review and make suggestions if necessary

capellini:
> 1. `wrds_credentials.py:500-513` — the one-time legacy-keyring migration (and the `_cleanup_legacy_keyring_section` call at line 496) only runs when password resolution reaches the keyring/pgpass path. If a user always authenticates via `WRDS_USERNAME`/`WRDS_PASSWORD` env vars, a lingering legacy `keyring_pass.cfg` `[WRDS]` section would never get cleaned up. Very low impact (it's dead weight, not a security regression — the plaintext file already existed under the old opt-in), but worth knowing if a lingering legacy file is ever a support question.

This is a bit of an obscure edge case and not worth adding code complexity to "fix." Once the user decides to stop using the env vars, the file will be migrated. If they never decide to stop using the env vars, there is no need to rename the files.

> 2. `cli.py`'s `connect` command doesn't actually validate connectivity to WRDS (it only resolves credentials, per `get_wrds_credentials()`) — the docstring says "Test or configure the WRDS connection." This predates this PR (verified against `origin/main`) so it's out of scope here, just flagging in case it's worth a follow-up issue.

As mentioned, this predates the PR, though it is a good idea for a follow-up issue IMO.

src/jkp/data/aux_functions.py
```diff
@@ -670,12 +671,22 @@ def gen_crsp_sf(paths: DataPaths, freq):
...
+    authenticates from ``$PGPASSFILE`` / ``~/.pgpass`` instead.
+    """
+    parts = [
+        f"host={WRDS_HOST}",
+        f"port={WRDS_PORT}",
+        f"dbname={WRDS_DB}",
+        f"user={user}",
+    ]
+    if password is not None:
+        parts.append(f"password={password}")
```
FernandoRDLL: The password goes into the conninfo raw, so a password containing a space, single quote, or backslash produces an invalid connection string (libpq parses the remainder as a bogus keyword). It also defeats the masking check in `_attach_wrds` — once libpq quotes only the offending fragment in its error, `password in str(e)` misses and fragments of the password can reach the logs. libpq accepts single-quoted values with `\'` and `\\` escaping, which handles all of these:

```suggestion
    if password is not None:
        escaped = password.replace("\\", "\\\\").replace("'", "\\'")
        parts.append(f"password='{escaped}'")
```

Worth noting the PR already escapes on the `.pgpass` side (`_esc_pgpass`) — this is the one spot where the secret is still interpolated verbatim.

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+            user_file.unlink()
+
+    if username is None:
         print("No stored username found — nothing to reset.")
+        return
+    print(f"Removed stored username '{username}'")
+
+    if full_reset:
+        removed_keyring = _keyring_delete(username)
+        removed_pgpass = _remove_pgpass_entry(username)
```
FernandoRDLL: I think `full_reset` also needs to clear the legacy plaintext store. As written it only touches the system keyring and `~/.pgpass`, so if a `keyring_pass.cfg` `[WRDS]` section still exists (the exact population this PR migrates), the next `get_wrds_credentials()` run finds it, and `_migrate_legacy_keyring` writes the just-deleted password straight back into `~/.pgpass` — without prompting. For someone rotating a compromised password via `jkp connect --reset` && `jkp connect`, the revoked secret both persists on disk and becomes the active credential again.

```suggestion
    if full_reset:
        removed_keyring = _keyring_delete(username)
        removed_pgpass = _remove_pgpass_entry(username)
        _cleanup_legacy_keyring_section()
```

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
 
-    def set_password(self, username: str, password: str) -> None:
+
+def _keyring_set(username: str, password: str) -> bool:
+    """Store the password in the system keyring; return False if none is available."""
+    try:
         keyring.set_password(SERVICE_NAME, username, password)
+        return True
+    except keyring.errors.NoKeyringError:
+        return False
```
FernandoRDLL, line 144: Catching only `NoKeyringError` here covers the daemon-entirely-absent case, but a present-but-broken backend raises siblings of it: SecretService raises `InitError`/`KeyringLocked` when the collection is locked (headless SSH with gnome-keyring up is the classic case), and KWallet raises `KeyringLocked`. Those propagate out of the unguarded `_keyring_get` call in `get_wrds_credentials` before the `~/.pgpass` fallback is ever consulted — so the exact headless scenario this PR targets can still crash with a keyring traceback despite a valid pgpass entry two steps down the chain. Catching the common base class keeps the intent and covers all of them:

```suggestion
def _keyring_get(username: str) -> str | None:
    try:
        return keyring.get_password(SERVICE_NAME, username)
    except keyring.errors.KeyringError:
        return None


def _keyring_set(username: str, password: str) -> bool:
    """Store the password in the system keyring; return False if none is usable."""
    try:
        keyring.set_password(SERVICE_NAME, username, password)
        return True
    except keyring.errors.KeyringError:
        return False
```

slurm/submit_job_som_hpc.slurm
```diff
@@ -30,17 +30,20 @@ export NUMEXPR_NUM_THREADS=${SLURM_CPUS_PER_TASK}
...
 #
-# Alternative (for shared service accounts or container deployments): unset
-# JKP_ALLOW_PLAINTEXT_KEYRING and export WRDS_USERNAME / WRDS_PASSWORD instead,
-# e.g. by sourcing a mode-600 ~/.wrds_secrets file with `export WRDS_USERNAME=...`.
-export JKP_ALLOW_PLAINTEXT_KEYRING=1
+# The pipeline reads the password from ~/.pgpass via libpq; jkp never stores a
+# separate copy. Point PGPASSFILE at a non-default location if you prefer.
+#
+# Fully non-interactive alternative (no login-node terminal): export
+# WRDS_USERNAME / WRDS_PASSWORD, or create ~/.pgpass yourself (chmod 600):
```
FernandoRDLL: The second alternative doesn't work on its own: `get_wrds_credentials` resolves the username *before* the pgpass lookup, and `_resolve_username` never parses it out of `~/.pgpass`. A user who only hand-creates the file (no `jkp connect` run, no `WRDS_USERNAME`) gets `RuntimeError: No WRDS username available and no terminal to prompt at` at job start, despite a fully valid entry. The comment should pair the pgpass route with exporting the username:

```suggestion
# Fully non-interactive alternative (no login-node terminal): export
# WRDS_USERNAME / WRDS_PASSWORD; or export WRDS_USERNAME and create
# ~/.pgpass yourself (chmod 600):
```

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+        username = _LEGACY_USER_FILE.read_text().strip()
+        _persist_username(username)
+        with contextlib.suppress(OSError):
+            _LEGACY_USER_FILE.unlink()
+        return username
+    if _interactive():
         username = input(f"Username for {SERVICE_NAME}: ").strip()
-        LAST_USER_FILE.write_text(username)
+        _persist_username(username)
+        return username
```
FernandoRDLL, line 538: Hitting Enter at the prompt persists `""`, and the cached-read path guards on file existence rather than content — so an accidental empty username is cached forever, every later run skips the prompt, and auth fails for user `''` with nothing re-asking until the user discovers `--reset`. Treating an empty cache as missing and rejecting empty input closes both ends:

```suggestion
    if LAST_USER_FILE.exists():
        cached = LAST_USER_FILE.read_text().strip()
        if cached:
            return cached
    if _LEGACY_USER_FILE.exists():  # migrate ~/.wrds_user -> state dir
        username = _LEGACY_USER_FILE.read_text().strip()
        _persist_username(username)
        with contextlib.suppress(OSError):
            _LEGACY_USER_FILE.unlink()
        return username
    if _interactive():
        username = input(f"Username for {SERVICE_NAME}: ").strip()
        if not username:
            raise RuntimeError(f"Empty username for {SERVICE_NAME}.")
        _persist_username(username)
        return username
```

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+        return
+    if not cp.has_section(SERVICE_NAME):
+        return
 
-    if LAST_USER_FILE.exists():
-        username = LAST_USER_FILE.read_text().strip()
+    # If ~/.pgpass already resolves for this user, that is their current
+    # credential — never clobber it with the (possibly stale) legacy value.
+    # Just drop the now-superseded legacy [WRDS] section.
+    if _pgpass_has_entry(username):
```
FernandoRDLL: This guard has a gap that can destroy the user's current password: `_pgpass_has_entry` returns False when the file's permissions are looser than 0600, so a 0644 `~/.pgpass` holding the user's *current* WRDS line skips the never-overwrite branch, and `_write_pgpass` replaces that line with the (possibly stale) legacy value — then `_atomic_write` chmods to 600, making the wrong entry the active one. That's the opposite of the docstring's "never overwritten" promise, and it's silent apart from the chmod warning.

The guard wants "does an entry exist", not "would libpq use it". Maybe split the two concerns:

```python
def _pgpass_has_entry(username: str, *, check_perms: bool = True) -> bool:
    ...
    if check_perms and _perms_too_open(path):
        ...
        return False
```

and call `_pgpass_has_entry(username, check_perms=False)` here, so a loose-permission file still counts as the user's current credential (the perms warning still fires on the resolution path at the bottom of `get_wrds_credentials`).

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
-        stacklevel=2,
-    )
-    from keyrings.alt.file import PlaintextKeyring  # noqa: PLC0415
+    password: str | None  # None => authenticate via ~/.pgpass (omit password=)
 
-    keyring.set_keyring(PlaintextKeyring())
 
+# --------------------------------------------------------------------------- #
+# Small helpers
+# --------------------------------------------------------------------------- #
```
FernandoRDLL: We stripped dashed section banners from the codebase a while back (32aee0c) and no other file under `src/jkp/data/` uses them — could we drop these too (also at 108, 136, 282, 414, and the mirrored ones in `tests/unit/test_wrds_credentials.py`)? Docstrings on the functions already carry the structure.

```suggestion
```

FernandoRDLL (commented):
Ran a deep review over this. The pgpass design and the end-to-end HPC validation look solid, and the deliberate behavior changes (decoupled env resolution, tty-gated prompting, one-way migration) all match the PR description, so I haven't flagged those.

Inline comments fall into three buckets: a handful of correctness issues in credential lifecycle (reset/migration interplay, conninfo escaping, keyring error handling, the slurm doc path), some cleanup (banners, the duplicate `__main__` CLI, scope of the platformdirs move), and a few narrow edge cases marked as such — take or leave those. Most comments carry a committable suggestion.

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+        help="Remove stored username and password (keyring and ~/.pgpass entry).",
     )
     args = parser.parse_args()
 
     if args.reset:
-        reset_credentials(full_reset=args.reset)
+        reset_credentials(full_reset=True)
     else:
         creds = get_wrds_credentials()
         print(f"Using credentials for '{creds.username}'")
```
FernandoRDLL: This argparse block is a second CLI for the same operations as `jkp connect` — and this PR already had to edit both in lockstep (reset semantics + help text), which is the maintenance cost in action. Since cli.py declares itself the canonical entry point, I'd delete this block (and the now-unused `import argparse` up top):

```suggestion
```

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+SERVICE_NAME = "WRDS"  # keyring service name and legacy plaintext-keyring section
 
-# Environment variable names.
 ENV_USERNAME = "WRDS_USERNAME"
 ENV_PASSWORD = "WRDS_PASSWORD"
-ENV_ALLOW_PLAINTEXT = "JKP_ALLOW_PLAINTEXT_KEYRING"
+ENV_PGPASSFILE = "PGPASSFILE"
+
+# Username cache: a per-OS state file (auto-migrated from the old ~/.wrds_user).
+LAST_USER_FILE = Path(platformdirs.user_state_dir("jkp")) / "wrds_user"
```
FernandoRDLL, line 59: The username-cache move to a platformdirs state dir feels orthogonal to the pgpass feature: it brings a new runtime dependency (this line is the sole use of `platformdirs`), a migration branch in `_resolve_username`, dual-file handling in `reset_credentials`, and its own tests. Was the relocation needed for something here? Keeping `~/.wrds_user` would drop the dependency and both migration paths — happy to be told there's a motivation I'm missing.

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+        items = cp.items(SERVICE_NAME)
+        if len(items) == 1:
+            # Exactly one stored user: take it even if the key doesn't match the
+            # resolved username. jkp only ever stored the current user's entry,
+            # so this rescues a renamed/re-escaped username. Risk if the file
+            # somehow holds one *other* user's entry: its password migrates under
+            # the resolved username and auth fails (recoverable via `jkp connect`).
+            value = items[0][1]
+    if not value:
+        return
```
FernandoRDLL: The single-entry fallback documents its own failure mode in the comment: if the file holds one *other* user's entry, that password migrates under the resolved username, the keyed entry is destroyed, and auth just fails. Since jkp always stored entries via `keyring.set_password` with the same escaping `_escape_keyring_option` reproduces, the exact-key lookup already covers everything jkp itself ever wrote — the fallback only serves a hypothetical renamed username. I'd drop it:

```suggestion
    option = _escape_keyring_option(username)
    if not cp.has_option(SERVICE_NAME, option):
        return
    value = cp.get(SERVICE_NAME, option)
    if not value:
        return
```

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
-    def get_password(self, username: str) -> str | None:
+
+def _atomic_write(path: Path, text: str, mode: int = 0o600) -> None:
+    """Write ``text`` to ``path`` atomically with restrictive permissions."""
+    path.parent.mkdir(parents=True, exist_ok=True)
+    fd, tmp = tempfile.mkstemp(dir=str(path.parent))
+    try:
+        if hasattr(os, "fchmod"):
+            os.fchmod(fd, mode)
+        with os.fdopen(fd, "w") as fh:
```
FernandoRDLL: Two bits here that can go: no caller ever passes `mode`, and `tempfile.mkstemp` already creates the file 0600 on POSIX (and Windows has no `fchmod`, so that branch is a no-op there anyway):

```suggestion
def _atomic_write(path: Path, text: str) -> None:
    """Write ``text`` to ``path`` atomically; mkstemp creates it mode 600."""
    path.parent.mkdir(parents=True, exist_ok=True)
    fd, tmp = tempfile.mkstemp(dir=str(path.parent))
    try:
        with os.fdopen(fd, "w") as fh:
```

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+        _migrate_legacy_keyring(username)
+    except (OSError, ValueError) as exc:
+        print(
+            f"Warning: could not migrate the legacy WRDS credential ({exc}); "
+            "continuing without it.",
+            file=sys.stderr,
+            flush=True,
+        )
+
+    if _pgpass_has_entry(username):
```
FernandoRDLL: Small one: when a legacy `keyring_pass.cfg` exists, `_pgpass_has_entry` runs twice per resolution (once inside `_migrate_legacy_keyring`, once here), re-reading and re-parsing the file — and because the loose-permissions warning lives inside the predicate, it prints twice per run whenever the legacy section can't be migrated. Making the predicate pure and warning once at this call site (or computing the result once and passing it into the migration) would fix both.

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+        cp.write(buf)
+        _atomic_write(path, buf.getvalue())
+    else:
+        with contextlib.suppress(OSError):
+            path.unlink()
 
 
-def get_wrds_credentials() -> Credentials:
-    """Resolve WRDS credentials following the documented precedence order.
+def _cleanup_legacy_keyring_section() -> None:
```
FernandoRDLL: `_cleanup_legacy_keyring_section` repeats `_migrate_legacy_keyring`'s open-and-check preamble (exists / `RawConfigParser` / `read` / `has_section`), and the two already diverge on error policy — this one swallows `OSError`, the migration lets it propagate. A tiny shared `_load_legacy_cfg() -> RawConfigParser | None` would keep the parsing in one place; the callers can keep their different error stances.

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+    path = _pgpass_path()
+    if not path.exists():
+        return False
+    out: list[str] = []
+    removed = False
+    for raw in path.read_text().splitlines():
+        if _is_wrds_line(_split_pgpass(raw.strip()), username):
+            removed = True
+        else:
+            out.append(raw)
```
FernandoRDLL: This read/`_is_wrds_line`-filter/rewrite loop is the same one `_write_pgpass` runs (lines 245–253). If a future tweak to matching or file handling (encoding, CRLF, comment lines) lands in one and not the other, `--reset` starts removing a different set of lines than `connect` writes. A shared "partition into (wrds_lines, other_lines)" helper both could call would keep them in lockstep.

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+            and port in ("*", WRDS_PORT)
+            and db in ("*", WRDS_DB)
+            and user in ("*", username)
+        ):
+            return True
+    return False
+
+
+def _write_pgpass(username: str, password: str) -> Path:
+    """Append or replace jkp's WRDS line, leaving every other entry untouched."""
```
FernandoRDLL, line 326: nit: the new helpers use free-form docstrings rather than the repo's documented `Description:/Steps:/Output:` format (only `get_wrds_credentials` kept a `Steps:` section). The module was already free-form before this PR, so this is take-or-leave — just flagging the drift from the CLAUDE.md convention.

tests/unit/test_wrds_credentials.py
```diff
@@ -59,187 +53,652 @@ def _boom(*a, **kw):
...
+    def _fn(*a, **kw):
+        raise exc("no backend")
+
+    return _fn
+
+
+def _write_pgpass(mod, content):
+    path = mod._pgpass_path()
+    path.write_text(content)
+    if not __import__("sys").platform.startswith("win"):
```
FernandoRDLL: `__import__("sys")` inline reads oddly next to the local `import sys` two other tests already do — a top-level `import sys` next to `base64` would serve all three (same for the five local `import configparser`s, if you feel like it):

```suggestion
    if not sys.platform.startswith("win"):
```

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+            else:
+                out.append(raw)
+    if not replaced:
+        out.append(new_line)
+    _atomic_write(path, "\n".join(out) + "\n")
+    return path
+
+
+def _remove_pgpass_entry(username: str) -> bool:
+    """Remove jkp's WRDS line; return True if one was removed. Preserves others."""
```
FernandoRDLL, line 352: Minor asymmetry, probably fine to leave: resolution honors wildcard pgpass entries (`host in ("*", WRDS_HOST)` etc.), but removal only matches exact `host:port:db` lines. So a hand-written `*:*:*:user:pw` line survives `--reset`, and after a password rotation the README's reset-then-connect flow finds the stale wildcard entry and never re-prompts. Narrow (needs a hand-authored wildcard line), but maybe worth a warning here when a wildcard match remains after removal.

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+            if _is_wrds_line(_split_pgpass(raw.strip()), username):
+                if not replaced:
+                    out.append(new_line)
+                    replaced = True
+                # drop any duplicate WRDS lines for this user
+            else:
+                out.append(raw)
+    if not replaced:
+        out.append(new_line)
+    _atomic_write(path, "\n".join(out) + "\n")
```
FernandoRDLL, line 347: Edge case: if the file exists but isn't readable (mode 000, or root-owned from an old sudo), `_pgpass_has_entry` deliberately reports "no entry", so the user gets prompted — and then this `read_text()` raises a raw `PermissionError` right after they've typed their password, which ends up stored nowhere. A guarded read turns that into an actionable message:

```suggestion
    out: list[str] = []
    replaced = False
    existing: list[str] = []
    if path.exists():
        try:
            existing = path.read_text().splitlines()
        except OSError as exc:
            raise RuntimeError(
                f"Cannot read existing {path} ({exc}); fix its permissions and retry."
            ) from exc
    for raw in existing:
        if _is_wrds_line(_split_pgpass(raw.strip()), username):
            if not replaced:
                out.append(new_line)
                replaced = True
            # drop any duplicate WRDS lines for this user
        else:
            out.append(raw)
    if not replaced:
        out.append(new_line)
    _atomic_write(path, "\n".join(out) + "\n")
```

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
+def _atomic_write(path: Path, text: str, mode: int = 0o600) -> None:
+    """Write ``text`` to ``path`` atomically with restrictive permissions."""
+    path.parent.mkdir(parents=True, exist_ok=True)
+    fd, tmp = tempfile.mkstemp(dir=str(path.parent))
+    try:
+        if hasattr(os, "fchmod"):
+            os.fchmod(fd, mode)
+        with os.fdopen(fd, "w") as fh:
+            fh.write(text)
+        os.replace(tmp, path)
```
FernandoRDLL, line 113: Niche one: `os.replace` swaps a symlinked destination for a regular file, so a stow/chezmoi-managed `~/.pgpass` gets silently detached from its dotfiles source. Writing through the link is a one-word change:

```suggestion
        os.replace(tmp, os.path.realpath(path))
```

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,567 @@
...
-        except keyring.errors.PasswordDeleteError:
-            return False
+
+def _keyring_delete(username: str) -> bool:
+    """Delete the keyring entry; return False if absent or no backend."""
+    try:
+        keyring.delete_password(SERVICE_NAME, username)
         return True
+    except (keyring.errors.NoKeyringError, keyring.errors.PasswordDeleteError):
+        return False
```
FernandoRDLL: When the backend is merely unreachable (SSH session, no DBus) rather than absent, this returns False and `reset_credentials` reports "No stored password found" — while the entry actually survives in the keyring. For a security-motivated reset that's a misleading success message; the old code deliberately let `NoKeyringError` propagate for exactly this reason. Distinguishing the two keeps reset honest:

```suggestion
def _keyring_delete(username: str) -> bool:
    """Delete the keyring entry; return False if absent or no backend."""
    try:
        keyring.delete_password(SERVICE_NAME, username)
        return True
    except keyring.errors.PasswordDeleteError:
        return False
    except keyring.errors.KeyringError as exc:
        print(
            f"Warning: could not reach the system keyring ({exc}); "
            "a stored entry may still exist.",
            file=sys.stderr,
            flush=True,
        )
        return False
```

src/jkp/data/wrds_credentials.py
```diff
@@ -1,207 +1,697 @@
...
+            return False, False
+        lines = path.read_text(encoding="utf-8").splitlines()
+    except UnicodeDecodeError:
+        # jkp reads the file as UTF-8 — a deliberate choice, since libpq itself is
+        # byte-oriented and mandates no encoding. A file another tool wrote in a
+        # different encoding isn't "invalid"; jkp just can't read it here, so
+        # ignore it (with a warning) rather than crash the whole resolution.
+        print(
+            f"Warning: jkp reads {path} as UTF-8, but it is not valid UTF-8, so jkp "
+            f"cannot use it. Set {ENV_PASSWORD} or use the system keyring.",
```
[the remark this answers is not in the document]
capellini: False positive, same class as the earlier one. I replaced the interpolation with the literal `WRDS_PASSWORD` (with a comment so it does not get reverted), so no password-named identifier reaches the sink. Fixed in `afc4de0`.

src/jkp/data/aux_functions.py
capellini: Applied, and it grew in review. Your libpq single-quoting is in (`_pg_escape_value`), and a follow-up caught that the conninfo is itself embedded in a single-quoted SQL literal at the `ATTACH`/`postgres_scan` sites, so a quoted password terminated the SQL string (a `ParserException` on every password) until it is also SQL-escaped (`_sql_literal`). Both layers are covered now, with a network-free parse test plus a real-DuckDB `ATTACH` integration test. On the masking point you raised, the error echoes the password in its libpq-escaped form, so `_password_in_error` and `_redact_password` match that form (and the SQL-doubled composition form), pinned against a real DuckDB auth-failure echo. I also quoted the `user` field, which would break libpq the same way. `217e9b4`

src/jkp/data/wrds_credentials.py
capellini: Good catch. This is the rotation path (`--reset` then `connect`) writing the revoked password straight back via `_migrate_legacy_keyring`. `reset_credentials` now clears the legacy plaintext entry alongside the keyring and pgpass removals, so nothing re-migrates after a reset. Done in `d72e4d6`.

src/jkp/data/wrds_credentials.py
capellini: Agreed. A locked-but-present SecretService or KWallet is exactly the headless case, and `NoKeyringError` alone misses it. `_keyring_get` and `_keyring_set` now catch the base `KeyringError` and return None, treating any backend failure (locked, no DBus, init error) as "no keyring credential", so resolution falls through to `~/.pgpass` instead of surfacing a traceback. Done in `78239aa`.

slurm/submit_job_som_hpc.slurm
capellini: You're right that the hand-create-`~/.pgpass` route can't stand on its own, since `_resolve_username` never reads the username out of the file. Rather than always require `WRDS_USERNAME`, I dropped that alternative. The login-node `jkp connect` step already stores both the username and the pgpass entry, so the remaining fully non-interactive path is just exporting `WRDS_USERNAME` and `WRDS_PASSWORD`. I also fixed the same overclaim in the module docstring, which still presented a hand-created `~/.pgpass` as a standalone provisioning path. Done in `a2e7a34`.

src/jkp/data/wrds_credentials.py
capellini: Agreed on both ends. `_resolve_username` now treats an empty or whitespace cache as missing and rejects empty prompt input with a clear error, so a stray Enter cannot get persisted and skip the prompt on every later run. A later pass found the same gap on the legacy `~/.wrds_user` migration branch and on a whitespace-only `WRDS_USERNAME`, so those paths share the empty check too. Done in `762d879`.

src/jkp/data/wrds_credentials.py
capellini: Good catch that the guard asked "would libpq use it" rather than "does an entry exist". The migration defer-check now calls `_pgpass_has_entry(username, check_perms=False)`, so an existing WRDS line counts as the user's current credential regardless of mode and is never overwritten. The perms warning still fires on the resolution path. Worth noting that a 0644 `~/.pgpass` would not have authenticated in the first place (libpq ignores it), so this is about not clobbering a line the user still needs to `chmod`, not a working setup. Done in `f3e50bb`.

src/jkp/data/wrds_credentials.py
capellini: Done. Removed the dashed banners here and the mirrored ones in the test file. The function docstrings carry the structure. `dc8f5d7`.

src/jkp/data/wrds_credentials.py
capellini: Removed the block and the now-unused `import argparse`. Done in `05b1285`.

src/jkp/data/wrds_credentials.py
capellini: It is less orthogonal than it looks. The PR's theme is getting jkp's credential state out of project-bespoke files and into more standardized and accepted locations. The secret moves from the plaintext keyring to `~/.pgpass` (the standard libpq/Postgres file), and the non-secret username cache moves from `~/.wrds_user` (a bespoke home-root file) to a proper per-OS state dir. Doing one and leaving the other keeps the ad-hoc pattern we are retiring. `platformdirs` is now an explicit dependency in `pyproject.toml` (not incidental), and the `_resolve_username` branch exists to relocate existing `~/.wrds_user` files silently. I would prefer to keep it here since it is the same change.

src/jkp/data/wrds_credentials.py
capellini: Agreed. jkp only ever wrote entries under `_escape_keyring_option(username)`, so the exact-key lookup already covers everything it created. The fallback only served a hypothetical renamed user and could migrate the wrong password. Dropped it. Done in `2ac571e`.

src/jkp/data/wrds_credentials.py
capellini: Half-applied. I dropped the unused `mode` parameter, since no caller passes it. I kept the explicit `os.fchmod(fd, 0o600)` rather than leaning on `mkstemp`'s default, though. For a file holding a plaintext secret I would rather the 0600 be stated in the code than implied by the tempfile implementation, so I left it with a comment. Done in `8bb6364`.

src/jkp/data/wrds_credentials.py
capellini: Two halves here. The double-warning is already gone as a side effect of the loose-perms fix above. The migration guard now passes `check_perms=False`, so only the resolution path can emit the perms warning, and it prints once (`f3e50bb`). The double-read remains, but it only happens while a legacy `keyring_pass.cfg` still exists (the migration then clears it), so it is a one-time cost on a path that is on its way out. I am leaving that micro-optimization for now.

src/jkp/data/wrds_credentials.py
capellini: Agreed. Extracted `_load_legacy_cfg() -> RawConfigParser | None`, so the open/parse/`has_section` preamble lives in one place. To be precise about the error stances, both callers now swallow read/parse errors via the helper (and `configparser.read()` already ignores an unopenable file), so the difference between them is only on the drop/write side. `_cleanup_legacy_keyring_section` suppresses `OSError` there, while `_migrate_legacy_keyring` lets a write error propagate. Done in `8921777`.

src/jkp/data/wrds_credentials.py
capellini: Fair that the `--reset` removal loop and the `_write_pgpass` loop could drift. The matching itself is already shared, since both go through `_split_pgpass` and `_is_wrds_line`, so a change to what counts as a WRDS line lands in both. Given that, I am leaving the two small loop bodies rather than adding a partition abstraction, but noted if the file handling grows.

src/jkp/data/wrds_credentials.py
capellini: Fair flag, and thanks for it. You are right that these new helpers drifted from the repo's documented `Description:/Steps:/Output:` convention. I kept them free-form for now, since the module was already free-form and a one-line docstring reads clearly for small credential helpers, but I do not feel strongly about it.

Stepping back, that format works well for the pipeline functions, but it is a project-specific convention rather than a widely used standard, so it may not be the best fit for functions across the rest of the codebase. Rather than settle it here, it might be worth a separate issue on docstring-format standardization, perhaps moving toward a more widely adopted style like Google-style that reads well throughout the codebase rather than just the pipeline ones. There is no hard-and-fast rule that we must adopt another convention, but it feels like a good conversation to have as the codebase grows. Happy to open that issue if you think it is worthwhile.

tests/unit/test_wrds_credentials.py
capellini: Done. Hoisted `sys`, `base64`, and `configparser` to the module top and dropped the inline `__import__` and local imports. Done in `e280e9d`.

src/jkp/data/wrds_credentials.py
capellini: Agreed it is narrow, and I want `--reset` to leave a hand-authored `*:*:*:*` line alone, since it is not ours to delete. The deeper issue is that while such a wildcard matches, `jkp connect` cannot set or change a WRDS password at all, because resolution finds the wildcard usable and never prompts. I added a warning when WRDS is served only by a wildcard so it is at least visible. A follow-up pass also made the check honor libpq's first-match rule, so a wildcard sitting above jkp's exact line correctly triggers the warning too. Done in `97fb26d`.

src/jkp/data/wrds_credentials.py
capellini: Edge case (mode 000, or a root-owned leftover from `sudo`), but cheap to handle well. `_write_pgpass` now guards the read and raises an actionable "fix its permissions and retry" instead of a bare `PermissionError`. The same guard covers `_remove_pgpass_entry`, so `--reset` is handled too. `jkp connect` now catches these `RuntimeError`s and exits 1 with the message rather than a traceback, and `reset_credentials` drops the legacy plaintext section before touching `~/.pgpass`, so a failure there cannot leave a plaintext copy behind on a security reset. Done in `3f897b1`.

src/jkp/data/wrds_credentials.py
capellini: I am leaving the realpath rewrite out. A symlinked `~/.pgpass` almost always points into a dotfiles repo, and writing through the link would push a plaintext WRDS password into a tracked tree, which I would rather not silently encourage. Replacing the link with a regular file is the safer default of the two. You are right that it was silent, though, so `_atomic_write` now warns when it is about to replace a symlink, noting that the original target is detached and may still hold the old entry. That also covers the `--reset` path, where the "Removed WRDS entry" message would otherwise overstate what happened. Done in `8bb6364`.

src/jkp/data/wrds_credentials.py
capellini: Agreed on distinguishing unreachable-but-present from absent. One tweak to your version. `NoKeyringError` is a subclass of `KeyringError`, so catching the base class alone would print "a stored entry may still exist" even when there is genuinely no backend and nothing stored. So `_keyring_delete` treats `NoKeyringError` and `PasswordDeleteError` as a quiet `False` (nothing survives) and warns only for the reachable-but-locked case. While in there I also cleared up a related contradiction, where a locked keyring warned "an entry may still exist" and then printed "No stored password found". The latter now only prints on a definitive `False`. Done in `ae88d62`.

capellini:
Thanks so much for the **very** thorough review, @FernandoRDLL. It was genuinely valuable and caught things I would not have found on my own, from the two-layer conninfo quoting to the reset and migration edge cases.

I have replied inline on each thread. Every issue you flagged is now addressed in the branch, each fix in its own commit referenced in the replies, and the few I left as-is have a short rationale. I also validated the full set end-to-end on real WRDS (connection, masking, and a multi-user migration against a real keyrings.alt file), so these are exercised against the actual system and not just unit tests.

Thanks again for the careful pass!

FernandoRDLL (approved):

PR #247, opened 2026-07-14
Fix US daily coverage filters

### LINKED ISSUE #235: [Data]: US daily firm characteristics data under-covered
### Type of Discrepancy

Missing data (NaN/null where values expected)

### Affected Data Type

Firm characteristics

### Affected Characteristic(s)

_No response_

### Country/Region Affected

USA

### Start Date of Affected Period

1960-01-01

### End Date of Affected Period

2025-12-31

### Comparison Baseline

Other external source

## Summary

| Field | Value |
|-------|-------|
| **Type** | Missing data; differs from CRSP and recipe-faithful rebuild |
| **Scope** | US firm characteristics from daily returns (1960–2025; worst in 1970s, especially 1973–1982) |
| **Baseline** | December-2025 WRDS vintage vs. CRSP-only rebuild under JKP's own rolling-window recipe |

US daily-return characteristics are materially under-covered relative to CRSP. On 3,997,228 US stock-months (common, main exchange, primary security):

| Group | p10 | Mean |
|-------|-----|------|
| 21-day return family (`rvol`, `ivol`, `iskew`, `rskew`, `rmax*`, `beta_dimson`, …) | 0.55 | 0.80 |
| 1260-day family (`betabab_1260d`, `corr_1260d`) | 0.40 | 0.61 |

**Affected:** return-based rolling stats above; `ami_126d` (return channel only).  
**Not affected:** `turnover_126d`, `dolvol_126d`, `zero_trades_*`, `bidaskhl_21d`, `beta_60m`, `resff3_12_1` — their gaps match a CRSP rebuild and reflect data availability, not build bugs.

Neutralizing the three mechanisms below (present in both SAS and Python builds) raises post-1970 coverage for the 21-day family from **0.81 → 0.94** and 1970s coverage from **0.57 → 0.91**.

---

## Root cause

All three live in the daily prep path (`prepare_daily()`, `base_data_filter_exp()` in `aux_functions.py`):

1. **Market-return join** — `.filter(col("mktrf").is_not_null())` after joining daily factors drops valid Compustat-pipeline US rows when the factor file has gaps.
2. **Zero-return gate** — Stock-months with ≥10 zero-return days (`zero_obs`) are excluded from all return-based stats. CRSP's 1973–1982 spurious zero-return days cause whole stock-months to be dropped, not individual days.
3. **Return-gap cap** — Returns nulled when `ret_lag_dif > 14`; mainly binds on Compustat rows, reducing valid days in rolling windows.

Identical p10 across the 21-day family (0.548 to three decimals) is the fingerprint of mechanism (2).

---

## Secondary finding

JKP applies the **Gao–Ritter (2010)** NASDAQ dealer-volume adjustment in `adj_trd_vol_NASDAQ()` but does not document it. NYSE/AMEX dollar-volume matches a raw-CRSP rebuild exactly (ratio 1.000, 1990–2019); NASDAQ pre-2004 matches the Gao–Ritter divisor schedule. Minor boundary bug: `< 2003-12-31` leaves 2003-12-31 unadjusted.

---

## Proposed fixes

1. Make the zero-return gate source-conditional (or volume-based for CRSP US data).
2. Check factor-file completeness before the `mktrf` join; fall back to a complete series (e.g. Ken French) where available.
3. Document the Gao–Ritter adjustment; fix the 2003-12-31 boundary if intended.

**Validation:** 21-day post-1970 coverage ≥ 0.90; 1970s ≥ 0.85; no regression on volume-family coverage; NYSE/AMEX dollar-volume unchanged; `CHANGELOG_DATA.md` entry.

### Sample of Problematic Data

```markdown

```

### Code to Reproduce (Optional)

```python

```

### Date Code Was Run

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this discrepancy has not already been reported
- [x] I have verified my comparison baseline is reliable

miahdy:
## Summary

Fix US daily-characteristic undercoverage caused by a global data-quality filter being applied to CRSP observations.

- Apply the `zero_obs < 10` screen only to Compustat-sourced rows; CRSP-sourced rows are exempt.
- Include 2003-12-31 in the Gao–Ritter NASDAQ `/1.6` volume-adjustment window and document the adjustment.

The missing-country-market-return change (dropping the `mktrf` filter in `prepare_daily`) has been split out to [#252](https://github.com/bkelly-lab/jkp-data/pull/252): it is +0.0000 for the US and needs to be evaluated separately from the `zero_obs` gate. 

## Change Type

- [x] Methodology change (affects factor definitions or calculations)
- [ ] Documentation only

## Methodological Impact

The previous daily preparation logic removed valid CRSP observations: stock-months with at least ten zero-return days were excluded from all return-based rolling statistics. This screen was intended for Compustat Global data quality, but it also removed CRSP observations—particularly during 1973–1982.

The zero-return screen is now source-conditional: CRSP rows pass it, while Compustat rows retain the existing quality check. The `mktrf` filter in `prepare_daily` is unchanged in this PR (stock-days without a country market return are still dropped); relaxing it is handled separately in #252.

The Gao–Ritter NASDAQ adjustment now applies through 2003-12-31 inclusively, correcting the previous one-day boundary inconsistency. The `<=` boundary matches the documented schedule (the `/1.6` window runs through year-end 2003) but differs from the original SAS `<`, so published NASDAQ volume shifts on 2003-12-31.

## Data Impact

This intentionally increases coverage of US daily-return characteristics. Targets are from the [jkp_coverage_note_v2.pdf](https://github.com/user-attachments/files/30064752/jkp_coverage_note_v2.pdf); validation on the rebuilt 1960–2025 output shows:

- 21-day return-family post-1970 coverage: approximately `0.81` → `0.95` (note target: `0.94`)
- 21-day return-family 1970s coverage: approximately `0.57` → `0.91` (note target: `0.91`)
- `corr_1260d` and `betabab_1260d` post-1970 coverage: approximately `0.62` → `0.72`
- `ami_126d` post-1970 coverage: approximately `0.80` → `0.86`

Coverage of unaffected volume-based characteristics remains unchanged, including `turnover_126d`, `dolvol_126d`, `zero_trades_*`, and `bidaskhl_21d`.

The Gao–Ritter boundary correction may change NASDAQ volume-dependent values on 2003-12-31 and rolling characteristics whose windows include that date.

### Factor-output comparison

The rebuilt factor outputs in `data/processed/portfolios/` were compared with the previous previous undercovered outputs, matching observations by country, characteristic, and date.

For monthly `hml.parquet`:

- The rebuilt file contains 2,710,345 rows versus 2,712,226 previously, with 2,710,292 matched rows.
- Across the 153 value-weighted factor series, the pooled country-month correlation has a median of `1.000` and a minimum of `0.968`; 150 factors exceed `0.99`.
- The three correlations below `0.99` are `betabab_1260d` (`0.968`), `corr_1260d` (`0.983`), and `qmj_safety` (`0.986`).
- For US factors, the largest full-sample differences are concentrated in the early CRSP period. From 1952 onward, the affected examples `ami_126d`, `corr_1260d`, `ivol_capm_21d`, `ivol_capm_252d`, `iskew_capm_21d`, and `rmax5_rvol_21d` all have correlations above `0.995`.

For daily `hml_daily.parquet`:

- The rebuilt file contains 56,271,410 rows versus 56,310,154 previously, with 56,269,901 matched rows.
- Across the 153 value-weighted factor series, the pooled country-day correlation has a median above `0.99999999` and a minimum of `0.977`; 138 factors exceed `0.999` and 147 exceed `0.99`.
- Differences larger than floating-point noise (`|Δ ret_vw| > 10^-8`) occur in approximately 1.56 million matched rows, or 2.8%; only 16,164 rows differ by more than one percentage point.
- Full-sample US daily correlations are lower for some characteristics because extreme Depression-era returns amplify portfolio-composition differences—for example, `ami_126d` has a correlation of `0.552`. From 1952 onward, every factor exceeds `0.995`.

The largest changes therefore occur where expected: long-window daily characteristics (`corr_1260d` and `betabab_1260d`), characteristics derived from them (`qmj_safety` and `qmj`), sparse markets, and the early US sample. Post-1952 US factor returns remain nearly unchanged, indicating that the coverage fixes restore observations without materially altering modern-sample factor behavior.

## Breaking Changes

Previously computed characteristic and factor outputs will change when the pipeline is rebuilt. The changes are intentional bug fixes that restore valid observations and align the NASDAQ volume-adjustment boundary with its documented schedule. `CHANGELOG_DATA.md` has been updated.

## Tests

Added `tests/unit/test_daily_coverage_fixes.py` covering:

- source-conditional zero-return screening;
- dropping of stock-days with missing `mktrf` in `prepare_daily`; and
- the Gao–Ritter 2003-12-31 boundary.

Updated existing daily preparation, CRSP volume-boundary, download-configuration, and auxiliary-function tests. The targeted suite passes, and the full pipeline completed successfully. Replication notebooks confirm the rebuilt coverage against Tables 1 and 2 of the [jkp_coverage_note_v2.pdf](https://github.com/user-attachments/files/30064752/jkp_coverage_note_v2.pdf).

## Checklist

- [x] I have run the relevant tests locally (`pytest`)
- [x] I have run the linter locally (`ruff check`)


theisij:
@miahdy can you please resolve the merge conflicts

miahdy:
> @miahdy can you please resolve the merge conflicts

Done!

FernandoRDLL:
Hello, I will review this tomorrow morning

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

FernandoRDLL:
@claude  please run /code-review on this and make suggestions

miahdy:
Hi! I checked this directly: the FF fallback adds 0 dates and 0 observations in the current build. Our USA mktrf series is already complete from 1926–2025, so FF is only a safety net and isn’t responsible for the coverage increase. It is just one of the things that was mentioned in the notes Professor David Schreindorfer sent over. We can drop the FF download and fallback if we'd rather keep the change minimal. The coverage results should be unaffected either way.

________________________________

FernandoRDLL:
Thanks @miahdy. I ran this against a full build on the cluster to answer my question above about how much coverage the FF fallback adds. Findings first, then the changes I'd like.

## What the numbers say

Decomposing the three changes (US, stock-months clearing the 15-day window minimum; denominator is US stock-months with ≥15 days present in `world_dsf`, so the levels differ from your table but the attribution doesn't):

```
                                  post-1970          1970s
main (both filters)             0.6722            0.6031
+ zero_obs gate only            0.7252  (+0.053)  0.9569  (+0.354)
+ drop mktrf filter only        0.6722  (-0.000)  0.6031  (+0.000)
both fixes (this PR)            0.7252  (+0.053)  0.9569  (+0.354)
```

The `zero_obs` gate accounts for 100% of the coverage gain. The FF fallback recovers **0 dates and 0 observations**, and structurally it can't recover any: the only US dates without a `mktrf` are 163 market closures — 1984-01-02 (New Year's), 1985-09-27 (Hurricane Gloria), 2001-09-12, 2012-10-29 (Sandy), plus Thanksgivings, MLK Day, and the Reagan/Bush funeral closures. FF's daily factors are CRSP-derived, so they have no observation on those days either; I checked all 163 and FF fills 0 of them. The gaps have nothing to recover anyway — 1984-01-02, the largest at 27,404 stock-days, is 100% Compustat rows with `ret_exc` non-null = 0, so `base_data_filter_exp` drops them regardless of `mktrf`. The 25% coverage gate in `market_returns` was doing its job.

## Changes to make

**1. Remove the Ken French daily fallback.**
- Delete `coalesce_usa_mktrf_with_ff_daily` and its call in `prepare_daily`.
- Remove `ff.factors_daily` from `download_raw_data_tables` — both the `table_names` list and the date-filter dict.
- Remove the FF-daily block from `gen_raw_data_dfs`.
- Delete `test_ff_daily_fills_usa_mktrf_gap`, and revert the `ff.factors_daily` additions in `test_download_and_config.py`.

This also disposes of four review points, so please do it before the others: the unconditional download that would hard-fail the raw-data step while both consumers `.exists()`-guard it as optional (a failed download would silently yield the old low-coverage data); the untested `"Mkt-RF"`/`"Date"` fallbacks and dual candidate paths defending a fixed-schema WRDS table; the `.unique(subset=["date"])` default `keep="any"`; and the docstring saying "Left-join" / "same schema" when it's `how="full"` and grows rows (its only measurable effect is adding 102 USA rows dated 2026-01-02 to 2026-05-29, past the 2025-12-31 sample end, into `mkt_lead_lag`).

**2. Restore the `mktrf` filter in `prepare_daily`.**
- Put `.filter(col("mktrf").is_not_null())` back.
- Drop the `market_stats` branch in `base_data_filter_exp` — it exists only to compensate for that deleted line.
- Keep `source_crsp` in the `data.select(...)` list and keep `zero_obs_gate_ok()`.

The change is conceptually right, but it's world-wide rather than US (+0.0000 for the US, 1,841 stock-days) and can't be evaluated while bundled with the `zero_obs` gate. It moves to its own PR — I've opened #252 with the full numbers and the `zero_obs` coupling to handle there.

**3. Keep the `zero_obs` gate and the Gao–Ritter boundary.** With 1 and 2 done, that's the PR: one real coverage fix and one boundary correction.

**4. Smaller changes:**
- Drop `"mktrf_vol"` from the `market_stats` set — no caller passes it, only `"mktvol"` appears in `ROLLING_DAILY_SPECS`.
- Add the source-conditional `zero_obs` screen to `documentation.tex`'s dated changelog table (around line 126). Right now the docs cover Gao–Ritter but not the change that actually moves the data.
- Redate the `CHANGELOG_DATA.md` entry to the merge date rather than 13-07-2026.
- Add the Gao–Ritter one-day deviation to that changelog entry explicitly: `<=` is correct (the schedule runs /1.6 through year-end 2003), but it differs from the original SAS `<`, so published NASDAQ volume shifts on 2003-12-31.
- Link the coverage note behind the 0.94 / 0.91 targets in the PR description so reviewers can check the rebuild against it.

miahdy:
> Thanks @miahdy. I ran this against a full build on the cluster to answer my question above about how much coverage the FF fallback adds. Findings first, then the changes I'd like.
> 
> ## What the numbers say
> Decomposing the three changes (US, stock-months clearing the 15-day window minimum; denominator is US stock-months with ≥15 days present in `world_dsf`, so the levels differ from your table but the attribution doesn't):
> 
> ```
>                                   post-1970          1970s
> main (both filters)             0.6722            0.6031
> + zero_obs gate only            0.7252  (+0.053)  0.9569  (+0.354)
> + drop mktrf filter only        0.6722  (-0.000)  0.6031  (+0.000)
> both fixes (this PR)            0.7252  (+0.053)  0.9569  (+0.354)
> ```
> 
> The `zero_obs` gate accounts for 100% of the coverage gain. The FF fallback recovers **0 dates and 0 observations**, and structurally it can't recover any: the only US dates without a `mktrf` are 163 market closures — 1984-01-02 (New Year's), 1985-09-27 (Hurricane Gloria), 2001-09-12, 2012-10-29 (Sandy), plus Thanksgivings, MLK Day, and the Reagan/Bush funeral closures. FF's daily factors are CRSP-derived, so they have no observation on those days either; I checked all 163 and FF fills 0 of them. The gaps have nothing to recover anyway — 1984-01-02, the largest at 27,404 stock-days, is 100% Compustat rows with `ret_exc` non-null = 0, so `base_data_filter_exp` drops them regardless of `mktrf`. The 25% coverage gate in `market_returns` was doing its job.
> 
> ## Changes to make
> **1. Remove the Ken French daily fallback.**
> 
> * Delete `coalesce_usa_mktrf_with_ff_daily` and its call in `prepare_daily`.
> * Remove `ff.factors_daily` from `download_raw_data_tables` — both the `table_names` list and the date-filter dict.
> * Remove the FF-daily block from `gen_raw_data_dfs`.
> * Delete `test_ff_daily_fills_usa_mktrf_gap`, and revert the `ff.factors_daily` additions in `test_download_and_config.py`.
> 
> This also disposes of four review points, so please do it before the others: the unconditional download that would hard-fail the raw-data step while both consumers `.exists()`-guard it as optional (a failed download would silently yield the old low-coverage data); the untested `"Mkt-RF"`/`"Date"` fallbacks and dual candidate paths defending a fixed-schema WRDS table; the `.unique(subset=["date"])` default `keep="any"`; and the docstring saying "Left-join" / "same schema" when it's `how="full"` and grows rows (its only measurable effect is adding 102 USA rows dated 2026-01-02 to 2026-05-29, past the 2025-12-31 sample end, into `mkt_lead_lag`).
> 
> **2. Restore the `mktrf` filter in `prepare_daily`.**
> 
> * Put `.filter(col("mktrf").is_not_null())` back.
> * Drop the `market_stats` branch in `base_data_filter_exp` — it exists only to compensate for that deleted line.
> * Keep `source_crsp` in the `data.select(...)` list and keep `zero_obs_gate_ok()`.
> 
> The change is conceptually right, but it's world-wide rather than US (+0.0000 for the US, 1,841 stock-days) and can't be evaluated while bundled with the `zero_obs` gate. It moves to its own PR — I've opened #252 with the full numbers and the `zero_obs` coupling to handle there.
> 
> **3. Keep the `zero_obs` gate and the Gao–Ritter boundary.** With 1 and 2 done, that's the PR: one real coverage fix and one boundary correction.
> 
> **4. Smaller changes:**
> 
> * Drop `"mktrf_vol"` from the `market_stats` set — no caller passes it, only `"mktvol"` appears in `ROLLING_DAILY_SPECS`.
> * Add the source-conditional `zero_obs` screen to `documentation.tex`'s dated changelog table (around line 126). Right now the docs cover Gao–Ritter but not the change that actually moves the data.
> * Redate the `CHANGELOG_DATA.md` entry to the merge date rather than 13-07-2026.
> * Add the Gao–Ritter one-day deviation to that changelog entry explicitly: `<=` is correct (the schedule runs /1.6 through year-end 2003), but it differs from the original SAS `<`, so published NASDAQ volume shifts on 2003-12-31.
> * Link the coverage note behind the 0.94 / 0.91 targets in the PR description so reviewers can check the rebuild against it.

Hi @FernandoRDLL , I’ve addressed the comments mentioned. Please let me know if I missed anything. Thanks!

FernandoRDLL:
Hey @miahdy 

466472e ("Address followup comments…") rewrote aux_functions.py wholesale and dropped the Bessembinder Compustat corrections from #234 along the way — looks like an older copy of the file got pasted over the merged one. The merge at f5c7266 was clean; Bessembinder was still wired in there.

compustat_correction.py itself is untouched (byte-identical to main), but nothing calls it anymore. What went missing is the wiring:

- from .compustat_correction import correct_decimal_errors, drop_unreliable_observations — gone, module now orphaned
- gen_comp_dsf: apply_correction / correction_method / variation_threshold params and the correction body — gone
- gen_returns_df: the ret > 10 post-correction screen — gone
- the trfd never-dividend COALESCE substitution (inline SQL) — gone
- gen_comp_dsf views → back to 5 materialized DuckDB tables + the old NASDAQ UPDATE
- CHANGELOG_DATA.md: the 10-07-2026 entry was replaced rather than appended
- tests/unit/test_deterministic_dedup.py: lost 37 lines
- process_window: the "mktrf_vol" dispatch alias was dropped ("mktvol" remains, so the function is still reachable — probably inert, just collateral)

Worth flagging: CI is green because test_compustat_correction.py tests the module directly and the module is fine. Nothing asserts the pipeline actually calls it.

Don't rebase — it replays 466472e and the clobber comes with it. git merge origin/main alone won't help either: main hasn't touched these regions since the merge-base, so it auto-resolves and keeps the clobbered version.

Restore from main, then re-apply your change on top:

git switch 235-data-us-daily-firm-characteristics-data-under-covered
git merge origin/main
git checkout origin/main -- src/jkp/data/aux_functions.py \
                            CHANGELOG_DATA.md \
                            tests/unit/test_deterministic_dedup.py
git commit -m "Restore Compustat correction clobbered in 466472e; re-apply coverage fix"

The re-apply is ~40 lines: zero_obs_gate_ok() plus its two call sites (prepare_daily, base_data_filter_exp), Gao–Ritter < → <= in adj_trd_vol_NASDAQ and gen_secm_data, carrying source_crsp through the select, and the mkt_lead_lag null filter. test_daily_coverage_fixes.py stays as-is. Append a new CHANGELOG_DATA.md entry instead of replacing 10-07-2026.

The merge also picks up main's pgpass helpers (_pg_escape_value, _sql_literal, _redact_password, _password_forms, _password_in_error) — the branch just predates them, nothing was lost there.

FernandoRDLL (commented):
One fix before merge — the rest looks great.

src/jkp/data/aux_functions.py
```diff
@@ -7534,6 +7534,21 @@ def residual_momentum(paths: DataPaths, output_path, data_path, fcts_path, __n,
...
+        1) CRSP-sourced rows (source_crsp == 1) always pass.
+        2) Compustat rows pass only when zero_obs < 10.
+
+    Output:
+        Boolean Polars expression usable in .filter().
+    """
+    return (pl.col("source_crsp") == 1) | (pl.col("zero_obs") < 10)
+
+
 def prepare_daily(paths: DataPaths, data_path, fcts_path):
```
FernandoRDLL: The inserted `zero_obs_gate_ok` stole `prepare_daily`'s `@measure_time`, leaving `prepare_daily` (a real pipeline fn) undecorated and putting the decorator on a pure expr-builder that's called inside `.filter()` — timing noise every call, against the `@measure_time`-for-pipeline-fns-only convention. Move it back:

```suggestion
def zero_obs_gate_ok() -> pl.Expr:
    """
    Description:
        Source-conditional quality screen for months with many zero local-return days.

    Steps:
        1) CRSP-sourced rows (source_crsp == 1) always pass.
        2) Compustat rows pass only when zero_obs < 10.

    Output:
        Boolean Polars expression usable in .filter().
    """
    return (pl.col("source_crsp") == 1) | (pl.col("zero_obs") < 10)


@measure_time
def prepare_daily(paths: DataPaths, data_path, fcts_path):
```

tests/unit/test_deterministic_dedup.py
```diff
@@ -190,40 +190,3 @@ def test_no_duplicates_unchanged(self, test_paths) -> None:
 
         feb_ret = result.filter(pl.col("datadate") == date(2024, 2, 29))["ret"][0]
         assert feb_ret == pytest.approx(0.1, rel=1e-10)
```
FernandoRDLL, line 192: This deletes the only test guarding the `> 1000%` (`ret > 10`) nulling in `gen_returns_df`, which is still live code (`aux_functions.py` ~line 2456). Without it, a future regression weakening that clip would let implausible Compustat returns leak into `data/processed/`. The clip is unrelated to this PR's scope, so the removal looks accidental — restore it:

```suggestion
        assert feb_ret == pytest.approx(0.1, rel=1e-10)

    def test_nulls_return_above_10x(self, test_paths) -> None:
        """Post-correction returns > 1000% (ret > 10) are nulled; a return at
        the 10x bound is kept."""
        df = pl.DataFrame(
            {
                "gvkey": ["001"] * 4,
                "iid": ["01"] * 4,
                "datadate": [
                    date(2024, 1, 31),
                    date(2024, 2, 29),  # ret = 9.0  (<= 10, kept)
                    date(2024, 3, 29),  # ret = 11.0 (> 10, nulled)
                    date(2024, 4, 30),  # ret = 0.0  (kept)
                ],
                "eom": [
                    date(2024, 1, 31),
                    date(2024, 2, 29),
                    date(2024, 3, 29),
                    date(2024, 4, 30),
                ],
                "prcstd": [4, 4, 4, 4],
                "ri": [100.0, 1000.0, 12000.0, 12000.0],
                "ri_local": [100.0, 1000.0, 12000.0, 12000.0],
                "curcdd": ["USD", "USD", "USD", "USD"],
            }
        )
        df.write_parquet(test_paths.interim_dir / "__comp_msf.parquet")

        from jkp.data.aux_functions import gen_returns_df

        result = gen_returns_df(test_paths, "m")

        feb = result.filter(pl.col("datadate") == date(2024, 2, 29))
        mar = result.filter(pl.col("datadate") == date(2024, 3, 29))
        assert feb["ret"][0] == pytest.approx(9.0, rel=1e-10)  # at bound, kept
        assert mar["ret"][0] is None  # > 10x, nulled
        assert mar["ret_local"][0] is None
```

tests/unit/test_daily_coverage_fixes.py
```diff
@@ -0,0 +1,180 @@
...
+                "roe": 0.0,
+                "smb_hxz": 0.0,
+            }
+            for d in dates
+        ]
+        dsf_path, fcts_path = self._write_inputs(test_paths, rows, fcts_rows)
+        prepare_daily(test_paths, dsf_path, fcts_path)
+        corr = pl.read_parquet(test_paths.interim_dir / "corr_data.parquet")
+        # 10 days → 8 non-null 3-day sums for a contiguous series.
+        assert corr.height == 8
```
FernandoRDLL: `TestPrepareDailyCoverage` duplicates two tests added to `test_rolling_daily_metrics.py` in this same PR: `test_drops_stock_day_when_mktrf_null` ≈ `test_prepare_daily_drops_null_mktrf_stock_days`, and `test_crsp_zero_month_kept_in_corr_data` ≈ `test_corr_data_keeps_crsp_high_zero_obs`. Two copies of the same `prepare_daily` contract drift over time. `test_rolling_daily_metrics.py` is the established home (`TestCorrData` already lives there), so drop this class here:

```suggestion
        assert df["vol"][0] == pytest.approx(1600.0)
```

FernandoRDLL (approved):

FernandoRDLL:
@miahdy Just merged! Great stuff

PR #251, opened 2026-07-15
fix: mask the WRDS password on the default postgres_scan download path

### LINKED ISSUE #250: Mask the WRDS password on the default `postgres_scan` download path
## Summary
`get_columns` and `download_wrds_table` in `src/jkp/data/aux_functions.py` run their queries with the full libpq conninfo embedded in the SQL but have no password-masking wrapper. On a wrong or expired WRDS password, the DuckDB error echoes the conninfo, so the password lands in the terminal and in Slurm logs. This is the `postgres_scan('{conninfo}', ...)` branch, which is what `jkp build` runs with default flags (a single worker, no `--persistent-connection`), so it is the most common path.

## Reproduction
Confirmed against real WRDS on an HPC login node. Calling the unmasked `get_columns` with a deliberately wrong password fails with a DuckDB `IOException` whose text echoes the full conninfo, so the password appears in the error and in any logs that capture it.

## Context
This is pre-existing behavior, not a regression. It was found during the PR #245 review and deliberately deferred out of that PR to keep its scope to credential storage (keyring to `~/.pgpass`). PR #245 added the masking helpers that make the fix straightforward.

The sibling paths are already masked: `_attach_wrds` (the `--persistent-connection` branch) suppresses and redacts the password, and the parallel download worker redacts it via `_redact_password`. The default sequential `postgres_scan` path is the one gap left. A related instance is the `_compute_histograms` / `_year_histogram` error path, which has the same gap and should be covered in the same change.

## Impact
Wrong or expired password plus default flags puts the WRDS password into plaintext terminal output and Slurm log files. Low likelihood, since it only happens on an auth failure, but a plaintext credential in logs is worth closing before the next release.

## Fix
Wrap the two `postgres_scan` executions the way `_attach_wrds` does, reusing the helpers PR #245 added to `aux_functions.py`:
- `_password_in_error(str(e), password)` to detect whether the error carries the secret.
- `_redact_password(str(e), password)` to scrub the password from the error.

`get_columns` and `download_wrds_table` do not currently receive `password`, so this means threading a `password` parameter through those two signatures and their call sites. Apply the same wrapper to the `_compute_histograms` / `_year_histogram` error path.

This approach was validated against real WRDS: wrapping the failing call in the `_attach_wrds`-style guard, or redacting the error with `_redact_password`, removes the password from the surfaced error.

## Acceptance
- A failed default-flag download with a wrong password surfaces a generic error with no password in the message or logs.
- A unit test reproducing the leak (a synthesized DuckDB-style error containing the password) fails before the fix and passes after.

## References
- Deferred item from the PR #245 review.
- Helpers to reuse: `_password_in_error` and `_redact_password` in `src/jkp/data/aux_functions.py`.

capellini:
## Summary

Fixes #250. `get_columns` and `download_wrds_table` ran `postgres_scan('{conninfo}', ...)` with the full libpq conninfo (including the password) embedded in the SQL, but had no password-masking wrapper. On a wrong or expired WRDS password, DuckDB's `IOException` echoes the conninfo, so the password landed in the terminal and in Slurm logs. This is the `postgres_scan` branch that `jkp build` runs by default (single worker, no `--persistent-connection`), so it is the most common path. The sibling paths (`_attach_wrds`, the parallel worker) were already masked; this closes the remaining gap.

The fix adds a shared password-leak guard used at every conninfo-bearing call site:

- **`_guard_password_errors`** wraps the risky call. On an error whose text carries the password it re-raises a password-free `RuntimeError`; other errors propagate unchanged. It raises *outside* the `except` block (so the original isn't chained on via `__context__`), drops the raw error text from the raise-site frame's locals, and closes two exotic exits (a `MemoryError`/`KeyboardInterrupt` inside the transform, and a `Ctrl-C` in the detection window).
- **`_mask_password_errors`** (generic credential/network message) for connection-establishing calls — the first `postgres_scan` probe and `_attach_wrds`.
- **`_redact_password_errors`** (keeps the real diagnostic minus the secret) for post-connection failures — the COPY, the persistent per-table loop, and the histogram query.
- **Fragment-tolerant detection**: DuckDB's caret-centered `LINE 1: ...` statement excerpt can slice a password mid-string, leaving a tail that no whole-form check would catch. Detection now matches truncated fragments, and the redact path **fails closed** to the generic message when only a fragment matched (it can't be reliably scrubbed).
- `password` is now **keyword-only and required** on `get_columns`/`download_wrds_table`, so a caller can't silently omit it and re-open the leak. `None` (the `~/.pgpass` path, where the conninfo has no secret) is still valid and disables masking.

Validated against real WRDS on an HPC login node and against real DuckDB 1.5.2: the leak reproduces, the guards detect/scrub the real error format across all failure modes, the happy path is unaffected, and the `.pgpass` path (password never in the conninfo) is confirmed to carry no secret.

## Change Type

## Methodological Impact

None.

## Data Impact

None. No change to any values, columns, or methodology written to `data/processed/`.

## Breaking Changes

None to released datasets. Internal API: `get_columns` and `download_wrds_table` now take a keyword-only, required `password` argument (both are internal helpers with no external callers).

## Tests

Added a security-focused suite (`TestPostgresScanPasswordMasking` plus additions to `TestAttachWrds`, `TestGenWrdsConnectionInfo`, and `TestDownloadRawDataTablesBranching`): synthesized-leak masking/redaction, the required-keyword `TypeError`, exception-chain and frame-locals scrubbing, the transform-failure and Ctrl-C exotic exits, fragment detection + fail-closed redaction on a truncated echo, the per-table context prefix, and orchestrator-level end-to-end masking for both the sequential and persistent paths. The original synthesized-leak test fails before the fix and passes after. Full unit suite: 912 passed; ruff clean; pyright unchanged.

## Checklist

## Additional Notes

Pre-existing behavior, not a regression — deferred out of #245 to keep that PR scoped to credential storage. No `CHANGELOG_DATA.md` entry (no data-output change).

### DIFF
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -702,6 +702,16 @@ def gen_wrds_connection_info(user, password: str | None = None) -> str:
 
     When ``password`` is ``None`` the ``password=`` field is omitted, so libpq
     authenticates from ``$PGPASSFILE`` / ``~/.pgpass`` instead.
+
+    Password-masking note: DuckDB's *statement* echo (``LINE 1: ...``) is a caret-centered window
+    with front elision (``...``), not a head-anchored cut, so field ordering does NOT make it safe --
+    a left cut whose window starts inside the password leaves only a tail fragment, and ``password=``
+    coming last actually maximizes that tail exposure. This isn't defended here by ordering; it's
+    defended downstream: connection-stage errors echo the full conninfo untruncated (detection fires
+    and masks), and the guards use fragment-tolerant detection that fails closed on a truncated echo
+    (see :func:`_password_leak_in_error`, which also notes the one gap -- a truncated echo that
+    retains fewer than 8 consecutive form characters). A truncated statement echo is also unreachable
+    through the current call sites, whose SQL templates are fixed and conninfo is ``_sql_literal``d.
     """
     parts = [
         f"host={WRDS_HOST}",
@@ -725,9 +735,10 @@ def gen_wrds_connection_info(user, password: str | None = None) -> str:
 def _password_forms(password: str) -> tuple[str, str, str]:
     """The forms the password can take on its way into an error message: raw, the
     libpq-escaped conninfo form (echoed by a connection IOException), and the
-    SQL-escaped-then-libpq-escaped form (echoed from the raw statement text by a
-    parser error). Ordered most-escaped first so redaction replaces the longest
-    match before its shorter substrings."""
+    libpq-escaped form with SQL-escaping applied on top (echoed from the raw statement
+    text by a parser error -- ``_sql_literal(_pg_escape_value(pw))``, libpq first then SQL).
+    Ordered most-escaped first so redaction replaces the longest match before its shorter
+    substrings."""
     escaped = _pg_escape_value(password)
     return (_sql_literal(escaped), escaped, password)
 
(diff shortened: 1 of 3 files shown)


capellini:
I'm going to hold off on taking this out of draft until I merge https://github.com/bkelly-lab/jkp-data/pull/257.  @FernandoRDLL has made some wonderful code modularization changes in that PR that necessitate refactoring my code once merged.

PR #253, opened 2026-07-15
Add overnight and intraday return columns

### LINKED ISSUE #181: [Feature]: Overnight and intraday return
### Feature Description

We should add overnight (close-to-open) and intrday (open-to-close) return columns. 

### Use Case / Motivation

To increase the highest return frequency from daily to intraday/overnight

[Lou, Polk, and Skouras (2019)](https://personal.lse.ac.uk/polk/research/TugOfWar.pdf) is a really nice paper showing how some factors outperform overnight whereas others outperform intraday. Once we have everything implemented, we should try to replicate there paper across all JKP factors

We need open prices to make this decomposition. I know this is available in CRSP, but not sure if it's available in Compustat. If it's CRSP only, that's fine, but adding Compustat is preferable if possible. 

So here're the things we need to do:
1. Add the open price from CRSP and Compustat (if possible) to `prepare_comp_sf()` and `prepare_crsp_sf()`
2. Add the overnight and intraday return columns, following the implementation in Lou, Polk, and Skouras (2019) to `prepare_comp_sf()` and `prepare_crsp_sf()`
3. Add the overnight and intraday return of each factor as separate outputs to `portfolio.py`
4. Add the intraday and overnight returns for each factor to [JKPfactors.com](https://jkpfactors.com/) 

### Willingness to Contribute

No, I am just suggesting the idea

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

miahdy:
# Add open prices and LPS overnight/intraday returns to CRSP/Compustat security files

## Summary

This PR adds open prices and Lou, Polk, and Skouras (2019) overnight/intraday return decomposition to the CRSP and Compustat security-file preparation steps. The goal is to make stock-level overnight/intraday returns available in both USD (`ret_intraday`, `ret_overnight`) and local currency (`ret_intraday_local`, `ret_overnight_local`) in the unified world panel for downstream analysis (including a follow-up PR for factor-level decomposition in `portfolio.py`).

### Open prices

- **CRSP (`gen_crsp_sf` → `prepare_crsp_sf`)**: For daily data, reads `dlyopen` and `dlyclose` from CRSP DSF v2 and writes `prc_open` and `prc_close` (positive prices only; null otherwise). Monthly files carry null placeholders.
- **Compustat (`gen_comp_dsf` → `prepare_comp_sf`)**: For daily data, reads `prcod` from Compustat SECD (NA and Global) when present and positive, scales to local units, converts to USD via FX, and writes `prc_open`. Open price is omitted from monthly aggregation (not available at monthly frequency).

### Overnight/intraday returns (LPS 2019)

Following Lou, Polk, and Skouras (2019), daily close-to-close returns are decomposed in both USD and local currency:

- `ret_intraday = P_close / P_open − 1`
- `ret_overnight = (1 + ret) / (1 + ret_intraday) − 1`
- `ret_intraday_local = ret_intraday` (FX cancels in the open/close ratio)
- `ret_overnight_local = (1 + ret_local) / (1 + ret_intraday_local) − 1`

so that `(1 + ret_intraday)(1 + ret_overnight) = (1 + ret)` and `(1 + ret_intraday_local)(1 + ret_overnight_local) = (1 + ret_local)`. The overnight residual against `ret_local` instead of `ret` keeps daily FX moves out of `ret_overnight_local`. For CRSP, `ret_local` equals `ret`, so the local and USD overnight components coincide.

Implementation details:

- **`prepare_crsp_sf` (daily)**: Uses `prc_close` (actual closing trade price from `dlyclose`) rather than `prc` (`dlyprc`, which may be a bid–ask midpoint) for the intraday leg, to align with LPS.
- **`process_comp_sf1` / `prepare_comp_sf` (daily)**: Uses `prc` (close) and `prc_open` when the open column is present; otherwise writes null component returns.
### Downstream propagation

- **`combine_crsp_comp_sf`**: Passes `ret_intraday`, `ret_overnight`, `ret_intraday_local`, and `ret_overnight_local` through the CRSP/Compustat union into `world_dsf.parquet`.
- **Processed outputs**: `save_daily_ret` includes all four overnight/intraday columns. Monthly outputs do not contain overnight/intraday columns.

### Out of scope for this PR

- Factor portfolio construction using overnight/intraday returns (`portfolio.py`) — covered in a separate PR.

## Methodology reference

Lou, Polk, and Skouras (2019), *A Tug of War: Overnight Versus Intraday Expected Returns*, Journal of Financial Economics.

## Test plan

- [x] `uv run pytest tests/unit/test_overnight_intraday.py` — LPS formulas, return identity, null handling
- [x] `uv run pytest tests/unit/test_combine_crsp_comp_sf.py` — daily columns flow through CRSP/Compustat union
- [x] `uv run ruff check src/jkp/data/aux_functions.py src/jkp/data/main.py`
- [x] `uv run ruff format --check src/jkp/data/ tests/`
- [x] Spot-check daily `world_dsf.parquet`: non-null `ret_intraday`/`ret_overnight` where `prc_open` > 0; identity holds row-wise


miahdy:
Hi @theisij @FernandoRDLL , could you please review this PR when you have a chance? Thanks!

tests/unit/test_overnight_intraday.py
```diff
@@ -0,0 +1,397 @@
...
+        assert result["ret_intraday"][0] is None
+        assert result["ret_overnight"][0] is None
+
+
+# ---------------------------------------------------------------------------
+# Monthly compounding tests
+# ---------------------------------------------------------------------------
+
+
+class TestMonthlyCompounding:
```
theisij, line 159: The tests in this class seems to just re-implement the compounding logic, without testing the actual new functions, so they can't cath bugs in the actual; code. For example, test_compound_with_nulls copies the pl.when(col(...).is_null().any()) expression verbatim from compound_overnight_intraday. If the real function had a typo, the test wouldn't catch it. As a result, compound_overnight_intraday() has no test covereage at all. 

So please replace this class with a test that call the real function. Maybe a small data set where you call the function on while knowing the correct output, with a couple of curveball, like one month with a null component

src/jkp/data/aux_functions.py
```diff
@@ -2582,6 +2611,18 @@ def process_comp_sf1(paths: DataPaths, freq):
...
     __comp_sf2 = gen_temporary_sf(paths, freq, __returns, __delist)
     __comp_sf2 = add_rf_and_exchange_data_to_temporary_sf(paths, freq, __comp_sf2)
+
+    if freq == "d" and "prc_open" in __comp_sf2.columns:
+        __comp_sf2 = __comp_sf2.with_columns(
+            ret_intraday=pl.when((col("prc_open") > 0) & (col("prc") > 0))
+            .then(col("prc") / col("prc_open") - 1)
+            .otherwise(fl_none()),
+        ).with_columns(
+            ret_overnight=pl.when(col("ret_intraday").is_not_null() & col("ret").is_not_null())
```
theisij: One thing we need to handle deliberately here is the FX convention. ret_intraday is computed as prc / prc_open, and both prices are converted with the same day's FX rate, so FX cancels out and the intraday component is effectively a local-currency return. ret_overnight is then the residual against ret, which is in USD. The net effect is that the entire daily FX move (plus dividends) ends up in ret_overnight. For example, a Japanese stock that is flat in JPY all day, on a day the yen drops 2% vs USD, gets ret_intraday = 0 and ret_overnight = -2%. 

This convention is partly defensible, since the local session is only 6–8 of 24 hours and the overnight leg already holds dividends as the residual. But it is a choice that we need to be upfront about. It also makes international overnight returns hard to compare with the CRSP ones, which are USD to begin with.

Two things:

1. Add local versions alongside the USD ones. Compute ret_intraday_local and ret_overnight_local against ret_local, and keep the current USD ret_intraday / ret_overnight. Note that ret_intraday and ret_intraday_local coincide under this convention, but I'd include both anyway so each pair is complete and usable off the shelf. Please also propagate all four columns through the monthly compounding, the _lead1m columns, and save_daily_ret / save_monthly_ret.

2. State the convention explicitly in the docstrings, the documentation, the CHANGELOG_DATA.md entry, and the release notes: intraday is the local open-to-close price return, the USD overnight component contains the full daily FX move and dividends, and the identities (1+ret_intraday_local)(1+ret_overnight_local) = 1+ret_local and (1+ret_intraday)(1+ret_overnight) = 1+ret hold row-wise.

theisij (commented):
Hi Mia, 

This looks really good. Thank you. 

A couple of things:
1. Please make an entry to CHANGELOG_DATA.md, which explains what you added here.
2. Please update the documentation as well.
3. I feel like a lot of the tests don't really call the function they are trying to test. See my specific comment in the code. 
4. One thing that's problematic in our current interpretation is the way we handle FX. Best I can tell, it seems like the FX effect is entirely absorbed by ret_overnight. That's to some extent an okay convention, as the intraday period is maybe 6-8 hours of the overnight period. But it is something that we should flag. In addition to having the usd versions of ret_overnight and ret_intraday, it would be good if we could also add the two local versions (although ret_intraday and ret_intraday_local will coincide---better to include both). We also need to state this problem clearly in the documentation, changelog, and the release notes. See my comments in the code

miahdy:
> Hi Mia,
> 
> This looks really good. Thank you.
> 
> A couple of things:
> 
> 1. Please make an entry to CHANGELOG_DATA.md, which explains what you added here.
> 2. Please update the documentation as well.
> 3. I feel like a lot of the tests don't really call the function they are trying to test. See my specific comment in the code.
> 4. One thing that's problematic in our current interpretation is the way we handle FX. Best I can tell, it seems like the FX effect is entirely absorbed by ret_overnight. That's to some extent an okay convention, as the intraday period is maybe 6-8 hours of the overnight period. But it is something that we should flag. In addition to having the usd versions of ret_overnight and ret_intraday, it would be good if we could also add the two local versions (although ret_intraday and ret_intraday_local will coincide---better to include both). We also need to state this problem clearly in the documentation, changelog, and the release notes. See my comments in the code

Thank you for the detailed feedback! I’ll wait for Fernando’s review as well and then address all of the comments together.

capellini:
> The tests in this class seems to just re-implement the compounding logic, without testing the actual new functions

I'm leaving this here as a side note for the group. AI is notorious for these sorts of testing practices. They often "test the mock" where they mock behavior and then test their mocked behavior, instead of testing the underlying behavior (tautological tests). To combat this, I plan to add mutation testing to this repository.

Mutation testing mutates the source code in an attempt to identify tests that will always pass. This must be approached carefully because mutation testing is known to produce false positives, often consumes a large amount of memory, and can take some time to execute. But I still think it would be a valuable addition as an added safety check  for our code base.

theisij:
> > The tests in this class seems to just re-implement the compounding logic, without testing the actual new functions
> 
> I'm leaving this here as a side note for the group. AI is notorious for these sorts of testing practices. They often "test the mock" where they mock behavior and then test their mocked behavior, instead of testing the underlying behavior (tautological tests). To combat this, I plan to add mutation testing to this repository.
> 
> Mutation testing mutates the source code in an attempt to identify tests that will always pass. This must be approached carefully because mutation testing is known to produce false positives, often consumes a large amount of memory, and can take some time to execute. But I still think it would be a valuable addition as an added safety check for our code base.

Okay this is very good to know. @miahdy, @FernandoRDLL please watch out for this. It leads to unnecesary bloat in the code base, and should be avoided like the plague.

theisij (approved):
This looks great, Mia!

Issue #41, opened 2026-03-11
[Feature]: Implement end-to-end test of a tiny test data set


theisij:
## Overview
Following our discussion in #25, I thought we should get Fernando started on writing an end-to-end test on a tiny subset of the data. Our goal should be to verify that changes made to the codebase only affects key files like world_dsf, world_msf, and world_data in the expected way. For example, if we've added a new feature to world_data, then we should see one new column here and no other changes, plus no other changes to world_dsf and world_msf. 

## Step 1: creating the test sample
The code base is massive, so running an end-to-end test on all data is impractical. (although we may want to do that in the future). Further, the output could change because the input has changed (for example, Compustat data is frequently revised), and that's a separate problem. Therefore, I suggest that we start by creating a tiny test sample that we can use for the end-to-end test. 

Concretely, I think we should:

1. Pick a random subset of, say, 100 ids from the US and 100 ids from GBR. I haven't give much thought to whether we should have 10/100/1000 ids, or whether we should have 2/10/all countries represented, so feel free to play around with this. The key considerations are speed of running the test (should probably not be more than 5 minutes) and whether the codebase can actually be run on the test sample (for example, I think we have constraints on the minimum number of assets needed to create asset pricing factors, which is why I don't think we can go as low as 10 ids)
2. Create a version of all the files we usually save to `data/raw` that only have the information related to those ids, and their associated gvkeys.

Note: the script to create the test sample should be in our repo, such that we can quickly create a new one, if needed. 

## Step 2: creating the end-to-end test
We need to:

1. create a script to run the code on the test data
2. integrate this script into our test suite.
3. Debate how often it should run. I'm thinking everytime we open a pull request, and everytime we merge something to the main branch
4. Set the excecution up such that it runs at the step we decided on in 3.

## Feedback welcome
@FernandoRDLL: let me know if anything is unclear
@capellini: I have never in my life written an end-to-end test, so what do you think about the structure I'm proposing here? Anything you'd change?

### Use Case / Motivation

This will help us make sure that new additions to the code base doesn't break any of the old behavior

### Willingness to Contribute

Yes, I can help with testing/validation

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this feature has not already been requested
- [x] I have read the documentation to confirm this feature does not already exist

capellini:
I agree that a full end-to-end test should ultimately be the goal.

I think that for the immediate future, we should aim to build unit and snapshot test coverage.

[...]

PR #255, opened 2026-07-15
Add an inventory manifest guarding against silent deletions

### LINKED ISSUE #254: [Feature]: CI does not detect when a pipeline stage stops being called
## Feature Description

Nothing in CI detects when part of the repo goes missing. Our tests import helpers and test them in isolation — that proves a function works, never that `main.py` still reaches it, and never that the thing existed at all yesterday. The repo has no description of what it's *supposed* to contain, so nothing can notice when a piece disappears.

**Addressed by #255**, which checks in `tests/structure/pipeline_structure.txt`: every function and signature, import edge, pipeline stage, dispatch entry, test, tracked artifact, changelog entry, and characteristic name — derived from the AST, no fixtures, ~0.5s. Plus two assertions read from live code (so regenerating can't paper over them): no orphaned modules, and the three characteristic sources agree.

## Use Case / Motivation

No protection against a merge being silently undone. A bad conflict resolution or a stale copy pasted over a file leaves tests passing and CI green while the pipeline stops doing what we merged it to do.

Instance (footnote): in PR #247, commit `466472e` rewrote `aux_functions.py` and dropped the Bessembinder corrections from #234. All five checks passed — three separate holes:

| Dropped | Why nothing noticed |
|---|---|
| `## 10-07-2026` changelog entry | Nothing reads the changelog |
| `gen_comp_dsf`'s correction params | No test calls `gen_comp_dsf` |
| `compustat_correction` un-wired | Its own tests still pass; they never ask if the pipeline calls it |
| `test_nulls_return_above_10x` | The clobber deleted the test with the code — a test can't catch its own deletion |

Verified: #255 catches all four against #247's head.

## Scope

**In scope — existence and shape.** Something deleted, renamed, un-wired, or a signature changed.

**Out of scope — behaviour.** #255 would *not* have caught #247's trfd `COALESCE` removal: a pure inline edit that deletes nothing and renames nothing. Only an output-level test catches that class. Tracked separately in #41 / #97. This issue does not close that gap and shouldn't be read as if it does.

**Rejected: per-function body hashes.** They would catch the inline class, but every legitimate edit would fail the test, and a check that always fires trains people to regenerate without looking — which is exactly the #247 failure.

**Rejected: function-level reachability.** 17 of 300 defs unreachable, 5 genuinely dead — needs a baseline allowlist that rots. Module granularity covers #247 with zero noise.

## Follow-ups

- `_golden_helpers.py`'s `_link()` returns silently when a fixture is missing (`stage_synthetic_slices` docstring: missing files are "silently skipped"). Deleting `synthetic_wrds/comp_fundq.parquet` raises nothing. Files declared in `RAW_FILES`/`INTERIM_FILES` should raise. Latent bug, own PR.
- `documentation.pdf` is a checked-in build output with no tracked build step; #247 edits the `.tex`/`.bib` without it, leaving it stale.

## Willingness to Contribute

FernandoRDLL:
Closes #254.

## Simple explanation

PR #247 pasted an old copy of `aux_functions.py` over the current one and quietly undid work we had already merged. Every CI check passed anyway. Three different holes let it through:

- The `ret > 10` screen had a test — but the same commit deleted the test along with the code. A test can't catch its own deletion.
- `compustat_correction` stopped being called by the pipeline, yet its own unit tests still passed, because they test the module directly and never ask whether anything uses it.
- The Bessembinder changelog entry was replaced rather than appended, and nothing anywhere looks at the changelog.

The root problem: the repo never described what it's *supposed* to contain, so nothing could notice when a piece went missing.

This adds that description. `tests/structure/pipeline_structure.txt` lists every tracked file, every function and its signature, every import edge, the pipeline's stage order, the dispatch table, every test, every changelog entry, and every characteristic name. A test checks reality still matches it.

When you legitimately add or remove something, the test tells you to run one command and commit the updated file — the same contract as `EXPECTED_PUBLIC_API` in `test_public_api.py`. The manifest's own diff is then the thing a reviewer can't scroll past.

## Technical details

`tests/structure/build_manifest.py` derives everything from the AST — no imports of `jkp.data`, no fixtures, no pipeline run, 0.21s for 1,977 lines. Files come from `git ls-files` rather than filesystem globs, so untracked Finder-style `" 2.py"` duplicates don't pollute it.

Sections, one fact per line:

```
file     src/jkp/data/resources/factor_details.xlsx
fn       aux_functions.gen_comp_dsf(paths: DataPaths, apply_correction: bool = True, ...)
imp      aux_functions -> compustat_correction
stage    034 roll_apply_daily
dispatch process_window.funcs mktvol -> mktrf_vol
test     tests/unit/test_deterministic_dedup.py::TestGenReturnsDfDedup::test_nulls_return_above_10x
chlog    10-07-2026
char     portfolio ret_12_1
```

Three assertions in `tests/structure/test_manifest.py`:

1. **Manifest matches the repo**, failing with a unified diff and the regen command.
2. **No orphaned modules** — reachability from `{cli, main, portfolio, __init__}` over relative-import edges. Read from live code, not the manifest text, so regenerating can't paper over it. This is the `compustat_correction` case.
3. **Three characteristic sources agree** — `PORTFOLIO_CHARS`, `cluster_labels.csv`, and `factor_details.xlsx[abr_jkp]` (153 each). Two are opaque to review, so a vanishing row would otherwise shrink the output unnoticed. Today `test_download_and_config.py` only checks non-emptiness plus five anchor names — deleting 100 characteristics passes.

Design decisions worth knowing:

- **`file` lists every tracked file**, rather than filtering by a suffix blocklist. A blocklist left tracked `.py` outside `src/jkp/data/` and `tests/` covered by no section at all. The overlap with `fn`/`test` for Python files is deliberate: those pin contents, `file` pins existence.
- **`test` and `fn` names are fully qualified.** Names aren't unique — 34 test definitions share a name with another, two of them inside a single file — so bare names would collapse under de-duplication and hide a deletion.
- **`stage` is sorted by source line**, not `ast.walk` order. `ast.walk` is breadth-first, which buries `roll_apply_daily` (mid-pipeline, inside a loop) behind every top-level statement.
- **`file` lines are names only, no content hash** — the goal is catching a drop, not policing edits. Hashing would fire on every `.qmd`→`.html` re-render and get ignored.
- **`char` pins names, never counts** — a count doesn't tell a reviewer *which* one vanished.
- **`chlog` headers are verbatim, not date-parsed** — the format isn't uniform; the oldest five entries are `MM-DD-YYYY` (`02-19-2021` has no month 19).
- **A missing dispatch target raises** rather than emitting nothing. A spec that silently no-ops advertises coverage it doesn't provide.
- **Function-level reachability was rejected**: 17 of 300 defs are unreachable and 5 genuinely dead, so it needs a baseline allowlist that rots. Module granularity covers #247 with zero noise.

`.gitignore` needed a narrow negation: a blanket `*.txt` rule meant the manifest could never be committed (`Siccodes49.txt` survives only via force-add). The negation covers that file too, so `resources/` no longer depends on force-add.

No workflow change — `ci.yml` already runs `pytest tests/` unfiltered. No new dependencies; stdlib `ast` only.

### What this does not catch

Existence and shape, not behaviour. It would **not** have caught #247's trfd `COALESCE` removal — a pure inline edit that deletes nothing and renames nothing. Only an output-level test catches that class; that stays with #41/#97.

## Test plan

- [x] `pytest tests/` — 949 passed, 13s (locally excluding untracked `" 2.py"` junk).
- [x] `ruff check` / `ruff format --check` clean.
- [x] Regen is a fixed point — running `--regen` twice is byte-identical.
- [x] **Acceptance gate**: ran against PR #247's head in a scratch worktree. All four dropped things are caught — `chlog 10-07-2026`, `gen_comp_dsf`'s signature, `imp aux_functions -> compustat_correction`, and `test_nulls_return_above_10x`. The orphan check fails independently, naming `compustat_correction`.
- [x] Three-way character check fires when a row is removed from `cluster_labels.csv`.
- [x] Clean on `main`: zero orphans, zero diff.
- [x] Blind-spot probes: 0 test/fn lines collapse under de-duplication (was 34 hidden); 0 tracked files covered by no section (was 2).

## Review history

A high-effort review of the first cut found 9 defects, three of which defeated the manifest's purpose — the manifest could stay byte-identical while things were deleted. All are fixed in `811c59c`; the probes above are the regression evidence. Notably the first cut wired a dispatch spec to a `generate_factor_models` that doesn't exist on this branch, so it matched nothing and emitted zero lines while the docstring advertised coverage.

Worth stating plainly for reviewers: the original acceptance gate passed against #247 and that was **not** sufficient evidence. It proved the tool worked on one known incident and said nothing about the 34 tests it would quietly miss.

## Follow-ups (not in this PR)

- `_golden_helpers.py`'s `_link()` returns silently when a fixture is missing — `stage_synthetic_slices`'s docstring says missing files are "silently skipped". Deleting `synthetic_wrds/comp_fundq.parquet` raises nothing. Files *declared* in `RAW_FILES`/`INTERIM_FILES` should raise. Latent bug, deserves its own PR.
- `documentation.pdf` is a checked-in build output with no tracked build step; #247 edits the `.tex`/`.bib` without it, leaving it stale.


FernandoRDLL:
Hey @capellini 
Can you please review this? This is an idea I have but maybe there's a better way to implement?

capellini:
Absolutely @FernandoRDLL -- I'll take a look.

The git recovery steps that you provided in #247 are spot on. And the best line of defense is a careful code reviewer, as you have proven to be.

Some of the things this PR introduces are a little unconventional and may add complexity while also not being 100% fail-safe.

Let me get a better idea of what happened before providing an assessment and recommendation on how to proceed.

capellini:
Status update: I've finished tracing the mechanics of how the #247 regression happened. I have one last detail to confirm, and then I'll write up what I found and give feedback on this PR. I should have feedback soon. Thanks for your patience.

capellini:
Thanks for your patience @FernandoRDLL -- and thank you for both catching the #247 reversion in the first place and for creating a concrete proposal to identify this problem in the future. I've finished tracing the incident end to end, and here's my assessment.

The investigation behind this PR is sound, and two pieces of it are exactly what the repo needs:

- test_no_orphaned_modules -- this would have caught the #247 un-wiring automatically. It asserts against live code and I like that this test is self-anchored in that the structure of the code is, itself, the source of truth.
- test_characteristic_sources_agree -- closes a real, separate gap. Today you can delete 100 characteristics from cluster_labels.csv and CI passes. However, unlike the orphan test, whose invariant lives in the code structure itself, this one checks mutual agreement among three maintained artifacts, with no anchor saying which is correct. A test failure tells you the sources disagree, not which one is right, and an incomplete-but-legitimate change would cause a failure. The logic is sound and I think that the code base would benefit from keeping it. I just wanted to point out its maintenance profile is a bit heavier than the orphan test's.

I think that both are good additions. They extract cleanly from this PR into a small standalone test module along with the file-listing, import-graph, and reachability helpers they depend on.

I don't think we should adopt the manifest. The events surrounding PR #247 point to why. The reverting commit came from a stale workspace through GitHub's web upload path, which performs no fast-forward check. Content gets appended onto the current head no matter how old it is. If one replays the incident with the manifest in place, the check would fail once. Then the same stale workspace runs `--regen` against its own clobbered content, commits, and CI is green again. Detection still comes down to a reviewer reading a diff -- and the code review safeguard is what actually worked. Meanwhile the manifest requires updating on nearly every PR, making a ~2,000-line generated file a permanent regen-and-commit step and source of constant merge-conflicts. This would then train devs to reflexively regenerate the manifest file and reviewers to always expect a manifest file change, which actually effectively weakens the code review process.

A detail I want to point out from tracing the timeline is that the CI did briefly catch the reversion. The test jobs failed on 466472e, and they went green a few minutes later when the follow-up commit replaced the test file and the failing test went with it. The failure had come and gone before anyone had a chance to check it. That's the strength of the orphan test in particular. It stays red for as long as the module is unwired, no matter what happens to any individual test file.

Alongside the extraction I'm proposing two complementary guards that target the delivery path and the changelog directly:
1. a required GitHub Action check flagging non-merge web-created commits (committer noreply@github.com) with a maintainer-verified override label.
2. an append-only check on CHANGELOG_DATA.md at the dated-section level.

I'll share a full incident write-up with the team shortly that lays out the complete analysis. To keep the report visible only to those that need to see it, I'm going to create a private repository and post my findings there. The repository can serve as the place where we post other non-public content related to jkp-data.

My proposed next step is to keep the two aforementioned tests plus their helpers and drop the manifest and its regen machinery. What do you think?

PR #257, opened 2026-07-16
fix: make `jkp connect` actually verify the WRDS connection

### LINKED ISSUE #246: `jkp connect` reports success without actually testing the WRDS connection
## Problem

The `jkp connect` command's docstring says "Test or configure the WRDS connection", and it prints `Connected as: <username>` — but it never opens a connection to WRDS. It only resolves credentials locally via `get_wrds_credentials()` (env vars → keyring → file keyring) and echoes the username:

```python
creds = get_wrds_credentials()
typer.echo(f"Connected as: {creds.username}")
```

So `jkp connect` succeeds even when the credentials are wrong, expired, or WRDS is unreachable. Users get a misleading success signal and only discover the problem later, at `jkp build` time, deep into a run.

## Suggested fix

Have `jkp connect` actually establish a WRDS connection (e.g. open a `wrds.Connection` / run a trivial query) after resolving credentials, and report failure clearly if authentication or connectivity fails. `--reset` behavior can stay as is.

## Context

Flagged during review of PR #245; the behavior predates that PR (verified against `origin/main`), so it was out of scope there and is filed as a follow-up.

### LINKED ISSUE #274: WRDS credentials are persisted before they are verified

wrds_credentials.py (credential setup)

### Error Category

WRDS connection / authentication

### Bug Description

`jkp` persists WRDS credentials before it has any evidence they work. There is no
verification gate on the write path, so the first thing a user types becomes the
stored credential regardless of whether it authenticates.

The concrete consequence is a retry loop with no exit. In
`src/jkp/data/wrds_credentials.py`, `_prompt_and_store` prompts for the password and
immediately writes it to the system keyring (or `~/.pgpass` when no keyring is
available), then returns. Verification, where it happens at all, happens afterwards
and in a different module. So:

1. The user mistypes their password at the `jkp connect` prompt.
2. It is written to the keyring / `~/.pgpass`.
3. Verification fails and the command exits non-zero.
4. Every later `jkp connect` now *resolves* the stored bad password, so it never
   prompts again — it just fails identically, forever.

The only way out is `jkp connect --reset`, which the user has no particular reason
to know about. PR #257 adds a hint pointing at it, which makes the trap survivable
but does not remove it.

The same ordering applies on the `jkp build` path: `run_pipeline`
(`src/jkp/data/main.py:58`) calls `get_wrds_credentials()`, which reaches the same
prompt-and-store branch.

Framing this as the ordering rather than as the `jkp connect` symptom on purpose: a
credential store that accepts unverified input will keep producing variants of this
bug (a wrong username cached in the state dir, a stale keyring entry shadowing a
good `~/.pgpass` line) regardless of how many hints the CLI grows.

### Steps to Reproduce

1. Ensure no WRDS credentials are stored: `jkp connect --reset`.
2. Run `jkp connect` and deliberately enter a wrong password.
3. Verification fails; the command exits 1.
4. Run `jkp connect` again. It does not prompt — it resolves the stored bad
   password and fails with the same message. It will keep doing so indefinitely.
5. Inspect the store to confirm the bad password was persisted: the keyring entry
   for service `WRDS`, or the jkp line in `~/.pgpass`.

### Expected Behavior

A freshly prompted credential should be written to the keyring or `~/.pgpass` only
after it has been proven to authenticate against WRDS. If verification fails,
nothing is persisted and the next `jkp connect` prompts again — no `--reset`
required, because there is no bad state to reset.

Credentials that were *resolved* from an existing store (environment variables, an
existing keyring entry, an existing `~/.pgpass` line) are not rewritten, so this
only changes the freshly-prompted path.

### Proposed Fix

The obstacle is that `wrds_credentials` cannot call `verify_wrds_connection`
directly: the dependency runs `wrds_credentials <- wrds_connection`
(`wrds_connection.py:16` imports the WRDS host/port/db constants), so importing the
verifier back would be a cycle.

Inject the verifier instead of importing it. Add an optional
`verify: Callable[[str, str | None], None] | None = None` parameter to
`get_wrds_credentials()`, and have the freshly-prompted path call it *between* the
prompt and the store:

```python
password = getpass.getpass(...)
if not password:
    raise RuntimeError(...)
if verify is not None:
    verify(username, password)   # raises -> nothing below runs
_persist_username(username)
if _keyring_set(username, password):
    ...
```

`jkp connect` then passes `verify_wrds_connection` in, and stops calling it
separately, so every resolution path is verified exactly once and no path triggers
two Duo MFA pushes. With `verify=None` (the default) behavior is exactly as today,
so `run_pipeline` is unaffected until we decide to opt it in.

Injection is preferable to the alternative of splitting `_prompt_and_store` into a
prompt-without-store function plus an explicit store step orchestrated from the CLI,
for two reasons: it keeps the plaintext password out of the `connect()` frame (see
 #258 — Typer prints frame locals on typer <0.23.0, and `pyproject.toml` allows
typer>=0.15.0), and it keeps the store ordering in one place rather than making
every future caller responsible for getting it right.

Two known limits worth recording rather than silently accepting:

- On the no-keyring fallback the verification uses the typed password in the
  conninfo, then writes `~/.pgpass`. That proves the credential is correct; it does
  not separately prove the written `~/.pgpass` line parses. Re-verifying after the
  write would mean a second connection for a case `_write_pgpass` already covers.
- `_migrate_legacy_keyring` (`wrds_credentials.py:630`) writes `~/.pgpass` during
  what is otherwise a read-only resolution, so it remains an unverified persist.
  Out of scope here, noted so the invariant is not overstated.

Tests should cover: verify is called before anything is written; a failing verify
leaves the keyring and `~/.pgpass` untouched and propagates the error; `verify=None`
preserves today's behavior; and `jkp connect` passes the verifier through rather
than calling it itself.

Follow-up to the review of #257, where this ordering was identified but deliberately
scoped out of that PR.

### Operating System

N/A (affects all platforms)

### Python Version

3.11.x

### Pre-submission Checklist

FernandoRDLL:
Closes #246.
Closes #274.

## Simple explanation

`jkp connect` used to print `Connected as: <user>` after only resolving credentials locally — it never actually opened WRDS. So it reported success even when the credentials were wrong/expired or WRDS was unreachable, and users only discovered the problem later, deep into a `jkp build` run.

Now `jkp connect` opens a real WRDS connection (attaches the database read-only and runs a trivial query) before reporting success, so a green result means credentials, connectivity, and MFA actually work. On failure it prints an actionable, password-free message and exits non-zero instead of dumping a traceback.

Because opening the connection triggers the WRDS **Duo MFA push**, the command waits up to **25 seconds** for the user to approve it on their phone before giving up (the libpq `connect_timeout`). WRDS trusts a username/IP pair for 30 days after a successful approval, so expect a prompt on the first run from a new IP and again once that window expires — not on every run.

## Also includes #276 (verify before persist)

**Also includes the verify-before-persist work from #276.** That was opened as a separate follow-up, then merged into this branch (`467eb23`) rather than retargeted to `main`, so it ships here: roughly 217 lines adding `_prompt_verify_store`, `_verified`, and a `verify` parameter on `get_wrds_credentials`.

The mechanism is dependency injection rather than CLI-layer orchestration. `wrds_connection` imports from `wrds_credentials`, so `wrds_credentials` cannot import the verifier back without a cycle; instead the CLI passes `verify_wrds_connection` in as `get_wrds_credentials(verify=...)`. On the freshly-prompted path the check runs between the prompt and the store, so a mistyped password is never written and the next run simply re-prompts, with no `jkp connect --reset` needed. On the paths that read an existing store (env vars, keyring, `~/.pgpass`) it runs as a plain connectivity check. Every path verifies exactly once, so callers must not verify the result again: a second call would open a second WRDS connection and risk a second Duo push.

This also keeps the plaintext password out of the `connect()` frame on the prompt path, which the CLI-orchestrated alternative would not have done.

`run_pipeline` (`main.py:58`) calls `get_wrds_credentials()` with no argument, so `verify` defaults to `None` and the `jkp build` path is behaviorally unchanged.

## Technical details

- New module `src/jkp/data/wrds_connection.py` holds the WRDS connection primitives, extracted from `aux_functions.py`: `gen_wrds_connection_info` (now with an optional `connect_timeout`), `_attach_wrds`, `_install_postgres_extension`, and the conninfo/password-masking helpers, plus the new `verify_wrds_connection(username, password, *, connect_timeout=25)`. Keeping these in a standalone module lets the CLI verify a connection without importing the heavy pipeline module. Dependency direction is one-way: `wrds_credentials ← wrds_connection ← {aux_functions, cli}`.
- `aux_functions.py` re-imports the five primitives still used by the download workers; ~100 lines removed.
- `cli.py::connect` calls `verify_wrds_connection` before echoing success; the existing `except RuntimeError` surfaces the message and exits 1. `--reset` is unchanged. The docstring now states it verifies connectivity and when to expect a Duo MFA push.
- `verify_wrds_connection` wraps the attach/probe so **any** failure — including the `~/.pgpass` auth path where `password is None` and `_attach_wrds` re-raises the raw DuckDB exception — surfaces as a clean, password-free `RuntimeError` rather than a traceback. (Caught during review of this branch.)
- `connect_timeout` defaults to 25s so libpq doesn't give up before the user approves the Duo push.

No `CHANGELOG_DATA.md` entry — this is CLI/infra only and does not change anything written to `data/processed/`.

## Test plan

- `PYTHONPATH=src uv run --no-sync python -m pytest tests/unit/ -m "not wrds"` — **911 passed**.
- New `tests/unit/test_wrds_connection.py` (conninfo building/escaping, password masking relocated from `test_wrds_download.py`, `connect_timeout` present/absent, `verify_wrds_connection` raising a password-free `RuntimeError` on attach failure for both the password and `~/.pgpass`/`None` paths) and `tests/unit/test_cli_connect.py` (connect success / failure / no-password-leak).
- `ruff check` and `ruff format --check` clean; `pyright` on the changed source: 0 errors.
- Live check (manual, not in CI): `jkp connect` with valid creds returns cleanly after approving the MFA push; with wrong creds it prints the guidance message and exits 1.

/cc @capellini for review (infra/CLI change).

### DIFF
--- a/src/jkp/data/cli.py
+++ b/src/jkp/data/cli.py
@@ -120,7 +120,9 @@ def connect(
 
     Opens a real WRDS connection (attaches the database read-only and runs a
     trivial query), so a successful run confirms credentials, connectivity,
-    and MFA all actually work. This triggers a WRDS Duo MFA push on every run.
+    and MFA all actually work. This may trigger a WRDS Duo MFA push: WRDS trusts a
+    username/IP pair for 30 days after a successful approval, so expect one on the
+    first run from a new IP and again once that window expires.
 
     Credential precedence (highest first):
 
@@ -147,10 +149,15 @@ def connect(
             typer.echo("Credentials reset.")
             return
 
-        creds = get_wrds_credentials()
-
+        # Imported before resolving credentials, not after: this is the process's
+        # first `duckdb` import, and a broken native wheel (GLIBC mismatch on an HPC
+        # node) raises ImportError, which is not in the except tuple below and so
+        # escapes to Typer's pretty-exception handler. That handler prints frame
+        # locals on typer < 0.23.0, and pyproject allows typer>=0.15.0 — so no
+        # plaintext password may be bound in this frame when the import runs.
         from .wrds_connection import verify_wrds_connection
 
+        creds = get_wrds_credentials()
         verify_wrds_connection(creds.username, creds.password)
         typer.echo(f"Connected as: {creds.username}")
     except (RuntimeError, ValueError, OSError) as exc:
@@ -161,6 +168,14 @@ def connect(
         # Their messages are password-free; surface the message and exit non-zero rather
         # than dumping a traceback.
         typer.echo(str(exc), err=True)
+        # A mistyped password is stored before it is ever verified, so the next run
+        # finds it, never re-prompts, and fails identically forever. --reset is the
+        # only way out, and it belongs here rather than in the wrds_connection
+        # messages, which pipeline workers share and for which it is not the fix.
+        typer.echo(
+            "If a stored password is wrong, run `jkp connect --reset` and re-enter credentials.",
+            err=True,
+        )
         raise typer.Exit(1) from exc
 
 
--- a/src/jkp/data/cli.py
+++ b/src/jkp/data/cli.py
@@ -153,11 +153,13 @@ def connect(
 
         verify_wrds_connection(creds.username, creds.password)
         typer.echo(f"Connected as: {creds.username}")
-    except RuntimeError as exc:
-        # Credential resolution and connection verification raise RuntimeError
-        # for anticipated, actionable conditions (no/empty username, an
-        # unreadable ~/.pgpass, a failed WRDS attach). Surface the message and
-        # exit non-zero rather than dumping a traceback.
+    except (RuntimeError, ValueError, OSError) as exc:
+        # Anticipated, actionable failures from credential resolution and connection
+        # verification: RuntimeError (no/empty username, unreadable ~/.pgpass, a failed
+        # WRDS attach), ValueError (e.g. a password containing a newline), and OSError
+        # (e.g. an unwritable state dir when persisting the username / writing ~/.pgpass).
+        # Their messages are password-free; surface the message and exit non-zero rather
+        # than dumping a traceback.
         typer.echo(str(exc), err=True)
         raise typer.Exit(1) from exc
 
--- a/src/jkp/data/cli.py
+++ b/src/jkp/data/cli.py
@@ -120,7 +120,9 @@ def connect(
 
     Opens a real WRDS connection (attaches the database read-only and runs a
     trivial query), so a successful run confirms credentials, connectivity,
-    and MFA all actually work. This may trigger a WRDS Duo MFA push: WRDS trusts a
+    and MFA all actually work. A password entered at the prompt is verified
+    before it is stored, so a typo is never persisted and simply re-prompts on
+    the next run. This may trigger a WRDS Duo MFA push: WRDS trusts a
     username/IP pair for 30 days after a successful approval, so expect one on the
     first run from a new IP and again once that window expires.
 
@@ -157,8 +159,13 @@ def connect(
         # plaintext password may be bound in this frame when the import runs.
         from .wrds_connection import verify_wrds_connection
 
-        creds = get_wrds_credentials()
-        verify_wrds_connection(creds.username, creds.password)
+        # Injected rather than called on the result: on the freshly-prompted path the
+        # check has to run between the prompt and the store, which is inside
+        # get_wrds_credentials. It verifies every resolution path exactly once, so
+        # verifying again here would open a second connection (and risk a second Duo
+        # push). Passing the function also keeps the plaintext password out of this
+        # frame on the prompt path.
+        creds = get_wrds_credentials(verify=verify_wrds_connection)
         typer.echo(f"Connected as: {creds.username}")
     except (RuntimeError, ValueError, OSError) as exc:
         # Anticipated, actionable failures from credential resolution and connection
@@ -168,10 +175,12 @@ def connect(
         # Their messages are password-free; surface the message and exit non-zero rather
         # than dumping a traceback.
         typer.echo(str(exc), err=True)
-        # A mistyped password is stored before it is ever verified, so the next run
-        # finds it, never re-prompts, and fails identically forever. --reset is the
-        # only way out, and it belongs here rather than in the wrds_connection
-        # messages, which pipeline workers share and for which it is not the fix.
+        # Still worth pointing at: a password typed at the prompt is now verified
+        # before it is stored, but one that arrived from an env var, an entry
+        # predating this check, or a hand-edited ~/.pgpass can still be wrong, and
+        # resolution never re-prompts while a stored value exists. Kept at the CLI
+        # layer because the wrds_connection messages are shared with pipeline worker
+        # paths, where --reset is not the right advice.
         typer.echo(
             "If a stored password is wrong, run `jkp connect --reset` and re-enter credentials.",
             err=True,
--- a/src/jkp/data/cli.py
+++ b/src/jkp/data/cli.py
@@ -116,7 +116,11 @@ def connect(
         help="Reset stored WRDS credentials.",
     ),
 ) -> None:
-    """Test or configure the WRDS connection.
+    """Verify the WRDS connection, or configure/reset stored credentials.
+
+    Opens a real WRDS connection (attaches the database read-only and runs a
+    trivial query), so a successful run confirms credentials, connectivity,
+    and MFA all actually work. This triggers a WRDS Duo MFA push on every run.
 
     Credential precedence (highest first):
 
@@ -144,11 +148,16 @@ def connect(
             return
 
         creds = get_wrds_credentials()
+
+        from .wrds_connection import verify_wrds_connection
+
+        verify_wrds_connection(creds.username, creds.password)
         typer.echo(f"Connected as: {creds.username}")
     except RuntimeError as exc:
-        # Credential resolution raises RuntimeError for anticipated, actionable
-        # conditions (no/empty username, an unreadable ~/.pgpass). Surface the
-        # message and exit non-zero rather than dumping a traceback.
+        # Credential resolution and connection verification raise RuntimeError
+        # for anticipated, actionable conditions (no/empty username, an
+        # unreadable ~/.pgpass, a failed WRDS attach). Surface the message and
+        # exit non-zero rather than dumping a traceback.
         typer.echo(str(exc), err=True)
         raise typer.Exit(1) from exc
 
--- a/src/jkp/data/wrds_connection.py
+++ b/src/jkp/data/wrds_connection.py
@@ -39,7 +39,7 @@ def _sql_literal(value: str) -> str:
 
 
 def gen_wrds_connection_info(
-    user, password: str | None = None, *, connect_timeout: int | None = None
+    user: str, password: str | None = None, *, connect_timeout: int | None = None
 ) -> str:
     """Build a libpq conninfo for WRDS.
 
@@ -97,17 +97,22 @@ def _attach_wrds(con: duckdb.DuckDBPyConnection, conninfo: str, password: str | 
     """ATTACH the WRDS Postgres database read-only on an existing DuckDB connection.
 
     DuckDB's postgres extension embeds the full connection string (including the password) in
-    ATTACH error text, so on failure suppress the original exception and raise a generic,
+    ATTACH error text, so on failure drop the original exception and raise a generic,
     password-free error. Errors that don't contain the password propagate unchanged.
     """
+    leaked = False
     try:
         con.execute(f"ATTACH '{_sql_literal(conninfo)}' AS wrds (TYPE postgres, READ_ONLY)")
     except Exception as e:
-        if password and _password_in_error(str(e), password):
-            raise RuntimeError(
-                "Failed to attach WRDS connection. Check credentials and MFA approval."
-            ) from None
-        raise
+        if not (password and _password_in_error(str(e), password)):
+            raise
+        leaked = True
+    if leaked:
+        # Raised after the except block exits rather than with `from None` inside it. The
+        # interpreter clears the handled exception on exit, so __context__ is None here;
+        # `from None` only hides the chain from printed tracebacks, leaving the DuckDB
+        # error (whose text embeds the password) reachable on the exception object.
+        raise RuntimeError("Failed to attach WRDS connection. Check credentials and MFA approval.")
 
 
 def _install_postgres_extension() -> None:
@@ -135,6 +140,7 @@ def verify_wrds_connection(
     it must leave the user time to approve the Duo MFA push, so it defaults to 25s.
     """
     conninfo = gen_wrds_connection_info(username, password, connect_timeout=connect_timeout)
+    failed = False
     try:
         # Inside the try: INSTALL can itself fail (e.g. no network to DuckDB's
         # extension repo, plausible on the headless HPC nodes this targets), and
@@ -149,13 +155,15 @@ def verify_wrds_connection(
         # the failure text embeds the password; pass it through unchanged.
         raise
     except Exception:
-        # Any other failure (a failed INSTALL/LOAD, or the ~/.pgpass auth path where
-        # password is None and _attach_wrds re-raises the raw DuckDB exception, or a
-        # query error). Surface one actionable, password-free message so callers that
-        # only handle RuntimeError (e.g. `jkp connect`) exit cleanly instead of a
-        # traceback. Use `from None`, not `from e`: the raw DuckDB error can embed the
-        # conninfo/password, and chaining it as __cause__ would carry the secret on the
-        # exception object even though this message is clean.
+        # Any other failure (a failed INSTALL/LOAD, the ~/.pgpass auth path where password
+        # is None and _attach_wrds re-raises the raw DuckDB exception, or the probe query
+        # failing). Recorded here and raised below.
+        failed = True
+    if failed:
+        # Raised outside the handler for the same reason as in _attach_wrds: the raw DuckDB
+        # error can embed the conninfo and password, and `from None` would leave it on
+        # __context__. One actionable, password-free message also lets callers that only
+        # handle RuntimeError (e.g. `jkp connect`) exit cleanly instead of dumping a traceback.
         raise RuntimeError(
             "Failed to connect to WRDS. Check your network, credentials, and MFA approval."
-        ) from None
+        )
--- a/src/jkp/data/wrds_connection.py
+++ b/src/jkp/data/wrds_connection.py
@@ -123,7 +123,7 @@ def _install_postgres_extension() -> None:
 
 
 def verify_wrds_connection(
-    username: str, password: str | None, *, connect_timeout: int = 10
+    username: str, password: str | None, *, connect_timeout: int = 25
 ) -> None:
     """Open a real WRDS connection and confirm it is queryable.
 
@@ -131,8 +131,8 @@ def verify_wrds_connection(
     it. The ATTACH authenticates eagerly (opening the libpq connection triggers the
     WRDS Duo MFA push), so a successful return means credentials, connectivity, and
     MFA all succeeded. Raises :class:`RuntimeError` with a password-free message on
-    any failure. ``connect_timeout`` bounds how long libpq waits on an unreachable
-    host before failing.
+    any failure. ``connect_timeout`` bounds how long libpq waits before failing —
+    it must leave the user time to approve the Duo MFA push, so it defaults to 25s.
     """
     conninfo = gen_wrds_connection_info(username, password, connect_timeout=connect_timeout)
     _install_postgres_extension()
--- a/src/jkp/data/wrds_connection.py
+++ b/src/jkp/data/wrds_connection.py
@@ -1,0 +1,142 @@
+"""WRDS connection primitives.
+
+Builds the libpq conninfo for the WRDS Postgres endpoint, attaches it read-only
+on a DuckDB connection, and verifies connectivity. Kept separate from the heavy
+``aux_functions`` pipeline module so the CLI's ``jkp connect`` command can check
+a connection without importing the whole pipeline.
+
+Password handling: the password only ever appears inside the conninfo and the
+DuckDB ATTACH statement. DuckDB's postgres extension echoes the full connection
+string (password included) in ATTACH error text, so the masking helpers here
+detect and redact every escaped form the password can take in an error message.
+"""
+
+import duckdb
+
+from .wrds_credentials import WRDS_DB, WRDS_HOST, WRDS_PORT
+
+
+def _pg_escape_value(value: str) -> str:
+    """libpq-escape a conninfo value (user or password) for a single-quoted field.
+
+    libpq accepts single-quoted values with backslash-escaped ``\\`` and ``'``,
+    so quoting lets a value hold spaces or special characters without breaking the
+    conninfo. For the password this is also the form that appears in any error text
+    echoing the connection string, so the credential-masking checks reuse it rather
+    than matching the raw password.
+    """
+    return value.replace("\\", "\\\\").replace("'", "\\'")
+
+
+def _sql_literal(value: str) -> str:
+    """Escape a string for embedding inside a single-quoted DuckDB SQL literal.
+
+    The conninfo is interpolated into ``ATTACH '...'`` / ``postgres_scan('...')``
+    SQL, so any single quote it contains (e.g. around a libpq-quoted password)
+    must be doubled or it terminates the SQL string literal.
+    """
+    return value.replace("'", "''")
+
+
+def gen_wrds_connection_info(
+    user, password: str | None = None, *, connect_timeout: int | None = None
+) -> str:
+    """Build a libpq conninfo for WRDS.
+
+    When ``password`` is ``None`` the ``password=`` field is omitted, so libpq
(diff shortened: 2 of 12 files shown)


FernandoRDLL:
@claude review this PR thoroughly and make suggestions if appropriate

FernandoRDLL:
Thanks @claude — both valid.

- **INSTALL not wrapped:** fixed in 1d62476. Moved `_install_postgres_extension()` inside the `try` so a failed `INSTALL postgres` (no network to the extension repo on a headless node) surfaces as the same password-free `RuntimeError` rather than a raw traceback. Added `test_install_extension_failure_wrapped_as_runtime_error` covering that path.
- **Stale comment:** updated `wrds_credentials.py` to point at `wrds_connection.gen_wrds_connection_info`.

17 connection/cli tests green, ruff + format clean.

FernandoRDLL:
Addressed the review:
- Wrapped the INSTALL failure and fixed the stale `wrds_credentials` comment (1d62476).
- Broadened `connect`'s handler to also catch `ValueError`/`OSError` from credential resolution so it exits cleanly instead of a traceback (af9faa1).

tests/unit/test_wrds_download.py
```diff
@@ -383,7 +230,15 @@ def mock_duckdb_multi(self):
...
         """
-        with patch("jkp.data.aux_functions.duckdb") as mock:
+        # Patch duckdb in both modules that open connections on the download path:
+        # the per-worker ATTACH connections live in aux_functions, while the one
+        # up-front INSTALL connection is opened by wrds_connection._install_postgres_extension.
+        # Pointing both names at the same mock keeps a single ordered connect() sequence
+        # (install consumes conns[0], workers consume conns[1:]).
+        with (
+            patch("jkp.data.aux_functions.duckdb") as mock,
+            patch("jkp.data.wrds_connection.duckdb", mock),
```
capellini, line 240: Three tests in `TestDateRangeSplitting` kept their own inline `patch("jkp.data.aux_functions.duckdb")` blocks (`test_parallel_download_aggregates_concat_failures`, `test_parallel_download_splits_giant_tables`, `test_parallel_download_concatenates_each_split_table`) and now run a real `INSTALL postgres;`. The reason is that `_install_postgres_extension` moved to `wrds_connection`, so its `duckdb` lookup resolves there and a single-module patch no longer covers it. This fixture dual-patches both modules for exactly that reason. The three inline blocks just didn't get the same treatment.

Confirmed by running one of them alone with a fresh `HOME`. On this branch it downloads the extension from extensions.duckdb.org, while on `main` it makes no network connections. The full suite doesn't show it because the real-INSTALL integration test in `test_wrds_connection.py` runs first and warms the cache. But on a network-isolated machine (the HPC case this PR targets) or a cold-cache subset run, these three tests error, and unlike the integration test they have no skip guard.

Fix: use `mock_duckdb_multi` in those three tests, or add `patch("jkp.data.wrds_connection.duckdb", mock_duckdb)` alongside each inline patch.

src/jkp/data/cli.py
```diff
@@ -144,11 +148,18 @@ def connect(
             return
 
         creds = get_wrds_credentials()
+
+        from .wrds_connection import verify_wrds_connection
```
capellini, line 152: There is a password leak path here on older typer versions. `creds` binds the plaintext password as a frame local, and this deferred import is the process's first import of `duckdb`. If that import fails (broken native wheel, GLIBC mismatch on an HPC node), `ImportError` is not in the except tuple below, so the exception escapes to Typer's pretty exception handler. On typer older than 0.23.0 the default for `pretty_exceptions_show_locals` was `True`, so the handler prints the frame's locals, including the `Credentials` repr with the plaintext password. `pyproject.toml` allows `typer>=0.15.0`, so a fresh install can legally resolve a leaking version. Dev environments have 0.24.2 (the default flipped in 0.23.0), which is why this never reproduces locally. Reproduced in a scratch venv: typer 0.22.0 prints `password='...'` in the locals panel, typer 0.24.2 does not.

Suggested fix for this PR: move this import above `creds = get_wrds_credentials()`. Then a broken duckdb install fails before any secret exists in the frame, and no except-tuple change is needed. The underlying hazard (the generated `Credentials` repr prints the password) predates this PR and is being fixed independently in #259 (issue #258), which masks the repr and disables `pretty_exceptions_show_locals` on the app, so there is no need to address either of those here.

src/jkp/data/wrds_connection.py
```diff
@@ -0,0 +1,161 @@
...
+        # Any other failure (a failed INSTALL/LOAD, or the ~/.pgpass auth path where
+        # password is None and _attach_wrds re-raises the raw DuckDB exception, or a
+        # query error). Surface one actionable, password-free message so callers that
+        # only handle RuntimeError (e.g. `jkp connect`) exit cleanly instead of a
+        # traceback. Use `from None`, not `from e`: the raw DuckDB error can embed the
+        # conninfo/password, and chaining it as __cause__ would carry the secret on the
+        # exception object even though this message is clean.
+        raise RuntimeError(
+            "Failed to connect to WRDS. Check your network, credentials, and MFA approval."
+        ) from None
```
capellini, line 161: Low severity: in the standard CLI and Slurm workflow this is safe as written, since the default traceback machinery honors `__suppress_context__` and `jkp connect` catches and discards this RuntimeError. Flagging it because the comment's claim is not quite what `from None` does. `raise ... from None` sets `__cause__` to None and suppresses the chain from printed tracebacks, but it does not remove the original exception from the object: `__context__` still references the raw DuckDB error, whose message can embed the conninfo and password (verified by execution in this repo's venv). The secret stays reachable to anything that walks the exception object rather than printing it, such as a crash reporter or a post-mortem debugger.

Since the stated intent is that the exception object carries no secret, suggest making that literally true by raising outside the handler. When an except block exits, the interpreter clears the active exception, so the raise has nothing to chain to:

```python
    failed = False
    try:
        _install_postgres_extension()
        with duckdb.connect(":memory:") as con:
            con.execute("LOAD postgres;")
            _attach_wrds(con, conninfo, password)
            con.execute("SELECT 1 FROM wrds.information_schema.schemata LIMIT 1")
    except RuntimeError:
        raise
    except Exception:
        failed = True
    if failed:
        raise RuntimeError(
            "Failed to connect to WRDS. Check your network, credentials, and MFA approval."
        )
```

A one-line assertion in the existing verify tests (`exc.__context__ is None`) would pin it. The same applies to the `from None` in `_attach_wrds` above, which is pre-existing code moved here, but whose branch fires exactly when the password is confirmed present in the error text, so its `__context__` provably carries the secret.

src/jkp/data/cli.py
```diff
@@ -144,11 +148,18 @@ def connect(
...
+
+        verify_wrds_connection(creds.username, creds.password)
         typer.echo(f"Connected as: {creds.username}")
-    except RuntimeError as exc:
-        # Credential resolution raises RuntimeError for anticipated, actionable
-        # conditions (no/empty username, an unreadable ~/.pgpass). Surface the
-        # message and exit non-zero rather than dumping a traceback.
+    except (RuntimeError, ValueError, OSError) as exc:
```
capellini, line 156: When the interactive first run stores a mistyped password, this failure path becomes a trap. `_prompt_and_store` persists the typed password (keyring or ~/.pgpass) before verification runs, so the next `jkp connect` finds the stored bad password, never prompts again, and fails with the same message on every rerun. The only way out is `jkp connect --reset`, which neither failure message mentions. The store-before-verify ordering predates this PR, and pre-PR the trap was worse (connect reported success and the user discovered the bad password mid `jkp build`), so this is not a regression, just the missing half of this PR's goal: a passing `jkp connect` now proves the connection works, and a failing one should tell the user how to get back to a working state.

Suggested fix: after echoing the error here, also echo a hint along the lines of "If a stored password is wrong, run `jkp connect --reset` and re-enter credentials." Keeping the hint at the CLI layer is deliberate, since the RuntimeError messages in `wrds_connection` are shared with pipeline worker paths where a `--reset` hint would not always be the right advice.

The fuller fix, verifying before persisting, is intentionally not requested in this PR. The dependency direction is `wrds_credentials <- wrds_connection`, so `_prompt_and_store` cannot call `verify_wrds_connection` without an import cycle. Store-after-verify would need CLI-layer orchestration and new API surface on `wrds_credentials` (a prompt-without-store variant plus an explicit store step, with the pgpass path holding the password in memory for the verify). If we want that, it should be its own issue rather than part of this PR.

tests/unit/test_wrds_connection.py
```diff
@@ -0,0 +1,297 @@
...
+    def test_password_none_omits_password_field(self):
+        """password=None (no positional password arg) omits password= entirely."""
+        from jkp.data.wrds_connection import gen_wrds_connection_info
+
+        result = gen_wrds_connection_info("u", None)
+
+        assert "password=" not in result
+
+
+class TestVerifyWrdsConnection:
```
capellini, line 189: One branch of `verify_wrds_connection` is untested: ATTACH succeeds and the probe query fails. The class covers the success path, both ATTACH failure flavors, and INSTALL failure, but the probe is the step this function exists for (as the success test's docstring puts it, it confirms the DB is queryable, not just attached), and its failure leg is a real WRDS scenario: valid login with revoked or limited product permissions, or a connection dropped between the attach and the first query. Runtime behavior is already correct (the generic handler wraps it), so this is just a coverage gap. A test in the same shape as the existing FakeConnection ones would close it: let `execute` succeed for the ATTACH and raise for SQL containing `information_schema`, and assert the password-free RuntimeError.

tests/unit/test_wrds_connection.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def test_connect_timeout_omitted_by_default(self):
+        """Without an explicit connect_timeout, libpq's default (no timeout) applies."""
+        from jkp.data.wrds_connection import gen_wrds_connection_info
+
+        result = gen_wrds_connection_info("u", "p")
+
+        assert "connect_timeout" not in result
+
+    def test_password_none_omits_password_field(self):
```
capellini, line 180: Two trivial ones: this test is a strict subset of `test_password_omitted_when_none` above (same call, weaker assertions), so one of the two can go. And in `gen_wrds_connection_info`, the `user` parameter is now the only unannotated one on the signature this PR touched, `user: str` would complete it.

src/jkp/data/cli.py
```diff
@@ -116,7 +116,11 @@ def connect(
         help="Reset stored WRDS credentials.",
     ),
 ) -> None:
-    """Test or configure the WRDS connection.
+    """Verify the WRDS connection, or configure/reset stored credentials.
+
+    Opens a real WRDS connection (attaches the database read-only and runs a
+    trivial query), so a successful run confirms credentials, connectivity,
+    and MFA all actually work. This triggers a WRDS Duo MFA push on every run.
```
capellini, line 123: Small accuracy point on "This triggers a WRDS Duo MFA push on every run": WRDS validates the username/IP combination for 30 days after a successful MFA, so the push fires on the first connection from a given IP and again after expiry or an IP change, not on every run (WRDS doc: https://wrds-www.wharton.upenn.edu/pages/about/log-in-to-wrds-using-two-factor-authentication/multi-factor-authentication-for-sasconnect-and-postgresql/). That matches the existing `--persistent-connection` help text, which only makes sense under IP-based trust. Suggest rewording to something like "This may trigger a WRDS Duo MFA push (expect one on the first run from a new IP, or after the 30 day MFA window expires)." Worth mirroring in the PR description too, since it makes the same claim.

capellini (changes requested):
Thanks for this one. The core is in very good shape: `jkp connect` now genuinely proves connectivity, the extraction into `wrds_connection.py` is a clean move (I diffed the moved functions against main and they are unchanged, the re-imports and test patch targets in the shared fixtures all resolve, and the new module gets full coverage from its own tests), and the conninfo escaping and password redaction held up under deliberate attempts to break them.

Requesting changes for one confirmed test regression, plus a handful of smaller items. Details are in the inline comments:

1. Three `TestDateRangeSplitting` tests now perform a real `INSTALL postgres` (network access) because their inline duckdb patches no longer reach the moved `_install_postgres_extension`. Confirmed empirically. This is the one blocking item.
2. A password leak path via Typer traceback locals on typer older than 0.23.0. The fix here is a one line import reorder, with the underlying repr hazard handled separately in #258/#259.
3. The `from None` comment overstates what Python guarantees: `__context__` still references the original exception. Low severity, suggested restructure inline.
4. A mistyped password gets stored before verification runs, and the failure message never mentions `jkp connect --reset`, leaving the user stuck in a retry loop.
5. One branch of `verify_wrds_connection` is untested (attach succeeds, probe query fails).
6. Docstring accuracy: the Duo push fires per username/IP per 30 day window, not on every run.
7. Two trivial nits (a duplicate test, one missing annotation).

The numbering is reading order, not priority, except that item 1 is the reason for the verdict.

tests/unit/test_wrds_download.py
FernandoRDLL: Fixed in c675603. Went with the first option: `mock_duckdb_multi` is now a module-level fixture and all three tests use it, so the dual patch lives in one place instead of being re-hand-rolled per test. The three inline blocks also allocated 40 connections where the fixture's 32 is already more than enough (these run at `max_workers=4`).

Reproduced your finding and confirmed the fix the same way. With a fresh `HOME`, `test_parallel_download_splits_giant_tables` alone takes 9.16s on the old code and writes `.duckdb/extensions/v1.5.2/osx_arm64/postgres_scanner.duckdb_extension`. After the change the whole file is 43 tests in 0.27s with nothing downloaded.

src/jkp/data/cli.py
FernandoRDLL: Fixed in 41d7e79 — the import now sits above `creds = get_wrds_credentials()`, with a comment recording why the ordering matters so it doesn't drift back. Agreed on leaving the repr hazard to #259; this just removes the frame that would expose it.

Worth noting the reset path still doesn't import duckdb, since the import stays below the `--reset` early return.

src/jkp/data/wrds_connection.py
FernandoRDLL: You're right, and I confirmed it before changing anything:

```
from None __context__: ValueError('secret=hunter2')
raise-outside-handler __context__: None | __cause__: None
```

Took your restructure in e18bcc1, and applied it to `_attach_wrds` as well — as you say, that's the branch where `__context__` provably holds the password, since it only fires once the password has been found in the error text. Pinned with `exc.__context__ is None` in the attach-failure, pgpass, and new probe tests.

The comments that claimed `from None` dropped the original exception were wrong in the same way, so they've been rewritten to say what actually makes `__context__` empty.

src/jkp/data/cli.py
FernandoRDLL: Added the hint in 41d7e79, at the CLI layer as you suggested:

```
If a stored password is wrong, run `jkp connect --reset` and re-enter credentials.
```

It goes to stderr right after the error message, so it covers the credential-resolution failures too, not just verification.

Agreed on leaving verify-before-persist out of this PR — the import cycle makes it a real API change on `wrds_credentials` rather than a tweak. Happy to open an issue for it if you want it tracked.

tests/unit/test_wrds_connection.py
FernandoRDLL: Added `test_probe_query_failure_wrapped_as_runtime_error` in e18bcc1, in the shape you described: the `FakeConnection` lets the ATTACH through and raises on the SQL containing `information_schema`, then asserts the RuntimeError is password-free (and that `__context__` is None). `wrds_connection.py` is now at 100% statement and branch coverage.

tests/unit/test_wrds_connection.py
FernandoRDLL: Both done in e18bcc1 — dropped `test_password_none_omits_password_field` (kept `test_password_omitted_when_none`, which is the stronger of the two) and annotated `user: str`.

src/jkp/data/cli.py
FernandoRDLL: Reworded in 41d7e79, and mirrored in the PR description:

> This may trigger a WRDS Duo MFA push: WRDS trusts a username/IP pair for 30 days after a successful approval, so expect one on the first run from a new IP and again once that window expires.

Thanks for the link — the IP-based trust is also what makes the 25s `connect_timeout` reasonable rather than something users hit on every invocation.

FernandoRDLL:
Thanks — all seven are addressed across three commits. Replies are on the individual threads; summary here.

**c675603 — the blocking one.** `mock_duckdb_multi` is now module-scoped and used by the three `TestDateRangeSplitting` tests, so the dual patch is defined once instead of re-hand-rolled. I reproduced your finding first: with a fresh `HOME`, `test_parallel_download_splits_giant_tables` alone takes 9.16s on the old code and writes `postgres_scanner.duckdb_extension`; after the change the whole file is 43 tests in 0.27s with nothing downloaded.

**e18bcc1 — exception chaining and coverage.** You were right about `from None`; confirmed by execution before changing anything. Raising outside the handler now, in `_attach_wrds` as well as `verify_wrds_connection`, with `exc.__context__ is None` pinned in three tests. Added the probe-failure test, so `wrds_connection.py` is at 100% statement and branch coverage. Also dropped the duplicate test and annotated `user: str`.

**41d7e79 — CLI.** The duckdb import moved above `get_wrds_credentials()`, so no password is bound in the frame if the import fails; the reset path still doesn't import duckdb. Failure paths now point at `jkp connect --reset`. The Duo MFA docstring is corrected, and mirrored in the PR description.

Not done, per your note: verify-before-persist stays out of this PR. Say the word and I'll open an issue for it.

Full unit suite: 916 passed. `ruff check` and `ruff format --check` clean.

FernandoRDLL:
@capellini — all seven items are addressed and I've resolved the threads. CI is green on both 3.11 and 3.12, and the full unit suite is at 916 passed.

Quick recap of what changed since your review:

- The blocking test regression is fixed, and I verified it the way you did: fresh `HOME`, one test alone was 9.16s with a real `postgres_scanner.duckdb_extension` download; it's now 43 tests in 0.27s with nothing downloaded.
- Your `__context__` point was correct — confirmed by execution before I touched anything. Both `_attach_wrds` and `verify_wrds_connection` now raise outside the handler, pinned by `exc.__context__ is None` in three tests.
- Probe leg is covered; `wrds_connection.py` is at 100% statement and branch coverage.
- CLI: import moved above the credential resolution, `--reset` hint added, MFA docstring corrected here and in the PR description.

Verify-before-persist is deliberately still out of scope here, as you suggested. I'm opening it as its own issue and PR — will link them from this thread shortly.

Could you take another look and approve if it looks right to you?

FernandoRDLL:
Opened the verify-before-persist follow-up you scoped out of this PR: #274 (issue) and #276 (PR).

#276 is based on this branch rather than `main`, since it builds on `verify_wrds_connection`, which only exists here. It'll need retargeting to `main` once this merges.

One thing worth your eye there, since it changes the shape you sketched: rather than splitting `_prompt_and_store` into a prompt-without-store function plus a CLI-orchestrated store step, I injected the verifier into `get_wrds_credentials(verify=...)`. Two reasons — the CLI-orchestration version binds the plaintext password in the `connect()` frame across the verify, which is the hazard we just closed at `cli.py:152`; and it keeps the store ordering in one place instead of making every future caller responsible for it. Happy to switch if you'd rather have the explicit two-step API.

capellini (approved):
Approving. All seven items from the earlier review are addressed, and I checked the two most important by execution rather than by reading: the `__context__` chaining really is empty on both the attach and verify paths, and the date-splitting tests no longer download the postgres extension.

I agree with the dependency injection refactor, so I'm keeping `get_wrds_credentials(verify=...)`. Your reasoning holds. The CLI-orchestrated alternative binds the plaintext password in the `connect()` frame across the verify, which is the hazard #259 exists to close, and the injected form keeps store ordering in one place instead of making every future caller responsible for it. The cost is that a function named "get credentials" can now open a network connection and trigger a Duo push, which is why the "callers must not verify the result again" caveat has to live in the docstring. I think that is the right trade-off, but it is worth remembering the type signature does not enforce it.

I pushed one commit (`28922f8`) rather than sending it back to you, to save a round trip.

`test_password_never_appears_in_output` could not fail. Its `password` local was never passed to any code under test: the mocked resolver raised a hardcoded, password-free message, so the string only ever existed in the test's own scope. I mutated `connect()` to print the plaintext password on success and all five tests in the file still passed. The assertion moved to `test_success_exits_zero_and_reports_username`, where the password actually reaches the command; that version fails on the same mutation. The deleted test was otherwise a duplicate of `test_failure_exits_nonzero_...`: same mocks, same message, byte-identical stdout/stderr/exit code, and the identical set of executed lines in `cli.py`. Coverage is unchanged.

I also updated three docs to keep them in sync with the changes in this branch. The README told users to run `jkp connect --reset` whenever a password needed changing, which was advice born of the exact trap verify-before-persist removes, and it never mentioned that `jkp connect` now verifies the connection. `wrds_connection.py` is a new module and appeared in no markdown file, so it is now in the CONTRIBUTING.md tree and the CLAUDE.md source list.

One thing for after the merge: the PR description now carries `Closes #274` as well as `Closes #246`. #274 would not have closed on its own, because its closing keyword was in #276, which merged into this branch rather than into `main`.

PR #259, opened 2026-07-21
fix: mask WRDS password in Credentials repr

### LINKED ISSUE #258: Credentials dataclass repr exposes the WRDS password to locals-rendering tools

wrds_credentials.py (credential setup)

### Error Category

WRDS connection / authentication

### Bug Description

The frozen dataclass `Credentials` in `src/jkp/data/wrds_credentials.py` holds the
WRDS password in plaintext, and its auto-generated dataclass repr prints it:

```
Credentials(username='...', password='<plaintext>')
```

Any tool that renders frame locals via `repr()` can therefore print the password.
That includes Typer's pretty exception handler (on typer < 0.23.0,
`pretty_exceptions_show_locals` defaulted to True, and `pyproject.toml` allows
typer>=0.15.0), pytest `--showlocals`, interactive debuggers, and crash reporters.

The hazard is the default dataclass repr, not any single call site. Wherever a
`Credentials` object is a local in a frame that gets rendered, the secret is
exposed.

This was surfaced during review of PR #257 (which adds a fallible statement after
`creds` is bound in the `connect` CLI command, widening the window where an
exception could render that frame's locals), but the unsafe repr predates that PR
and the fix is independent of it.

### Steps to Reproduce

1. Construct a `Credentials("user", "secret")` (as `get_wrds_credentials()` does
   for the env-var and keyring paths).
2. Trigger anything that renders the object via `repr()`: an unhandled exception
   under a pretty-traceback handler that shows locals, `pytest --showlocals`, a
   debugger `p creds`, or a plain `print(creds)`.
3. The plaintext password appears in the rendered output.

### Expected Behavior

`repr()` of a `Credentials` object should never reveal the plaintext password. It
should mask the secret while still distinguishing `password=None` (which in this
codebase means `~/.pgpass` authentication, a distinction that is useful when
debugging) from a non-None password.

### Proposed Fix

Two small, self-contained changes:

1. Add a custom `__repr__` to the `Credentials` dataclass that masks the password
   but preserves the None vs non-None distinction. A user-defined `__repr__` in
   the class body is respected by `@dataclass` (it is not overwritten), so no
   decorator change is needed. The username is intentionally not masked, since the
   CLI already prints it. An empty-string password is masked too.

   ```python
   def __repr__(self) -> str:
       pw = "***" if self.password is not None else None
       return f"Credentials(username={self.username!r}, password={pw!r})"
   ```

2. Set `pretty_exceptions_show_locals=False` on the `typer.Typer(...)` app in
   `src/jkp/data/cli.py` as defense in depth for the same hazard (typer only
   changed this default in 0.23.0, and the project floor is typer>=0.15.0).

Add unit tests covering: repr with a password (username present, `***` present,
real value absent), repr with `password=None` (still shown as None), repr with an
empty-string password (still masked), and that `str()` is likewise password-free.

This will be fixed in its own PR so a future revert of #257 cannot remove the
protection.

### Operating System

N/A (affects all platforms)

### Python Version

3.11.x

### Pre-submission Checklist

capellini:
## Summary

Masks the WRDS password in the `Credentials` dataclass repr so it can no longer be
leaked by tools that render frame locals.

Closes #258

## The hazard

The frozen `Credentials` dataclass in `src/jkp/data/wrds_credentials.py` holds the
WRDS password in plaintext. Its auto-generated dataclass repr printed the secret:

```
Credentials(username='...', password='<plaintext>')
```

Wherever a `Credentials` object is a local in a frame that gets rendered via
`repr()`, the password is exposed. That includes Typer's pretty exception handler
(on typer < 0.23.0, `pretty_exceptions_show_locals` defaulted to True, and
`pyproject.toml` allows typer>=0.15.0), pytest `--showlocals`, interactive
debuggers, and crash reporters. The hazard is the default repr, not any single
call site.

## The fix

1. A custom `__repr__` on the `Credentials` dataclass that masks the password with
   `***` but preserves the None vs non-None distinction. `password=None` marks
   `~/.pgpass` authentication in this codebase, and keeping that visible is useful
   when debugging. An empty-string password is masked too. The username is left
   unmasked since the CLI already prints it. A user-defined `__repr__` in the class
   body is respected by `@dataclass` (it is not overwritten), so no decorator
   change is needed.

2. `pretty_exceptions_show_locals=False` on the `typer.Typer(...)` app in
   `src/jkp/data/cli.py`, as defense in depth for the same hazard. Typer only
   changed this default in 0.23.0, and the project floor is typer>=0.15.0, so it is
   set explicitly.

## Tests

New unit tests in `tests/unit/test_wrds_credentials.py`:

- repr with a password: username present, `***` present, real value absent
- repr with `password=None`: still shown as None (the pgpass path stays
  distinguishable)
- repr with an empty-string password: still masked, real value absent
- `str()` is likewise password-free (str falls through to `__repr__`)

## Why this is a standalone PR

The unsafe repr predates PR #257 and the fix is independent of it. This was
surfaced during review of #257 (which adds a fallible statement after `creds` is
bound in the `connect` command, widening the window where an exception could
render that frame's locals), but keeping this fix in its own PR means a future
revert of #257 cannot remove the protection.

No `CHANGELOG_DATA.md` entry: this changes nothing written to `data/processed/`.

## Verification

- `uv run pytest tests/unit/` (all pass)
- `uv run ruff check src/jkp/data/ tests/` and `ruff format --check` (clean)
- `uv run pyright src/jkp/data/` (no new findings in the changed files)

### DIFF
--- a/src/jkp/data/wrds_credentials.py
+++ b/src/jkp/data/wrds_credentials.py
@@ -71,6 +71,15 @@ class Credentials:
     username: str
     password: str | None  # None => authenticate via ~/.pgpass (omit password=)
 
+    def __repr__(self) -> str:
+        # Mask the password so tools that render frame locals via repr() (pretty
+        # tracebacks, pytest --showlocals, debuggers) cannot leak the secret. The
+        # None vs non-None distinction is preserved because password=None marks
+        # ~/.pgpass authentication, which is useful to see when debugging. The
+        # username is intentionally not masked (the CLI already prints it).
+        pw = "***" if self.password is not None else None
+        return f"Credentials(username={self.username!r}, password={pw!r})"
+
 
 def _interactive() -> bool:
     return sys.stdin.isatty()
--- a/src/jkp/data/cli.py
+++ b/src/jkp/data/cli.py
@@ -19,6 +19,11 @@ app = typer.Typer(
     name="jkp",
     help="JKP Factor Data generation pipeline.",
     no_args_is_help=True,
+    # Defense in depth: never render frame locals in tracebacks. A WRDS
+    # Credentials object bound in a command frame would otherwise be exposed to
+    # Typer's pretty exception handler. Typer only defaulted this off in 0.23.0
+    # and the project floor is typer>=0.15.0, so set it explicitly.
+    pretty_exceptions_show_locals=False,
 )
 
 
--- a/tests/unit/test_cli.py
+++ b/tests/unit/test_cli.py
@@ -185,3 +185,15 @@ class TestConnectCommand:
         result = runner.invoke(app, ["connect", "-r"])
         assert result.exit_code == 0
         mock_reset.assert_called_once_with(full_reset=True)
+
+
+@pytest.mark.unit
+class TestCliAppConfig:
+    """Guard the app-level configuration that hardens credential handling."""
+
+    def test_pretty_exceptions_show_locals_disabled(self) -> None:
+        """Frame locals must never be rendered in tracebacks: a WRDS Credentials
+        object bound in a command frame would otherwise be exposed to Typer's
+        pretty exception handler. This pins the setting so a future refactor of
+        the Typer(...) constructor cannot silently drop it."""
+        assert app.pretty_exceptions_show_locals is False
--- a/tests/unit/test_wrds_credentials.py
+++ b/tests/unit/test_wrds_credentials.py
@@ -1147,3 +1147,56 @@ def _write_pgpass_append(mod, content):
     path = mod._pgpass_path()
     with path.open("a") as fh:
         fh.write(content)
+
+
+@pytest.mark.unit
+def test_repr_masks_password() -> None:
+    """repr() shows the username but replaces a real password with a mask, so a
+    tool that renders frame locals (pretty tracebacks, --showlocals) cannot leak
+    the secret."""
+    from jkp.data.wrds_credentials import Credentials
+
+    password = "super-secret-value"  # noqa: S105
+    creds = Credentials("some-user", password)
+    text = repr(creds)
+    assert "some-user" in text
+    assert "***" in text
+    assert password not in text
+
+
+@pytest.mark.unit
+def test_repr_preserves_none_password() -> None:
+    """password=None (the ~/.pgpass authentication path) stays distinguishable in
+    the repr, since that visibility is useful when debugging."""
+    from jkp.data.wrds_credentials import Credentials
+
+    creds = Credentials("pgpass-user", None)
+    text = repr(creds)
+    assert "pgpass-user" in text
+    assert "password=None" in text
+    assert "***" not in text
+
+
+@pytest.mark.unit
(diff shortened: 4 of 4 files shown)


capellini:
@claude review this PR thoroughly and make suggestions if appropriate.

capellini:
Thanks @claude. Addressed the one optional suggestion in 4d49b43: added `TestCliAppConfig.test_pretty_exceptions_show_locals_disabled`, a regression test asserting `app.pretty_exceptions_show_locals is False` so a future refactor of the `Typer(...)` constructor cannot silently drop the guard. No functional code changed.

src/jkp/data/wrds_credentials.py
```diff
@@ -71,6 +71,15 @@ class Credentials:
     username: str
     password: str | None  # None => authenticate via ~/.pgpass (omit password=)
 
+    def __repr__(self) -> str:
```
FernandoRDLL, line 74: This fixes the CLI path, but I don't think it's the whole repr story, and the title sort of reads like it is.

The password is still sitting in plain frame locals elsewhere: `password` in `_prompt_and_store`, `env_pw`/`keyring_pw` in `get_wrds_credentials`, and the big one, the `conninfo` string from `gen_wrds_connection_info` (`aux_functions.py:700`). That one has `password='...'` baked right into it and then gets passed around as an argument through `get_columns`, `_attach_wrds`, `_compute_histograms` and most of the download helpers, so it's a local in a lot of frames. If a pdb session or `pytest --showlocals` or some crash reporter fires inside any of those, the password prints. None of that goes through Typer, so the `pretty_exceptions_show_locals` flag doesn't cover it.

All of this predates your PR so I'm not asking you to fix it here. Could we just note the scope in the description, or open a follow-up for `conninfo`? Feels like it wants the same treatment `_redact_password` already gives the error strings.

FernandoRDLL (commented):
Nice catch on this one. I poked at both mechanisms to make sure they actually do what the description says, and they hold up.

The dataclass part I wasn't sure about going in, but you're right that it leaves a hand-written `__repr__` alone. The one I really wanted to check was rich, since it has its own dataclass pretty-printer that could easily have gone straight around the mask. It doesn't — once you override `__repr__` it stops treating the class as a dataclass and just calls `repr()`, so the mask holds in exactly the traceback path you're targeting.

Also confirmed `pretty_exceptions_show_locals` has been on `Typer.__init__` since well before 0.15.0, so setting it explicitly is safe at our floor. And asserting the attribute instead of driving the CLI makes sense given `CliRunner` skips `Typer.__call__`.

One thing inline about scope.

src/jkp/data/wrds_credentials.py
capellini: The title is pretty well-scoped to the Credentials repr, so I don't think it's claiming anything beyond that, but your point that this isn't the end of the story is valid and well taken. I'll open a separate PR to track down and suppress the password elsewhere. Opened #281 to track it.

PR #261, opened 2026-07-21
fix: pass --force to jkp build in slurm batch script

### LINKED ISSUE #260: [Bug]: Slurm batch script dies on interactive overwrite prompt from jkp build
## Which script were you running?
Other (slurm/submit_job_som_hpc.slurm)

## What happened?
`slurm/submit_job_som_hpc.slurm` runs `jkp build data` without the `--force` flag. When the output directory `data/` already contains data (the normal case when refreshing an existing dataset), `jkp build` calls `typer.confirm("Output directory 'data' already contains data. Overwrite?", abort=True)` (src/jkp/data/cli.py:77-81).

In a non-interactive Slurm batch job there is no TTY, so the confirmation reads EOF and aborts, and `set -e` then kills the job. The prompt can never be answered, so any resubmission against an existing `data/` directory fails immediately.

## Expected behavior
The batch script is non-interactive by design, and resubmitting it implies intent to regenerate the data, so it should pass `--force` to `jkp build`.

## Proposed fix
Add `--force` to both `jkp build data` invocations in `slurm/submit_job_som_hpc.slurm`.

theisij:
## Summary

`slurm/submit_job_som_hpc.slurm` is a non-interactive batch script, but `jkp build data` prompts `Output directory 'data' already contains data. Overwrite? [y/N]` when `data/` is non-empty (src/jkp/data/cli.py:77-81). With no TTY the confirm reads EOF and aborts, and `set -e` kills the job — so any resubmission against an existing data directory fails immediately. This PR adds `--force` to both `jkp build` invocations, since resubmitting the batch job implies intent to regenerate the data.

Fixes #260

## Change Type

## Methodological Impact

None

## Data Impact

None — the flag only skips the interactive confirmation; the pipeline output is unchanged.

## Breaking Changes

None

## Tests

None added — two-line change to a Slurm shell script; there is no test harness for the batch scripts.

## Checklist

- [x] I have run the tests locally (`pytest`) — not applicable, no Python code changed
- [x] I have run the linter locally (`ruff check src/jkp/data/ tests/`) — not applicable, no Python code changed

## Additional Notes

Encountered in practice: a resubmitted HPC job died on the prompt because `data/` already held output from a prior run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ If you do not have a WRDS subscription, you can still access pre-computed factor
      sbatch slurm/submit_job_som_hpc.slurm
      ```
      to create the factor returns, stock returns, and firm characteristics.
+     Note that the batch script passes `--force` to `jkp build`, so it overwrites an
+     existing output directory without prompting (an interactive `jkp build` asks first).
 
      In an interactive session, run:
      ```sh
--- a/slurm/submit_job_som_hpc.slurm
+++ b/slurm/submit_job_som_hpc.slurm
@@ -45,6 +45,8 @@ source .venv/bin/activate
 
 # Create char data
 # Set PERSISTENT_WRDS_CONNECTION=1 to use a single WRDS connection (reduces MFA on NAT-rotated networks)
+# Use --force because a batch job has no TTY to answer the interactive overwrite prompt, and
+# resubmitting the job implies intent to regenerate the data.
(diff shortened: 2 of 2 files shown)


capellini (commented):
LGTM. Overwriting `data/` is the right default for a resubmitted batch job.

FYI, I plan to add a short comment in the script explaining why `--force` is there, so no action needed on your end.

capellini (approved):

PR #263, opened 2026-07-22
feat: parameterize output directory in slurm batch script

### LINKED ISSUE #262: [Feature]: Parameterize the output directory in the Slurm batch script
### Feature Description

`slurm/submit_job_som_hpc.slurm` hardcodes the output directory as `data`. Add an optional positional argument so the directory can be chosen at submit time:

```sh
sbatch slurm/submit_job_som_hpc.slurm /gpfs/scratch/user/jkp-run
```

With no argument the script behaves exactly as it does today.

### Use Case / Motivation

Writing to a directory other than `data` (for example scratch storage, or keeping several dated runs side by side) currently requires editing the script. A submit-time argument avoids that and mirrors the `jkp build <output_dir>` CLI, which already takes the directory as an argument.

N/A (infrastructure change, no methodological impact).

- `slurm/submit_job_som_hpc.slurm`: set `OUTPUT_DIR="${1:-data}"` after the `cd "${SLURM_SUBMIT_DIR}"` block and pass `"${OUTPUT_DIR}"` to both `jkp build` and `jkp portfolio`, so the two stages always write to the same directory. Relative paths keep resolving against the submit directory, as they do now.
- `README.md`: add a line to the sbatch instructions showing the override.

### Willingness to Contribute

### Alternatives Considered

An environment variable (like the existing `PERSISTENT_WRDS_CONNECTION`). A positional argument makes the output directory more explicit on the submit command line, whereas an environment variable may be inadvertently set in one's environment: sbatch exports the submitting shell's environment into the job by default, so a stray `OUTPUT_DIR` would silently redirect the run.

### Additional Context

Follow-up to PR #261 (merged 22-07-2026), which added `--force` to the `jkp build` invocations in this script. This issue parameterizes the output directory those invocations write to.

### Pre-submission Checklist

capellini:
## Summary

`slurm/submit_job_som_hpc.slurm` hardcoded the output directory as `data`. This PR takes it from an optional `--output-dir` flag instead, used by both `jkp build` invocations and `jkp portfolio` so the two stages always write to the same place:

```sh
sbatch slurm/submit_job_som_hpc.slurm --output-dir /path/to/output
```

With no flag the script behaves exactly as before. Relative paths resolve against the submit directory, and the header block echoes the resolved destination so `slurm/output_jkp_%j.txt` records where the run actually wrote.

Options are parsed with the enhanced (util-linux) `getopt`, so an unrecognized flag, a missing value, or a stray positional argument is rejected with a usage message instead of being silently ignored. `getopt` is not a dependency of the `jkp` package, since these batch scripts are worked examples of running the pipeline under Slurm rather than part of it. The requirement is recorded in the script header and in the README, and the script also checks for it at startup: if `getopt` is missing, or only the legacy version is present, the job exits 1 with an explanatory message rather than misparsing its arguments.

Fixes #262

## Change Type

## Methodological Impact

None

## Data Impact

None. The argument changes where output is written, not what is written.

## Breaking Changes

None. The default output directory is unchanged.

## Tests

None added, there is no test harness for the batch scripts. `bash -n` passes and `shellcheck` reports the same single `SC1091` for `source .venv/bin/activate` as it does on `main`, so no new warnings.

Argument handling was verified on the SOM HPC across nine submissions of the real script with the pipeline invocations stubbed out. The five valid forms all COMPLETED with the resolved destination in the job log: no flag (`<submit dir>/data`), `--output-dir` with a space, `--output-dir=` with an equals sign, a relative path, and `--help`. The four invalid forms all FAILED with exit 1 and a usage message: an unquoted path containing a space (`unexpected argument: Runs/jkp`), a bare positional, an unknown flag (`unrecognized option '--frce'`), and `--output-dir` with no value.

Separately confirmed that `sbatch` passes `--output-dir` through to the script untouched, including the `=` form, and that even `--output`, which is a real `sbatch` option, reaches the script when it appears after the script path.

The `getopt` guard was exercised directly, with the binary absent from `PATH` and with a stub standing in for the legacy version. Both exit 1 with the explanatory message rather than proceeding.

An earlier full submission (job 1549091) confirmed end to end that the pipeline creates the `raw`/`interim`/`processed` tree under the requested directory and begins downloading WRDS tables into it, with no stray `data/` created. That job used the positional syntax this PR has since replaced, so it stands as evidence about the directory handling rather than about the current flag.

## Checklist

- [x] I have run the tests locally (`pytest`) - not applicable, no Python code changed
- [x] I have run the linter locally (`ruff check src/jkp/data/ tests/`) - not applicable, no Python code changed
- [ ] I have verified the pipeline runs successfully after my changes - smoke tested through the start of `DOWNLOAD_RAW_DATA_TABLES` only

## Additional Notes

Follow-up to #261, which added `--force` to the same invocations.

### DIFF
--- a/slurm/submit_job_som_hpc.slurm
+++ b/slurm/submit_job_som_hpc.slurm
@@ -22,6 +22,13 @@ echo "SLURM_NODES are ${SLURM_NODELIST}"
 echo '-------------------------------'
 echo
 
+# Output directory for the pipeline, taken from the first script argument.
+# Relative paths resolve against the submit directory. Defaults to "data":
+#
+#     sbatch slurm/submit_job_som_hpc.slurm /path/to/output
+#
+OUTPUT_DIR="${1:-data}"
+
 # (Optional) Make threaded libs respect Slurm CPU allocation
 export OMP_NUM_THREADS=${SLURM_CPUS_PER_TASK}
 export MKL_NUM_THREADS=${SLURM_CPUS_PER_TASK}
@@ -48,10 +55,10 @@ source .venv/bin/activate
 # Use --force because a batch job has no TTY to answer the interactive overwrite prompt, and
 # resubmitting the job implies intent to regenerate the data.
 if [ "${PERSISTENT_WRDS_CONNECTION:-0}" = "1" ]; then
-    jkp build data --force --persistent-connection
+    jkp build "${OUTPUT_DIR}" --force --persistent-connection
 else
-    jkp build data --force
+    jkp build "${OUTPUT_DIR}" --force
 fi
 
 # Create portfolio data
-jkp portfolio data
+jkp portfolio "${OUTPUT_DIR}"
--- a/slurm/submit_job_som_hpc.slurm
+++ b/slurm/submit_job_som_hpc.slurm
@@ -13,16 +13,59 @@
 # proceeding to `jkp portfolio` against incomplete data.
 set -e
 
-# Output directory for the pipeline, taken from the first script argument.
-# Relative paths resolve against the submit directory. Defaults to "data":
-#
-#     sbatch slurm/submit_job_som_hpc.slurm /path/to/output
-#
-# Assigned after the cd below so relative paths resolve against the submit
-# directory, and echoed so the job log records where the run actually wrote.
+# Option parsing requires the enhanced (util-linux) getopt(1). It is present in
+# the base install of every mainstream Linux distribution, including this
+# cluster's login and compute nodes, and is deliberately NOT a dependency of the
+# `jkp` package: these batch scripts are worked examples of running the pipeline
+# under Slurm, not part of the installed package. The --test guard below fails
+# loudly rather than misparsing if only the legacy getopt is available.
+
+usage() {
+    cat <<'USAGE'
+usage: sbatch slurm/submit_job_som_hpc.slurm [--output-dir DIR]
+
+  --output-dir DIR  Directory for pipeline output (raw, interim and processed
+                    data). Relative paths resolve against the submit directory.
+                    Defaults to "data".
+  -h, --help        Show this message and exit.
+USAGE
+}
+
+# `getopt --test` exits 4 on the enhanced version, so it cannot be run bare
+# under `set -e`.
+GETOPT_TEST_RC=0
+getopt --test >/dev/null || GETOPT_TEST_RC=$?
+if [ "${GETOPT_TEST_RC}" -ne 4 ]; then
+    echo "error: this script requires the enhanced (util-linux) getopt" >&2
+    exit 1
+fi
+
+PARSED=$(getopt -o 'h' --long 'output-dir:,help' -n 'submit_job_som_hpc.slurm' -- "$@") || {
+    usage >&2
+    exit 1
+}
+eval set -- "${PARSED}"
+
+OUTPUT_DIR=data
+while true; do
+    case "$1" in
+        --output-dir) OUTPUT_DIR="$2"; shift 2 ;;
+        -h|--help) usage; exit 0 ;;
+        --) shift; break ;;
+        *) echo "error: unhandled option: $1" >&2; exit 1 ;;
+    esac
+done
+
+if [ "$#" -ne 0 ]; then
+    echo "error: unexpected argument: $1" >&2
+    usage >&2
+    exit 1
+fi
+
+# Resolved after the cd so a relative --output-dir reports the same path the
+# pipeline will actually write to.
 echo '-------------------------------'
 cd "${SLURM_SUBMIT_DIR}"
-OUTPUT_DIR="${1:-data}"
 echo "${SLURM_SUBMIT_DIR}"
 echo "Running on host $(hostname)"
 echo "Time is $(date)"
--- a/slurm/submit_job_som_hpc.slurm
+++ b/slurm/submit_job_som_hpc.slurm
@@ -13,21 +13,23 @@
 # proceeding to `jkp portfolio` against incomplete data.
 set -e
 
-echo '-------------------------------'
-cd "${SLURM_SUBMIT_DIR}"
-echo "${SLURM_SUBMIT_DIR}"
-echo "Running on host $(hostname)"
-echo "Time is $(date)"
-echo "SLURM_NODES are ${SLURM_NODELIST}"
-echo '-------------------------------'
-echo
-
 # Output directory for the pipeline, taken from the first script argument.
 # Relative paths resolve against the submit directory. Defaults to "data":
 #
 #     sbatch slurm/submit_job_som_hpc.slurm /path/to/output
 #
+# Assigned after the cd below so relative paths resolve against the submit
+# directory, and echoed so the job log records where the run actually wrote.
+echo '-------------------------------'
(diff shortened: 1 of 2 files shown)


capellini:
@claude review this PR thoroughly and make suggestions if appropriate

capellini:
Took the first suggestion in 8a55640: the README now shows the persistent-connection example combined with a custom output directory.

On the second point, there is no gap to close. `setup_folder_structure` runs as the first pipeline step and creates the output directory and any missing parents (`mkdir(parents=True, exist_ok=True)`), and an unwritable path fails right there, seconds into the job, so no download work is wasted.

slurm/submit_job_som_hpc.slurm
```diff
@@ -22,6 +22,13 @@ echo "SLURM_NODES are ${SLURM_NODELIST}"
 echo '-------------------------------'
 echo
 
+# Output directory for the pipeline, taken from the first script argument.
+# Relative paths resolve against the submit directory. Defaults to "data":
+#
+#     sbatch slurm/submit_job_som_hpc.slurm /path/to/output
+#
+OUTPUT_DIR="${1:-data}"
```
FernandoRDLL: Would you consider echoing the resolved directory in the header block above? Something like `echo "OUTPUT_DIR is ${OUTPUT_DIR}"`, either here or by moving the assignment up above line 16.

The reasoning: the destination is user-supplied now, and `--force` unconditionally suppresses the non-empty-directory guard at `cli.py:77`, while `setup_folder_structure` creates the tree with `mkdir(parents=True)`. So `sbatch slurm/submit_job_som_hpc.slurm /gpfs/scratch/user` when you meant `/gpfs/scratch/user/jkp-run` proceeds silently and writes hundreds of GB into an unrelated existing directory. Since the header prints host, time, and nodelist but not the destination, `slurm/output_jkp_%j.txt` has no record of where the job actually wrote.

This was a non-issue while the target was the hardcoded `data`; it only becomes one now that it's an argument. One line makes each job log self-describing, which also helps with the "several dated runs side by side" use case from #262.

FernandoRDLL: Minor: extra positional arguments are silently dropped. `sbatch slurm/submit_job_som_hpc.slurm /scratch/run --output-format csv` runs happily against `/scratch/run` and ignores the flag with no hint that it did.

A `[ "$#" -le 1 ] || { echo "usage: sbatch $0 [output_dir]" >&2; exit 1; }` guard would catch it. I'd leave this to your judgement — it would be the script's first argument-validation code, and on a single-user HPC script the mistake is cheap to notice. Flagging it mainly because an unquoted path with a space lands here rather than in an obvious error.

(For contrast, a flag in *first* position already fails loudly: `jkp build --force --force` exits 2 on the missing positional and `set -e` aborts before any WRDS work.)

FernandoRDLL (commented):
Looks good — the implementation matches what #262 proposed, including the positional-argument-over-environment-variable choice and its reasoning.

I went looking for the failure modes this pattern usually has and didn't find any. Worth calling out that the `:-` in `${1:-data}` is doing real work: it substitutes on unset *or* null, so `sbatch slurm/submit_job_som_hpc.slurm ""` still lands on `data`. With the plain `${1-data}` form the empty string would have gone through to `jkp build`, and `Path("")` is `PosixPath('.')` — which exists and is non-empty, so `--force` would have suppressed the prompt and written `raw/`/`interim/`/`processed/` into the repo checkout. Good call.

Also checked: `bash -n` passes; `shellcheck` reports the same single `SC1091` for `source .venv/bin/activate` as it does on `main`, so no new warnings; absolute and deep paths work since `output_dir` is an unvalidated `typer.Argument(Path)` (`cli.py:46,92`) and `setup_folder_structure` uses `mkdir(parents=True, exist_ok=True)`; and the static reference data loads from package resources (`paths.py:39-70`) rather than the output tree, so pointing a run at an arbitrary path is safe. The comment at line 26 is accurate — the `cd` at line 17 comes first, and it holds even under `sbatch -D`.

Two optional bits of polish inline, neither a blocker.

slurm/submit_job_som_hpc.slurm
capellini: Good call. Moving the `OUTPUT_DIR` assignment into the header block and echoing it there, so the destination lands in `slurm/output_jkp_%j.txt` next to host, time and nodelist.

Echoing the resolved path rather than the raw argument, since a relative path resolves against the submit directory and the absolute form is what you want when comparing dated runs:

```sh
echo "Output directory is $(realpath -m "${OUTPUT_DIR}")"
```

`-m` because the directory does not exist yet at that point, `setup_folder_structure` creates it later in the run. That also puts the assignment after the `cd` rather than above it, or a relative path would resolve against the wrong directory.

capellini: Yes, you're right that it's better to have proper validation of parameters. I implemented something more extensible and less brittle than a strict argument count check. The script now takes a `--output-dir` keyword argument instead of a positional argument and it is parsed with the enhanced (util-linux) getopt.

My hesitation about the `$#` line was that it is an arity assertion rather than argument validation. It does not look at values, so `sbatch slurm/submit_job_som_hpc.slurm --force` would still have set `OUTPUT_DIR` to `--force` and only failed because typer rejects it downstream, as you found. And we would need a more robust solution as soon as we implemented a second argument. But converting the output directory argument from a positional to a keyword argument makes rejecting invalid arguments a stable rule that survives adding flags.

It also catches the case the positional form could not catch. Your unquoted path with a space now leaves `Runs/jkp` as a leftover positional and fails.

Verified on the SOM HPC across nine submissions of the real script with the pipeline calls stubbed. Valid forms COMPLETED with the resolved destination in the log: no flag, `--output-dir` with a space, `--output-dir=` with an equals sign, a relative path, and `--help`. Invalid forms FAILED with exit 1 and usage: an unquoted split path (`unexpected argument: Runs/jkp`), a bare positional, an unknown flag (`unrecognized option '--frce'`), and `--output-dir` with no value.

Four things worth mentioning:
1. `sbatch` passes the flag through untouched, including the `=` form. Even `--output`, which is a real `sbatch` option, reaches the script when it appears after the script path, so a keyword is as collision-safe here as a positional was.
2. `getopt --test` exits 4 on the enhanced version, so it cannot be called bare under the script's `set -e` or it aborts the job. It is run guarded. With `getopt` absent from `PATH`, or with only the legacy version, the script exits 1 with an explanatory message rather than misparsing.
3. `getopt -n 'submit_job_som_hpc.slurm'` puts the real script name in the error, instead of using `$0`, which under `sbatch` evaluates to the spool copy of the script (on our cluster, `/cm/local/apps/slurm/var/spool/job<id>/slurm_script`).
4. I don't think that getopt should be a package dependency. This repository's aim is to provide a Python package with which users can generate their own JKP factor data. The slurm scripts are provided for reference and are not authoritative dependencies in the package. They are provided for convenience as an example of how we run our pipeline (and, of course, to help us generate our data on our HPC cluster). The getopt requirement is in the script header and the README, backed by the startup check.

The README and PR description are updated.

FernandoRDLL (approved):
Both inline suggestions addressed. Approving; the empty `--output-dir ""` fallback noted above can be an optional follow-up.

PR #268, opened 2026-07-29
Compute overnight and intraday portfolio returns

miahdy:
# Daily overnight/intraday factor portfolios (USD and local)

## Summary

This PR builds on #253 and extends the portfolio step (`jkp portfolio`) to construct **daily** factor portfolios using overnight and intraday return components in both **USD** and **local currency**, writing them as parallel outputs alongside the existing excess-return factors under `data/processed/portfolios/`.

It reuses the standard JKP portfolio machinery (ECDF sorting, breakpoints, EW/VW/VW-cap weighting, HML/LMS aggregation, country/regional/cluster rollups).

**Prerequisite:** Stock-level O/I columns must already exist in `data/processed/return_data/daily_rets_by_country/` from the build step:
`ret_overnight`, `ret_intraday`, `ret_overnight_local`, `ret_intraday_local`.

### Return columns used

| Component | Daily portfolio return column | Currency |
|-----------|-------------------------------|----------|
| Overnight | `ret_overnight` | USD |
| Intraday | `ret_intraday` | USD |
| Overnight (local) | `ret_overnight_local` | Local |
| Intraday (local) | `ret_intraday_local` | Local |

- **Sorting / universe:** always the standard monthly `ret_exc_lead1m` from characteristics (unchanged).
- **Daily return leg:** the component column above (contemporaneous daily returns).
- **Monthly O/I factor portfolios:** not produced (monthly compounded O/I stock returns exist in characteristics, but are not used here).

### Implementation

- **`_build_oi_factor_returns`** (`portfolio.py`): Runs the full per-country portfolio loop for one return component at daily frequency only. Calls `portfolios()` with the standard sort (`ret_exc_lead1m`) and an overridden `daily_ret_col`, then stacks country results and builds daily HML, LMS, cluster, and regional outputs.
- **`run_portfolio`** (`portfolio.py`): After writing standard excess-return outputs, when `daily_pf` is enabled, loops over:

  ```python
  ("overnight", "ret_overnight"),
  ("intraday", "ret_intraday"),
  ("overnight_local", "ret_overnight_local"),
  ("intraday_local", "ret_intraday_local"),
  ```

  Checks that the required daily column exists in the daily return files; skips gracefully with a log message if not.
- **`portfolios()`** (`aux_functions.py`): Accepts `daily_ret_col` (default `ret_exc`). Dynamically selects the requested daily return column. Winsorization of daily returns is applied only for the standard `ret_exc` path; O/I legs are aggregated as-is.

### Output files

[...]

PR #267, opened 2026-07-25
docs: correct Ohlson O-score intercept

nightcityblade:
## Summary

Correct the Ohlson O-score intercept in the detailed documentation from `-1.37` to `-1.32`, matching both the Python implementation and the reference SAS implementation.

Fixes #265.

## Change Type

- [ ] Methodology change (affects factor definitions or calculations)
- [ ] Data/output change (affects computed outputs users download)
- [ ] Infrastructure change (CI, config, dependencies)
- [x] Documentation only
- [ ] Performance-impacting change

## Methodological Impact

None. This aligns the documentation with the coefficient already used by the implementations.

## Data Impact

None. No calculation or output code changes.

## Breaking Changes

None.

## Tests

- `uv run pytest tests/unit/test_accounting_formulas.py` — 18 passed
- `uv run ruff check src/jkp/data/ tests/` — passed
- `uv run ruff format --check src/jkp/data/ tests/` — 95 files already formatted

No test was added because this is a documentation-only correction; the existing O-score tests pass.

## Checklist

- [x] I have run the tests locally (`pytest`)
- [x] I have run the linter locally (`ruff check src/jkp/data/ tests/`)
- [ ] I have verified the pipeline runs successfully after my changes
- [x] I have updated documentation if needed
- [x] I have considered whether this change affects factor calculations

## Additional Notes

The data pipeline was not run because this change only corrects a documented constant and does not modify executable code.

PR #269, opened 2026-08-05
Constant-window rolling-daily characteristics junk values fix

### LINKED ISSUE #256: Constant-window rolling-daily characteristics leak junk values into the ECDF ranking
Follow-up to #247. While reviewing the daily-coverage changes there, I went looking at what happens when the newly-admitted firm-months are actually dead.

### What's going on

A bunch of our rolling-daily characteristics are computed over a stock's window of returns/prices/volume. If that window is *constant* — a halted or non-trading firm-month where nothing moves — the math underneath is meaningless, but we still hand back a number. Depending on the characteristic that number comes out as `NaN` (correlation, skew), `inf` (the turnover and dolvol coefficients of variation, amihud), `0.0` (the volatilities and rmax), or `1.0` (price-to-high).

The problem is that portfolio construction only throws away *nulls*. Everything else — including all of that junk — flows straight into the ECDF rank. And because of how Polars sorts, `inf` lands above every finite value and `NaN` above `inf`, so those rows end up pinned at the top breakpoint and get assigned to the extreme portfolio. The `0.0` and `1.0` cases aren't any better; they just sort to the bottom or top naturally. Either way a stock that should have been dropped ends up in a factor's long or short leg, and we'd never see it.

### How #247 leaks junk values

This isn't strictly new. Our quality screen has always been `zero_obs < 10` (`base_data_filter_exp`, `aux_functions.py:9006`, and the corr_data build in `prepare_daily` around line 7641), and a stock can sit under that threshold while still being flat, so some degenerate windows have always slipped through globally.

What #247 changed is that `zero_obs_gate_ok` now lets *every* CRSP row through regardless of `zero_obs` (`source_crsp == 1`). That's exactly the US dead-month case, so we're now feeding more of these flat windows into the stats than before. The loosening is US-only — non-US goes through the unchanged `zero_obs < 10` path — but the underlying weakness is global.

### Which characteristics are affected

Everything that touches variance, skew, or a ratio on the stock's own series:

- **NaN:** `mktcorr` (`pl.corr`), `rskew`, the FF3/HXZ4/CAPM idiosyncratic skews and `coskew`, and the turnover/dolvol coefficients of variation.
- **inf:** amihud (`|ret|/dolvol`), and again the turnover/dolvol ratios when the mean is zero.
- **0.0:** `rvol`, the three `ivol_*`, and `rmax` — a fake "zero risk" reading.
- **1.0:** `prc_to_high` — a constant price reads as "sitting at its 52-week high," which is nonsense for a stock that simply isn't trading.

A few are genuinely fine and I don't think we need to touch them: the betas divide by *market* variance (`cov/var(mktrf)`), which is basically never zero; `mktrf_vol` is market-side only; and `res_mom` already guards itself with `.fill_nan(None)` — which is more or less the pattern I think we want everywhere else.

### Reproducing the mechanism (no WRDS needed)

```python
import polars as pl
n = 800
const = pl.Series([0.0]*n); mkt = pl.Series([0.001*((i%7)-3) for i in range(n)])
df = pl.DataFrame({"r": const, "m": mkt})
df.select(pl.col("r").skew(bias=False))                                  # NaN  (rskew, iskew_*)
df.select(pl.corr("r", "m"))                                             # NaN  (mktcorr)
df.select(pl.col("r").std())                                             # 0.0  (rvol, ivol_*)
pl.DataFrame({"t": const}).select(pl.col("t").std() / pl.col("t").mean())# NaN  (turnover/dolvol CoV)
```

And the reason it survives to the output: `portfolios()` only filters `is_not_null()` (`aux_functions.py:10046`), `z_ranks()` only filters `is_not_nan()` and lets `inf` through (`8397`), and `add_ecdf()` (`9606`) sorts by the value so the non-finite rows pin to the top breakpoint.

### What I'd propose

The cleanest framing is: if the window is constant, we shouldn't be producing a characteristic for it at all — so null it and let the existing null-exclusion do its job. Concretely, guard each aggregation on its own input series:

```python
pl.when(<window varies>).then(stat_expr).otherwise(None)
```
Maybe a clean way to implement this would be to put a gate on the roll apply daily pipeline before the aggregation functions.

Some ideas on how to implement:

- `n_unique() == 1` is exact but builds a hash set per group — the slowest, and still only catches bit-for-bit identical values.
- `std() == 0` is cheap but fragile: a truly-constant window can come back as `std ≈ 1e-16` because of mean rounding, so it'd miss the very case we care about. It's also not meaningful for the ratio characteristics.
- `min() == max()` is cheap and exact, but has the same blind spot as `n_unique` — if there's any upstream float drift, a "flat" window won't be bit-identical and we'd miss it.
- `max() - min() < eps` catches the near-constant case too, but now we're picking an `eps`, and the right value depends on the characteristic's scale (returns are ~1e-2, dolvol ~1e6).

We want to make sure we don't slow this down since a lot of work has gone to making this faster.

One case the constant-window guard *won't* catch: amihud can still go `inf` from a single zero-`dolvol` day inside an otherwise-normal window. So I'd also add a plain non-finite scrub over the float characteristic columns in `finish_daily_chars()` (`aux_functions.py:8365`) as a backstop — `pl.when(pl.col(c).is_finite()).then(pl.col(c)).otherwise(None)`. We already do something similar for the >1000% return clip at `aux_functions.py:2457`.

Longer term, tightening the gate in root-cause-1 (dropping flat windows regardless of `source_crsp`/`zero_obs`) would handle all of this in one place, but the per-characteristic null is the smaller, safer change to make first.

### Done when

- No `NaN` or `inf` left in `data/processed/characteristics/*.parquet`. (At least in the chars we're touching)
- A regression test where a synthetic constant-window firm-month comes back null (not an extreme-portfolio value) for `corr`, `rskew`, `iskew_*`, `coskew`, `rvol`, `ivol_*`, `rmax`, `prc_to_high`, `turnover`, `dolvol`, and `ami` — plus the single zero-`dolvol` day nulling `ami` via the backstop.
- A `CHANGELOG_DATA.md` entry, since this shifts portfolio membership for the affected id-months.

miahdy:
**Closes:** #256

## Summary

- Added a `_guard_constant(input_col, stat_expr)` helper that nulls a statistic when `min == max` for the input series within a rolling window group, preventing degenerate values (NaN, inf, 0.0, 1.0) from reaching portfolio construction
- Applied the guard to all affected rolling daily functions: `rvol`, `rmax`, `skew`, `capm`, `capm_ext`, `ff3`, `hxz4`, `prc_to_high`, and `mktcorr`
- Hardened `ami` against zero-dolvol days by guarding with `pl.when(dolvol_d == 0).then(None)` and `.fill_nan(None)`
- Added a non-finite backstop in `finish_daily_chars` that scrubs any residual NaN/inf across all float columns before writing the final output

## Motivation

Issue #256 identified that stocks with constant (non-trading / halted) return windows produce mathematically meaningless values — NaN from correlation/skewness, inf from Amihud ratios, fake 0.0 from volatilities, and fake 1.0 from price-to-high. These values survived through portfolio construction because only nulls were filtered, causing dead stocks to be assigned to extreme portfolio legs.

The loosened daily coverage gate from #247 (allowing all CRSP rows through regardless of `zero_obs`) amplified the problem for US stocks, but the underlying weakness was global.

## Approach

**Per-characteristic guard (`_guard_constant`):**
- Uses `min() != max()` — two cheap aggregations that piggyback on the existing `group_by`, exact for bit-identical constant values (the dead-stock case)
- Applied at the point of aggregation so the null propagates naturally through downstream joins
- Beta columns are intentionally left unguarded since they divide by *market* variance which is never zero

**Non-finite backstop (`finish_daily_chars`):**
- Applies `pl.when(col.is_finite()).then(col).otherwise(None)` across all float columns
- Catches edge cases the constant-window guard cannot (e.g., a single zero-dolvol day producing inf within an otherwise normal window)

## Verification (on produced data)

Checked all 96 country parquet files + `world_data.parquet`:
- **Zero NaN or inf** in any affected rolling daily characteristic across all countries
- **Zero exact-zero** values in any volatility column (`rvol_*`, `ivol_*`), confirming constant windows are nulled rather than producing fake zero-risk readings
- All `prc_highprc_252d == 1.0` rows have positive `rvol_252d > 0`, confirming they represent genuine at-high stocks (not degenerate constant-price windows)
- Portfolio files clean (no NaN/inf in factor returns from these characteristics)

## Test plan

- [x] `TestGuardConstantHelper` — unit tests for the `_guard_constant` expression itself (constant → null, varying → value, all-null → null, single element → null)
- [x] `TestConstantWindow*` — per-characteristic regression tests confirming constant-window inputs produce null for: `rvol`, `rmax`, `skew`, `capm` (ivol nulled, beta preserved), `capm_ext` (ivol/iskew/coskew nulled, beta preserved), `ff3`, `hxz4`, `prc_to_high`, `mktcorr`, `ami`
- [x] `TestFinishDailyCharsBackstop` — verifies inf and NaN are scrubbed to null while finite values pass through unchanged
- [x] Full data rebuild passes with no NaN/inf in `data/processed/characteristics/*.parquet`

### DIFF
--- a/CHANGELOG_DATA.md
+++ b/CHANGELOG_DATA.md
@@ -3,6 +3,11 @@ This change log keeps track of changes to the underlying data set. In brackets, 
 
 This repository ports the original SAS pipeline ([ReplicationCrisis](https://github.com/bkelly-lab/ReplicationCrisis)) to Python using Polars. Entries up to and including 05-03-2025 are from the original change log.
 
+## 05-08-2026
+__Changes__:
+- Nulled rolling daily characteristics computed over constant (non-varying) return/price windows; affected columns include `rvol_*`, `ivol_capm_*`, `ivol_ff3_*`, `ivol_hxz4_*`, `rmax*`, `rskew_*`, `iskew_*`, `coskew_*`, `corr_*`, and `prc_highprc_*` ([#256](https://github.com/bkelly-lab/jkp-data/pull/256))
+- Added non-finite (NaN/inf) backstop scrub in `finish_daily_chars` to prevent degenerate values from reaching portfolio construction
+
 ## 23-07-2026
 __Changes__:
 - Added local-currency overnight/intraday return columns (`ret_intraday_local`, `ret_overnight_local`) alongside the existing USD versions, following the Lou, Polk, and Skouras (2019) decomposition:
--- a/src/jkp/data/aux_functions.py
+++ b/src/jkp/data/aux_functions.py
@@ -8567,6 +8567,12 @@ def finish_daily_chars(paths: DataPaths, output_path):
         betabab_1260d=col("corr_1260d") * col("rvol_252d") / col("__mktvol_252d"),
         rmax5_rvol_21d=col("rmax5_21d") / col("rvol_252d"),
     ).drop("__mktvol_252d")
+
+    float_cols = [name for name, dtype in daily_chars.collect_schema().items() if dtype.is_float()]
+    daily_chars = daily_chars.with_columns(
+        pl.when(pl.col(c).is_finite()).then(pl.col(c)).otherwise(None).alias(c) for c in float_cols
+    )
+
     daily_chars.collect().write_parquet(output_path)
 
 
@@ -9386,6 +9392,17 @@ def res_mom(df, sfx, __min, incl, skip):
     return df
 
 
+def _guard_constant(input_col: str, stat_expr: pl.Expr) -> pl.Expr:
+    """Null the stat when the input series is constant within a group_by group.
+
+    Uses min() != max() — two cheap aggregations that piggyback on the
+    existing group_by, exact for bit-identical values (the dead-stock case).
+    """
+    return (
+        pl.when(pl.col(input_col).min() != pl.col(input_col).max()).then(stat_expr).otherwise(None)
+    )
+
+
 def rvol(df, sfx, __min):
     """
     Description:
@@ -9399,7 +9416,7 @@ def rvol(df, sfx, __min):
         LazyFrame with f'rvol{sfx}'.
     """
     df = df.group_by(["id_int", "group_number"]).agg(
-        col("ret_exc").cast(pl.Float64).std().alias(f"rvol{sfx}")
+        _guard_constant("ret_exc", col("ret_exc").cast(pl.Float64).std()).alias(f"rvol{sfx}")
     )
     return df
 
@@ -9418,8 +9435,8 @@ def rmax(df, sfx, __min):
     """
     df = df.group_by(["id_int", "group_number"]).agg(
         [
-            col("ret").top_k(5).mean().alias(f"rmax5{sfx}"),
-            col("ret").max().alias(f"rmax1{sfx}"),
+            _guard_constant("ret", col("ret").top_k(5).mean()).alias(f"rmax5{sfx}"),
+            _guard_constant("ret", col("ret").max()).alias(f"rmax1{sfx}"),
         ]
     )
     return df
@@ -9438,7 +9455,7 @@ def skew(df, sfx, __min):
         LazyFrame with f'rskew{sfx}'.
     """
     df = df.group_by(["id_int", "group_number"]).agg(
-        col("ret_exc").skew(bias=False).alias(f"rskew{sfx}")
+        col("ret_exc").skew(bias=False).fill_nan(None).alias(f"rskew{sfx}")
     )
     return df
 
@@ -9465,9 +9482,10 @@ def prc_to_high(df, sfx, __min):
         df.group_by(["id_int", "group_number"])
         .agg(
             [
-                (col("prc_adj").sort_by("date").last() / col("prc_adj").max()).alias(
-                    f"prc_highprc{sfx}"
-                ),
+                _guard_constant(
+                    "prc_adj",
+                    col("prc_adj").sort_by("date").last() / col("prc_adj").max(),
+                ).alias(f"prc_highprc{sfx}"),
                 pl.count("prc_adj").alias("n"),
             ]
         )
@@ -9499,9 +9517,13 @@ def capm(df, sfx, __min):
         .agg(
             [
                 (pl.cov("ret_exc", "mktrf") / pl.var("mktrf")).alias(f"beta{sfx}"),
-                (col("ret_exc") - col("mktrf") * (pl.cov("ret_exc", "mktrf") / pl.var("mktrf")))
-                .std()
-                .alias(f"ivol_capm{sfx}"),
+                _guard_constant(
+                    "ret_exc",
+                    (
+                        col("ret_exc")
+                        - col("mktrf") * (pl.cov("ret_exc", "mktrf") / pl.var("mktrf"))
+                    ).std(),
+                ).alias(f"ivol_capm{sfx}"),
             ]
         )
     )
@@ -9526,7 +9548,7 @@ def ami(df, sfx, __min):
         df.group_by(["id_int", "group_number"])
         .agg(
             [
-                (col("ret").abs() / aux_1 * 1e6).mean().alias(f"ami{sfx}"),
+                (col("ret").abs() / aux_1 * 1e6).mean().fill_nan(None).alias(f"ami{sfx}"),
                 pl.count("dolvol_d").alias("n"),
             ]
         )
@@ -9603,9 +9625,13 @@ def capm_ext(df, sfx, __min):
     df = df.group_by(["id_int", "group_number"]).agg(
         [
             beta_col.cast(pl.Float64).alias(f"beta{sfx}"),
-            residual_col.std().alias(f"ivol_capm{sfx}"),
-            residual_col.skew(bias=False).alias(f"iskew_capm{sfx}"),
-            (exp_coskew1 / exp_coskew2).alias(f"coskew{sfx}"),
+            _guard_constant("ret_exc", residual_col.std()).alias(f"ivol_capm{sfx}"),
+            _guard_constant("ret_exc", residual_col.skew(bias=False))
+            .fill_nan(None)
+            .alias(f"iskew_capm{sfx}"),
+            _guard_constant("ret_exc", exp_coskew1 / exp_coskew2)
+            .fill_nan(None)
+            .alias(f"coskew{sfx}"),
         ]
     )
     return df
@@ -9630,8 +9656,10 @@ def ff3(df, sfx, __min):
         df.filter(col("smb_ff").is_not_null() & col("hml").is_not_null())
         .group_by(["id_int", "group_number"])
         .agg(
-            res_exp.std(ddof=3).alias(f"ivol_ff3{sfx}"),
-            res_exp.skew(bias=False).alias(f"iskew_ff3{sfx}"),
+            _guard_constant("ret_exc", res_exp.std(ddof=3)).alias(f"ivol_ff3{sfx}"),
+            _guard_constant("ret_exc", res_exp.skew(bias=False))
+            .fill_nan(None)
+            .alias(f"iskew_ff3{sfx}"),
         )
     )
     return df
@@ -9658,8 +9686,10 @@ def hxz4(df, sfx, __min):
         )
         .group_by(["id_int", "group_number"])
         .agg(
-            res_exp.std(ddof=4).alias(f"ivol_hxz4{sfx}"),
-            res_exp.skew(bias=False).alias(f"iskew_hxz4{sfx}"),
+            _guard_constant("ret_exc", res_exp.std(ddof=4)).alias(f"ivol_hxz4{sfx}"),
+            _guard_constant("ret_exc", res_exp.skew(bias=False))
+            .fill_nan(None)
+            .alias(f"iskew_hxz4{sfx}"),
         )
     )
     return df
@@ -9726,6 +9756,7 @@ def dolvol(df, sfx, __min):
             pl.when(col("dolvol_d").mean() != 0)
             .then(col("dolvol_d").std() / col("dolvol_d").mean())
             .otherwise(fl_none())
+            .fill_nan(None)
             .alias(f"dolvol_var{sfx}"),
         ]
     )
@@ -9757,6 +9788,7 @@ def turnover(df, sfx, __min):
             pl.when(col(f"turnover{sfx}") != 0)
             .then(col("turnover_std") / col(f"turnover{sfx}"))
             .otherwise(fl_none())
+            .fill_nan(None)
             .alias(f"turnover_var{sfx}"),
         )
         .filter(col("n") >= __min)
@@ -9781,7 +9813,7 @@ def mktcorr(df, sfx, __min):
         df.group_by(["id_int", "group_number"])
         .agg(
             pl.len().alias("n"),
-            pl.corr("ret_exc_3l", "mkt_exc_3l").alias(f"corr{sfx}"),
+            pl.corr("ret_exc_3l", "mkt_exc_3l").fill_nan(None).alias(f"corr{sfx}"),
         )
         .filter(col("n") >= __min)
         .drop("n")
--- a/tests/unit/test_constant_window_guard.py
+++ b/tests/unit/test_constant_window_guard.py
@@ -1,0 +1,335 @@
+"""Unit tests for constant-window guards and the non-finite backstop in finish_daily_chars.
+
+Verifies that rolling daily characteristics produce null (not NaN, inf, fake 0.0,
+or fake 1.0) when the input window is constant, and that finish_daily_chars scrubs
+any residual non-finite values.
+"""
+
+from __future__ import annotations
+
+from datetime import date
+
+import numpy as np
+import polars as pl
+import pytest
+
+from jkp.data.aux_functions import (
+    _guard_constant,
+    ami,
+    capm,
+    capm_ext,
+    ff3,
+    finish_daily_chars,
+    hxz4,
+    mktcorr,
+    prc_to_high,
+    rmax,
+    rvol,
+    skew,
+)
+
+# ---------------------------------------------------------------------------
+# Helper: _guard_constant expression
+# ---------------------------------------------------------------------------
+
+
+class TestGuardConstantHelper:
+    """Tests for the _guard_constant expression helper."""
+
+    def test_constant_series_returns_none(self):
+        df = pl.DataFrame({"x": [5.0, 5.0, 5.0], "grp": [1, 1, 1]})
+        result = df.group_by("grp").agg(_guard_constant("x", pl.col("x").std()).alias("val"))
+        assert result["val"][0] is None
+
+    def test_varying_series_returns_value(self, tolerance):
+        df = pl.DataFrame({"x": [1.0, 2.0, 3.0], "grp": [1, 1, 1]})
+        result = df.group_by("grp").agg(_guard_constant("x", pl.col("x").std()).alias("val"))
+        np.testing.assert_allclose(result["val"][0], 1.0, **tolerance.STANDARD)
+
+    def test_all_null_returns_none(self):
+        df = pl.DataFrame({"x": pl.Series([None, None, None], dtype=pl.Float64), "grp": [1, 1, 1]})
+        result = df.group_by("grp").agg(_guard_constant("x", pl.col("x").std()).alias("val"))
+        assert result["val"][0] is None
+
+    def test_single_element_returns_none(self):
+        df = pl.DataFrame({"x": [7.0], "grp": [1]})
+        result = df.group_by("grp").agg(_guard_constant("x", pl.col("x").std()).alias("val"))
+        assert result["val"][0] is None
+
+
+# ---------------------------------------------------------------------------
+# Per-stat constant-window regression tests
+# ---------------------------------------------------------------------------
+
+N = 20
+
+
+def _const_ret_df(value: float = 0.0, n: int = N) -> pl.DataFrame:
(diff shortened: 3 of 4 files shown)


src/jkp/data/aux_functions.py
```diff
@@ -9386,6 +9392,17 @@ def res_mom(df, sfx, __min, incl, skip):
     return df
 
 
+def _guard_constant(input_col: str, stat_expr: pl.Expr) -> pl.Expr:
```
FernandoRDLL, line 9395: One structural note on the helper before the site-specific comments.

`min() != max()` is a bit-exact test, so it fires only when the window is *perfectly* constant. The stocks that worry me most aren't quite those — they're the ones that trade once or twice in 21 days. Those have min ≠ max, sail through the guard, and still produce a volatility computed off twenty zeros and one real observation.

Worth thinking about whether this wants a threshold rather than an equality — see the `zero_obs` note in the review body.

src/jkp/data/aux_functions.py
```diff
@@ -9399,7 +9416,7 @@ def rvol(df, sfx, __min):
         LazyFrame with f'rvol{sfx}'.
     """
     df = df.group_by(["id_int", "group_number"]).agg(
-        col("ret_exc").cast(pl.Float64).std().alias(f"rvol{sfx}")
+        _guard_constant("ret_exc", col("ret_exc").cast(pl.Float64).std()).alias(f"rvol{sfx}")
```
FernandoRDLL, line 9419: I don't think this fires for the multi-month windows.

`ret_exc = ret − coalesce(t30ret, rf)/21` (`:2839`), and the risk-free is joined from the monthly tables on a month key — so it's a step function: one value for all of January, a different one for February.

Windows here are month-aligned blocks of k months (`_21d`→1, `_126d`→6, `_252d`→12, `_1260d`→60). At k=1 the window sits inside a single calendar month, rf is constant, and the guard works fine. At k>1 rf changes every month, so a genuinely dead stock has a *varying* `ret_exc` and `min() != max()` comes back True.

Running 12 months of `ret == 0.0` through it:

```
rvol_252d      = 7.99e-06   (expected null)
ivol_capm_252d = 7.99e-06   (expected null)
```

The awkward part is that this also quietly disarms the verification in the description. The fake `0.0` becomes a fake `8e-06`, so "zero exact-zero values in `rvol_*`/`ivol_*`" passes cleanly while the dead stock still wins the low-volatility sort. And `rvol_252d` is the denominator of published `rmax5_rvol_21d`, so a stock dead 11 of 12 months produces a large *finite* ratio that the `is_finite` backstop won't catch either.

`rmax` just below at `:9438` guards on `"ret"` and works correctly for exactly this reason — raw return really is constant for a dead stock. Same one-word change would fix it here: `_guard_constant("ret", ...)`.

Same applies to `capm`'s ivol, `capm_ext`, `ff3` and `hxz4` — flagged individually below.

src/jkp/data/aux_functions.py
```diff
@@ -9499,9 +9517,13 @@ def capm(df, sfx, __min):
         .agg(
             [
                 (pl.cov("ret_exc", "mktrf") / pl.var("mktrf")).alias(f"beta{sfx}"),
```
FernandoRDLL, line 9519: Left unguarded deliberately per the description — beta divides by market variance, which is never zero. That's true, but it establishes that the value is *defined*, not that it means anything.

For a constant-return window `cov(ret_exc, mktrf)` is ~0, so beta comes out at or near `0.0`, which reads as "perfectly immune to market moves". A halted stock ends up ranked as the most defensive name in the cross-section — the same failure mode #256 describes for `rvol`, just in a different column.

Two other betas outside this diff have the same shape, and theirs are exactly zero rather than near-zero:

- `betadown_252d` (`:9578`) → exactly `0.0`; published, `config.py:66`
- `beta_dimson_21d` (`:9823`) → exactly `0.0`, and the function's own `is_not_null() & is_not_nan()` filter passes it straight through; published, `config.py:64`

`betabab_1260d` gets saved incidentally because it multiplies by the now-null `rvol_252d`, but the plain betas don't get that.

(The `ivol_capm` guard just below has the `ret_exc` problem from `:9419`.)

src/jkp/data/aux_functions.py
```diff
@@ -9603,9 +9625,13 @@ def capm_ext(df, sfx, __min):
     df = df.group_by(["id_int", "group_number"]).agg(
         [
             beta_col.cast(pl.Float64).alias(f"beta{sfx}"),
-            residual_col.std().alias(f"ivol_capm{sfx}"),
-            residual_col.skew(bias=False).alias(f"iskew_capm{sfx}"),
-            (exp_coskew1 / exp_coskew2).alias(f"coskew{sfx}"),
+            _guard_constant("ret_exc", residual_col.std()).alias(f"ivol_capm{sfx}"),
```
FernandoRDLL, line 9628: Same `ret_exc` keying issue as `rvol` — see the note at `:9419`. Over a `_252d` window the rf drift makes `min() != max()` true for a fully dead stock, so this won't fire. Guarding on `"ret"` would.

src/jkp/data/aux_functions.py
```diff
@@ -9630,8 +9656,10 @@ def ff3(df, sfx, __min):
         df.filter(col("smb_ff").is_not_null() & col("hml").is_not_null())
         .group_by(["id_int", "group_number"])
         .agg(
-            res_exp.std(ddof=3).alias(f"ivol_ff3{sfx}"),
-            res_exp.skew(bias=False).alias(f"iskew_ff3{sfx}"),
+            _guard_constant("ret_exc", res_exp.std(ddof=3)).alias(f"ivol_ff3{sfx}"),
```
FernandoRDLL, line 9659: Same as `:9419` — keyed on `ret_exc`, so it won't fire on the 6- and 12-month windows where rf steps between months.

src/jkp/data/aux_functions.py
```diff
@@ -9658,8 +9686,10 @@ def hxz4(df, sfx, __min):
         )
         .group_by(["id_int", "group_number"])
         .agg(
-            res_exp.std(ddof=4).alias(f"ivol_hxz4{sfx}"),
-            res_exp.skew(bias=False).alias(f"iskew_hxz4{sfx}"),
+            _guard_constant("ret_exc", res_exp.std(ddof=4)).alias(f"ivol_hxz4{sfx}"),
```
FernandoRDLL, line 9689: Same as `:9419` — keyed on `ret_exc`, so it won't fire on the multi-month windows.

src/jkp/data/aux_functions.py
```diff
@@ -9438,7 +9455,7 @@ def skew(df, sfx, __min):
         LazyFrame with f'rskew{sfx}'.
     """
     df = df.group_by(["id_int", "group_number"]).agg(
-        col("ret_exc").skew(bias=False).alias(f"rskew{sfx}")
+        col("ret_exc").skew(bias=False).fill_nan(None).alias(f"rskew{sfx}")
```
FernandoRDLL, line 9458: Unlike the other stats, this relies on Polars happening to emit `NaN` for the degenerate case and then converting it, rather than going through `_guard_constant`.

Two things about that: `fill_nan` doesn't catch ±inf, and if a future Polars version returns `0.0` here instead of `NaN` the protection disappears silently — no test would fail, because the tests assert the outcome rather than the mechanism.

Small thing, but the description lists `skew` and `mktcorr` among the functions that got the guard; in the diff they only got `.fill_nan(None)`.

src/jkp/data/aux_functions.py
```diff
@@ -9526,7 +9548,7 @@ def ami(df, sfx, __min):
         df.group_by(["id_int", "group_number"])
         .agg(
             [
-                (col("ret").abs() / aux_1 * 1e6).mean().alias(f"ami{sfx}"),
+                (col("ret").abs() / aux_1 * 1e6).mean().fill_nan(None).alias(f"ami{sfx}"),
```
FernandoRDLL, line 9551: I think this line is a no-op. `aux_1` just above already turns a zero `dolvol_d` into null, and that predates this PR — so `|ret| / aux_1` is null, never NaN or inf, and `.fill_nan(None)` can't fire. The description credits the `pl.when(dolvol_d == 0)` guard to this PR, but it's already on `main`.

The case that does bite isn't covered: a dead stock that still has volume. `ret == 0` with `dolvol_d == 1e6` over 126 days gives `ami_126d = 0.0`, so it ranks as the single most liquid name in the universe.

src/jkp/data/aux_functions.py
```diff
@@ -9726,6 +9756,7 @@ def dolvol(df, sfx, __min):
             pl.when(col("dolvol_d").mean() != 0)
             .then(col("dolvol_d").std() / col("dolvol_d").mean())
             .otherwise(fl_none())
+            .fill_nan(None)
```
FernandoRDLL, line 9759: `.fill_nan(None)` only helps if `dolvol_d` already contains NaN. The degenerate case for a non-trading stock is *constant* volume, which gives `std == 0` and `dolvol_var_126d = 0.0` — a perfectly finite number meaning "perfectly stable volume", so it lands in an extreme leg. Published in `config.py:88`.

Same story for `turnover_var` at `:9791` (`config.py:201`): constant `tvol`/`shares` → `0.0`, not null.

FernandoRDLL (commented):
Could you merge `main` in first? The PR is showing as conflicting — `main` picked up #267 and #268 since you branched, and the only actual collision is `CHANGELOG_DATA.md`, where both sides added a dated entry at the top. Keeping both should do it, with your `## 05-08-2026` above the `## 30-07-2026` from #268. Everything else merges clean.

---

Went through this carefully and ran the functions against multi-month windows. The direction is right — nulling degenerate stats instead of letting fake `0.0`s through is exactly what #256 needs — but I think the guard is pointed at the wrong column, and there may be a simpler path available.

**The thing I keep coming back to: we already have `zero_obs`.**

`prepare_daily` builds it at `aux_functions.py:7498` as a per-`(id_int, eom)` count of days where `ret_local == 0`, and `zero_obs_gate_ok()` at `:7454` gates every return-based stat on it:

```python
(pl.col("source_crsp") == 1) | (pl.col("zero_obs") < 10)
```

That `source_crsp == 1` short-circuit is the #247 loosening you mention in the description — every CRSP row passes regardless of how many dead days it carries. So the mechanism for this problem already existed and got switched off for the whole US universe, and #256 is the fallout.

Two things make `zero_obs` a better basis than min/max:

- it's computed off raw `ret_local`, so it's immune to the risk-free-rate issue I've flagged on `rvol` below
- it's a *count*, so it catches near-dead stocks. A stock that trades once in 21 days isn't constant — min ≠ max, guard passes — but its volatility is computed off twenty zeros and one real observation and is just as meaningless. `min != max` is bit-exact, so it only fires on the perfectly-dead case, which I suspect is the smaller half of the problem.

`zero_obs` is already carried in `dsf1.parquet` and `corr_data.parquet` and survives `prepare_base_data` untouched, so it's in scope inside every function this PR touches. Was it considered and ruled out? There may well be a reason I'm not seeing.

One wrinkle if you do go that way: `zero_obs` is a monthly count broadcast onto every row of the month, so you can't sum it inside a k-month group-by without overcounting by ~21×. Computing `(col("ret") == 0).mean()` at window granularity is the same idea and reads cleaner — `zero_trades` at `:9385` already does that shape with `tvol`.

On the mechanism itself: I benchmarked `min != max` and it costs 3–5% over no guard at both 21d and 252d shapes, with `n_unique()` strictly worse and hoisting it to a single column not worth the plumbing. So no efficiency concern at all — it's cheap, and the null/NaN semantics check out. My worry is entirely about what it's applied to.

Rest of the notes inline.

src/jkp/data/aux_functions.py
```diff
@@ -9781,7 +9813,7 @@ def mktcorr(df, sfx, __min):
         df.group_by(["id_int", "group_number"])
         .agg(
             pl.len().alias("n"),
-            pl.corr("ret_exc_3l", "mkt_exc_3l").alias(f"corr{sfx}"),
+            pl.corr("ret_exc_3l", "mkt_exc_3l").fill_nan(None).alias(f"corr{sfx}"),
```
FernandoRDLL, line 9816: `fill_nan` won't fire here, for the same reason as `rvol`: `ret_exc_3l` inherits the month-varying rf, so `pl.corr` returns a finite number rather than NaN.

A dead stock over 12 months gives me `corr_1260d = 4.6e-17` rather than null — and `betabab_1260d = corr × rvol_252d / mktvol` stays finite junk as a result.

This is also why the CHANGELOG claim that `corr_*` is nulled over constant windows doesn't hold at `_1260d`.

CHANGELOG_DATA.md
```diff
@@ -3,6 +3,11 @@ This change log keeps track of changes to the underlying data set. In brackets,
 
 This repository ports the original SAS pipeline ([ReplicationCrisis](https://github.com/bkelly-lab/ReplicationCrisis)) to Python using Polars. Entries up to and including 05-03-2025 are from the original change log.
 
+## 05-08-2026
+__Changes__:
+- Nulled rolling daily characteristics computed over constant (non-varying) return/price windows; affected columns include `rvol_*`, `ivol_capm_*`, `ivol_ff3_*`, `ivol_hxz4_*`, `rmax*`, `rskew_*`, `iskew_*`, `coskew_*`, `corr_*`, and `prc_highprc_*` ([#256](https://github.com/bkelly-lab/jkp-data/pull/256))
```
FernandoRDLL, line 8: Link points at `.../pull/256`, but 256 is the issue this closes — the PR is #269, so the link lands somewhere unrelated.

Separately, given the `ret_exc` keying: as things stand `rvol_*`, `ivol_capm_*` and `corr_*` are only nulled at `_21d`, so this entry is a little ahead of the actual behaviour.

tests/unit/test_constant_window_guard.py
```diff
@@ -0,0 +1,335 @@
...
+            {
+                "id_int": [1] * n,
+                "group_number": [10] * n,
+                "ret": [0.0] * n,
+                "dolvol_d": [1e6] * n,
+            }
+        )
+        result = ami(df, "_126d", __min=60)
+        val = result["ami_126d"][0]
+        assert val is None or val == 0.0
```
FernandoRDLL, line 248: `assert val is None or val == 0.0` can't fail — it passes whether the guard works or not. And `0.0` is precisely the degenerate value we're trying to get rid of here (dead stock ranked most liquid), so this ends up baking it in as expected behaviour.

Think this wants to be `assert val is None`, which would currently fail and surface the `ami` gap.

tests/unit/test_constant_window_guard.py
```diff
@@ -0,0 +1,335 @@
+"""Unit tests for constant-window guards and the non-finite backstop in finish_daily_chars.
+
+Verifies that rolling daily characteristics produce null (not NaN, inf, fake 0.0,
```
FernandoRDLL, line 3: One coverage gap worth closing: every constant-window test in here uses a single-month window, which is exactly the case where the guard does work. A twelve-month case would have surfaced the `ret_exc` problem on `rvol`/`capm`/`ff3`/`hxz4` and the `corr_*` one straight away.

The suite is green at 970 passing, which is why none of this showed up.

src/jkp/data/aux_functions.py
miahdy: Redesigned `_guard_constant` so it still requires `min != max`, and for return-like columns also requires `(col == 0).mean() < 0.9` (window-level zero fraction, same shape as `zero_trades` on `tvol`). That catches “trades once in 21 days” without summing the monthly `zero_obs` broadcast (which would overcount in multi-month groups). For non-return columns (`prc_adj`, `dolvol_d`, …) we keep `max_zero_frac=None` so only the constantness check applies.

src/jkp/data/aux_functions.py
miahdy: Applied `_guard_constant("ret", …)` to beta in `capm`, `capm_ext`, `downbeta`, and `dimsonbeta`. Updating the tests that previously asserted beta is preserved under constant returns so they now expect null.

src/jkp/data/aux_functions.py
miahdy: Switched the guard key from `"ret_exc"` to `"ret"` so constant raw returns null the window regardless of RF steps. Same keying change for the other return-based stats that were keyed on `ret_exc`. Adding a `_252d` test where `ret` is constant and `ret_exc` varies month-to-month so this can’t regress.

src/jkp/data/aux_functions.py
miahdy: Replaced it with `_guard_constant("ret", skew_expr)` so constant / near-dead windows are nulled by the same mechanism as the other stats, independent of Polars’ NaN convention.

src/jkp/data/aux_functions.py
miahdy: Wrapped the Amihud mean in `_guard_constant("ret", …)` so all-zero (or ≥90% zero) return windows null `ami_*`. Dropping the no-op `.fill_nan`. Test assertion tightened to `assert val is None`.

src/jkp/data/aux_functions.py
miahdy: Applied `_guard_constant` on `dolvol_d` (min≠max only) for `dolvol_var`, and the same constantness check on the daily turnover series for `turnover_var`, so perfectly stable volume/turnover windows null rather than emit `0.0`.

CHANGELOG_DATA.md
miahdy: Fixed. changelog now links `#269` and the wording matches the post-fix behavior (guards keyed on raw `ret` / zero-fraction, covering multi-month windows).

PR #272, opened 2026-08-20
Fix CIZ delisting-return double-counting in CRSP security files

### LINKED ISSUE #271: [Bug]: CIZ delisting-return double-counting

main.py (data generation)

### Error Category

Computation error

### Bug Description

`prepare_crsp_sf` unconditionally compounds every non-null `DelRet` from `crsp.stkdelists` with `ret` (CIZ `MthRet` / `DlyRet`). That logic was correct under legacy CRSP SIZ, where `RET` excluded `DLRET`, but under CIZ many delisting returns are already embedded in the monthly/daily return field. Because the pipeline never reads `MthDelFlg` or `DlyDelFlg`, A/P (and related) delisting payoffs can be counted twice in CRSP monthly returns, and daily returns can be incorrectly adjusted when `DlyRet` is already the final return. This affects processed CRSP return outputs in `crsp_msf.parquet` / `crsp_dsf.parquet` and downstream world panels.

Identified by **Fan Xia** (Cox School of Business, Southern Methodist University) in:

> Xia, F. (2026). *Delisting-return measurement risk in accounting return tests*. Working paper. [SSRN 7243220](https://papers.ssrn.com/sol3/papers.cfm?abstract_id=7243220)

We thank Fan Xia for documenting this measurement error in public research code, including the maintained JKP Global Factor Data Python pipeline. We also thank [Patrick Weiss](https://github.com/patrick-weiss) (@patrick-weiss) for bringing it to our attention.

### Steps to Reproduce

1. Run the data pipeline through CRSP security-file preparation (e.g. `jkp build data/` or the `prepare_crsp_sf` step in `main.py`).
2. Inspect `src/jkp/data/aux_functions.py` in `prepare_crsp_sf`: after joining `crsp_{freq}sedelist`, the code compounds `delret` into `ret` for all non-null values without checking delisting flags.
3. Compare an A- or P-flagged delisting month where `MthRet` already includes `DelRet`:
   - Input: `MthRet = -0.40`, `DelRet = -0.50`, flag indicates `DelRet` is already in `MthRet`
   - Current output: `ret = (1 - 0.40) × (1 - 0.50) - 1 = -0.70` (double-counted)
   - Correct output: `ret = -0.40`
4. For daily data, inspect a row with `DlyDelFlg = Y` where `DlyRet` already reflects the delisting return; the same compounding step still runs.

### Expected Behavior

- **Monthly:** Respect `MthDelFlg`. Use `MthRet` as-is when `DelRet` is already included (A/P/V). For G, use contemporaneous timing (`MthRet` alone). Apply separate delisting adjustments only where the flag indicates an unobserved payoff (M), with any −30% imputation documented as a researcher assumption.
- **Daily:** Use `DlyRet` as-is for both `DlyDelFlg = Y` (delisting return present) and `N` (non-delisting return present). Do not compound `DelRet`.

### Error Output / Stack Trace

```shell
No runtime exception. This is a **data / measurement bug**: delisting-month returns in processed outputs are numerically incorrect for affected CRSP observations. Example of the incorrect transformation (monthly A/P case):


MthRet = -0.40, DelRet = -0.50 (already included in MthRet)
Current:  ret = (1 + MthRet) × (1 + DelRet) - 1 = -0.70
Expected: ret = MthRet = -0.40
```

### Operating System

Linux 5.14.0-427.13.1.el9_4.x86_64 (RHEL 9.4)

### Python Version

3.11.14 (`uv run python --version`)

### Polars Version

1.40.1

### uv Version

0.11.19

### Available RAM

_No response_

### WRDS Authentication Method

None

### WRDS Two-Factor Authentication

- [ ] Yes, I approved the 2FA request
- [ ] No 2FA prompt appeared
- [ ] 2FA failed or timed out
- [ ] Not applicable

### Additional Context

_No response_

### Pre-submission Checklist

miahdy:
Closes #271 

## Summary

This PR fixes a deterministic double-counting error in CRSP return preparation under the CIZ flat-file format. `prepare_crsp_sf` currently compounds every non-null `DelRet` from `crsp.stkdelists` with `ret` (= CIZ `MthRet` / `DlyRet`), even when the delisting payoff is already embedded in the monthly/daily return field. The fix reads CRSP delisting flags and applies flag-conditional return construction instead of the legacy SIZ “add-all” compounding rule.

## Change Type

- [x] Methodology change (affects factor definitions or calculations)
- [ ] Documentation only

### CRSP monthly returns (`prepare_crsp_sf`, `freq="m"`)

- Read `MthDelFlg` from `crsp.msf_v2` (via `gen_crsp_sf` / the augmented monthly file).
- Replace unconditional `DelRet` compounding with flag-conditional logic:

| Flag | Treatment |
|------|-----------|
| **A, P, V** | `DelRet` already in `MthRet` → use `MthRet` as-is |
| **G** | Contemporaneous timing → use `MthRet` alone |
| **M** | Unobserved terminal payoff → retain existing −30% imputation for “bad delist” buckets |
| **N** | No delisting adjustment → use `MthRet` as-is |

- Preserve the existing rule that sets `ret = 0` when `ret` is missing but `delret` is present, but only compound when the flag indicates it is appropriate (M / imputed missing-payoff cases).

### CRSP daily returns (`prepare_crsp_sf`, `freq="d"`)

- Read `DlyDelFlg` from `crsp.dsf_v2`.
- **Do not compound** `DelRet` from `stkdelists` into daily returns:
  - **Y** → delisting return already in `DlyRet`
  - **N** → ordinary (non-delisting) return in `DlyRet`
- Leave `ret = DlyRet` as-is.

### Out of scope

- Compustat synthetic delisting path (`gen_delist_df` / `gen_temporary_sf`) — uses security-status codes, not CRSP `MthRet`/`DelRet`
- Broader missing-payoff sensitivity bounds from Xia (0% / −100%); this PR fixes deterministic double-counting and the stated G-timing choice only
- Full historical rebuild / portfolio re-estimation (follow-up after merge)

## Methodological Impact

Under legacy SIZ, `RET` excluded separately reported `DLRET`, so compounding was correct. Under CIZ, `MthRet`/`DlyRet` already reflect delisting payoffs for many events. The old add-all rule repeated A/P (and related) delisting losses and incorrectly adjusted daily returns.

After this fix, CRSP monthly returns follow CIZ flag semantics with contemporaneous G timing. Daily returns use `DlyRet` without a second compounding step. Xia (2026) estimates the monthly error could move accounting portfolio spreads by up to ~9 bp/month and Fama–MacBeth coefficients by up to ~0.61× their Newey–West SE; processed JKP return outputs for affected delisting months will change accordingly.

## Data Impact

**Yes.** This changes values written to:

- `data/interim/crsp_msf.parquet`
- `data/interim/crsp_dsf.parquet`
- Downstream unified panels (`world_msf.parquet`, `world_dsf.parquet`) and processed return outputs for CRSP-sourced observations on delisting dates

Affected rows are concentrated on CRSP delisting months/days. Non-delisting observations are unchanged.

## Breaking Changes

**Yes.** Previously released CRSP return outputs that used the add-all compounding rule will differ for affected delisting observations. A `CHANGELOG_DATA.md` entry is required.

## Methodology reference

- Xia, F. (2026). *Delisting-return measurement risk in accounting return tests*. SSRN 7243220.
- CRSP CIZ documentation for `MthDelFlg` and `DlyDelFlg`

## Test plan

- [x] `uv run pytest tests/unit/test_prepare_crsp_sf.py` — flag-conditional monthly compounding (A/P/V: no double-count; G: `MthRet` only; M: imputation path) — 58 passed locally
- [x] New unit tests for daily path: `DlyDelFlg = Y` and `N` both leave `ret = DlyRet` (no compounding) — `test_daily_dlydelflg_Y_no_compound`, `test_daily_dlydelflg_N_no_compound`
- [x] `uv run pytest tests/unit/` — full unit suite passes — 943 passed locally
- [x] `uv run ruff check src/jkp/data/ tests/`
- [x] `uv run ruff format --check src/jkp/data/ tests/`
- [x] Spot-check delisting months in `crsp_msf.parquet`: A/P events match `MthRet`; daily delisting days match `DlyRet` — verified on full WRDS rebuild in this workspace: 0 mismatches for monthly A/P/V/G/N (5.2M rows) and 0 mismatches for all daily rows incl. Y/N (110M rows). Cross-check vs `jkp-data` (pre-fix): 23,131 monthly diffs concentrated on A/P/G/V (new `ret` = `MthRet`); 26,354 daily diffs on `N` days where old code compounded `DelRet` into `DlyRet`
- [x] `CHANGELOG_DATA.md` updated 

## Checklist

- [x] I have verified the pipeline runs successfully after my changes — full `jkp build data/` completed; interim + processed outputs present under `/gpfs/home/dh948/jkp-data-268/data`
- [x] I have updated documentation if needed — `CHANGELOG_DATA.md` entry added
- [x] I have considered whether this change affects factor calculations — yes; delisting-month/day returns change for CRSP-sourced rows vs pre-fix `jkp-data` run; portfolio re-estimation can follow as a separate step (`jkp portfolio`)

## Additional Notes

Example of the bug (monthly A/P case):

```
MthRet = -0.40, DelRet = -0.50 (already included in MthRet)
Before: ret = (1 + MthRet) × (1 + DelRet) - 1 = -0.70
After:  ret = MthRet = -0.40
```

Primary code location: `src/jkp/data/aux_functions.py`, `prepare_crsp_sf` (current unconditional compounding at the `delret` join step).


src/jkp/data/aux_functions.py
```diff
@@ -2838,14 +2844,37 @@ def prepare_crsp_sf(paths: DataPaths, freq):
     scale = 1 if (freq == "m") else 21
     ret_exc_exp = col("ret") - pl.coalesce(["t30ret", "rf"]) / scale
 
+    __crsp_sf = __crsp_sf.join(crsp_sedelist, how="left", on=merge_vars)
```
FernandoRDLL: On the daily path nothing reads the `crsp_sedelist` columns any more — `delret`, `delistingdt`, `delreasontype`, `delactiontype`, `delpaymenttype`, `delstatustype` are all dropped at the end of the function. That leaves a left join against a ~110M-row frame doing no work.

It's also a latent nondeterminism risk: if `stkdelists` ever carries more than one row per `(permno, delistingdt)`, the fan-out gets resolved arbitrarily by the later `.unique(["permno", "date"])` — same class as what we fixed in #230/#231.

Same for `del_flag` in `gen_crsp_sf` (lines 583/663/686): read, written into the 110M-row daily parquet, then dropped unused. Could/Should we skip both when `freq == "d"`?

src/jkp/data/aux_functions.py
```diff
@@ -2713,8 +2717,10 @@ def prepare_crsp_sf(paths: DataPaths, freq):
     Steps:
         1) Read raw_data_dfs/__crsp_sf_{freq}.parquet; cast key numeric columns; apply NASDAQ volume adjustment.
         2) Compute dollar volume and infer dividend totals from (ret − retx) scaled by lagged price and split factors.
-        3) Join CRSP delists (crsp_{freq}sedelist); impute missing delret = −0.30 for “bad delist” buckets defined by CIZ codes;
-           set ret=0 when ret is missing but delret exists; compound ret with delret.
+        3) Join CRSP delists (crsp_{freq}sedelist); apply CIZ flag-conditional delisting-return
```
FernandoRDLL: `documentation/documentation.tex` still describes the old rule — both CRSP return rows (lines 2055 and 2058) say we compute `(1+RET)*(1+DLRET)-1` on delisting. That's now false for daily always, and for monthly except under the M gate. `CHANGELOG_DATA.md` was updated; this should be too.

FernandoRDLL (commented):
One more thing on validation: could you run the market return series before and after this change and report the difference? The row-level diff counts tell us how many observations moved, but a value-weighted market return before/after gives a quick read on how much the aggregate actually shifts — that's the number most people will want to see alongside the changelog entry.

Two smaller comments inline.

src/jkp/data/aux_functions.py
```diff
@@ -2838,14 +2844,37 @@ def prepare_crsp_sf(paths: DataPaths, freq):
...
+    #
+    # Monthly: MthDelFlg (propagated as del_flag) indicates whether MthRet
+    # already includes the delisting payoff.  Only flag M (unobserved terminal
+    # payoff) requires compounding; A/P/V/G/N already reflect (or exclude) the
+    # payoff in MthRet.
+    #
+    # Daily: DlyDelFlg is Y (delisting return in DlyRet) or N (ordinary return);
+    # in neither case should DelRet be compounded.
+    if freq == "m":
+        is_m_flag = col("del_flag") == "M"
```
FernandoRDLL: Could you post a frequency table of `mthdelflg` and `dlydelflg` from the WRDS pull, cross-tabbed against non-null `stkdelists.delret`? I'd like to see the distribution of the codes.

src/jkp/data/aux_functions.py
```diff
@@ -2838,14 +2844,37 @@ def prepare_crsp_sf(paths: DataPaths, freq):
...
+        __crsp_sf = (
+            __crsp_sf
+            # impute missing delret to -0.30 for "bad delist" buckets, M flag only
+            .with_columns(
+                delret=pl.when(is_m_flag & c4).then(pl.lit(-0.3)).otherwise(col("delret"))
```
FernandoRDLL: I think this gate is the wrong way round?

A "bad delist" is one where CRSP couldn't observe what shareholders actually received — that's the entire reason the −30% imputation exists. On those rows `MthRet` can't already contain the payoff, because there is no measured payoff to contain. So gating the adjustment on "payoff unobserved" means the rows that most need the correction are exactly the ones that skip it, and they come through as plain `MthRet` with the delisting loss missing.

The double-counting you're fixing lives on the other side — the flags that mean "already included" (A/P). So the rule I'd expect is: compound *unless* the flag says the payoff is already in `MthRet`, rather than compound *only when* the flag says it isn't.

Related: `test_monthly_flag_N_no_adjustment` asserts `ret == 0.05` on a row carrying `delret = -0.30`, which bakes the dropped loss into the test.

src/jkp/data/aux_functions.py
```diff
@@ -2838,14 +2844,37 @@ def prepare_crsp_sf(paths: DataPaths, freq):
...
+
+    # CIZ flag-conditional delisting-return adjustment (Xia 2026, SSRN 7243220).
+    #
+    # Monthly: MthDelFlg (propagated as del_flag) indicates whether MthRet
+    # already includes the delisting payoff.  Only flag M (unobserved terminal
+    # payoff) requires compounding; A/P/V/G/N already reflect (or exclude) the
+    # payoff in MthRet.
+    #
+    # Daily: DlyDelFlg is Y (delisting return in DlyRet) or N (ordinary return);
+    # in neither case should DelRet be compounded.
```
FernandoRDLL: The daily and monthly paths now disagree about the same delisting event. Monthly keeps the −0.30 imputation for bad delists; daily drops it entirely and never compounds `DelRet` at all. Since `crsp_dsf` feeds the rolling-window characteristics (rvol, beta, skew, 21d–1260d), a delisting will show up in the monthly file and not in the daily one. Whatever we land on for the monthly gate, the two paths should follow the same rule.

src/jkp/data/aux_functions.py
```diff
@@ -2838,14 +2844,37 @@ def prepare_crsp_sf(paths: DataPaths, freq):
...
+            __crsp_sf
+            # impute missing delret to -0.30 for "bad delist" buckets, M flag only
+            .with_columns(
+                delret=pl.when(is_m_flag & c4).then(pl.lit(-0.3)).otherwise(col("delret"))
+            )
+            # if ret missing but delret exists under M flag, set ret=0
+            .with_columns(ret=pl.when(is_m_flag & c7).then(pl.lit(0.0)).otherwise(col("ret")))
+            # compound ret with delret only for M-flagged observations
+            .with_columns(
+                ret=pl.when(is_m_flag)
```
FernandoRDLL: Worth making this more robust against unexpected flag values. `col("del_flag") == "M"` evaluates to null when the flag is null, and `pl.when(null)` falls through to `otherwise`, so a null flag silently takes the no-adjustment branch. Concretely: a delisting month with null `MthRet` and `delret = -0.5` used to come out as `ret = -0.5`; now `ret` stays null and the row drops out of every return-based characteristic.

Something like an explicit `fill_null` plus an assertion (or at least a log) of the observed flag values would make an unexpected or absent code fail loudly instead of quietly disabling the adjustment.

FernandoRDLL (commented):
Following up with the substantive comments on the flag-conditional logic. The first one gates the rest — everything else depends on which codes `MthDelFlg` actually takes in the real pull.

src/jkp/data/aux_functions.py
miahdy: I think this is a genuine asymmetry that we need to make a decision about. The monthly and daily implementations agree for A/P, where compounding is skipped, and N/null, which are treated as no-ops. They differ for M and G, however: the monthly pipeline imputes −0.30 for M and compounds ``delret`` for G, while the daily pipeline skips the ``sedelist`` join altogether.

As a result, a CRSP delisting with an M flag appears in ``crsp_msf`` but not in ``crsp_dsf``. Any rolling-window characteristic built from the daily file, such as ``rvol``, ``beta``, or ``skew``, therefore misses the delisting return. Compustat daily already assigns −0.30 on the delisting date, so this also creates a CRSP–Compustat inconsistency in ``world_dsf``.

I see two possible approaches:

A — Align daily with monthly. Join ``sedelist`` on the last trading day or ``delistingdt``, apply the same flag rules, and assign the M imputation or G compounded return on that date. This would make the daily and monthly series consistent and restore CRSP–Compustat parity. The drawback is that assigning a monthly −0.30 imputation to a single day could overstate daily variance and skewness.

B — Keep the asymmetry and document it. Daily returns would represent actual trading-day outcomes, while −0.30 would remain a researcher-imposed assumption about the total delisting loss rather than a one-day event. From that perspective, excluding it from rolling-window characteristics may be reasonable. The tradeoff is that daily-to-monthly aggregation would not match for M/G cases, and the CRSP–Compustat inconsistency would remain. If we choose this approach, I think it should be documented as an explicit methodological decision.

What do you think?

src/jkp/data/aux_functions.py
miahdy: The gate was too narrow. The direction is correct for M specifically: ``pl.when(is_m_flag).then(compound)`` does apply the correction to M rows, not skip them, and A/P rows correctly pass through. But the real gap is G rows: CRSP defines G as "MthRet does not include the delisting return" (date gap >10 trading days), so G rows can carry a valid ``delret`` that we're silently dropping. The block comment staring on line 2849 conflates "already included" (A/P) with "deliberately excluded" (G).

I flipped the logic to "compound unless A/P" rule to fix this, as it captures G rows while being a no-op for N/null (where ``delret`` is null after the left join). I can flip the gate to ~col("del_flag").is_in(["A", "P"]).

Re the test: ``test_monthly_flag_N_no_adjustment`` is actually fine (N = didn't delist), but ``test_monthly_flag_G_uses_mthret_only`` does bake in the dropped loss.

src/jkp/data/aux_functions.py
FernandoRDLL: @theisij can you please review?

theisij: Thanks for writing this up. It's a tricky decision, beacuase adding a -30% return to a single day is quite substantial. But I think the most important thing is consistency. If we assign -30% to the monthly return, that's an assumption about what happens for delisting events that we do not see, and we should carry that assumption through to daily returns. So I vote for option A.

miahdy: Sounds good. I will proceed with option A and update the daily delisting treatment to align with the monthly implementation.

miahdy:
> One more thing on validation: could you run the market return series before and after this change and report the difference? The row-level diff counts tell us how many observations moved, but a value-weighted market return before/after gives a quick read on how much the aggregate actually shifts — that's the number most people will want to see alongside the changelog entry.
> 
> Two smaller comments inline.

I used the value-weighted market returns in `processed/other_output/market_returns*.parquet`.

### USA monthly `mkt_vw` (1926–2025; 1,200 months)

* **Arithmetic mean:** 0.9542% → 0.9528% per month (−0.14 bps/month, or approximately −1.7 bps/year)
* **Geometric annualized mean:** 10.220% → 10.201% (−1.86 bps/year)
* **Absolute differences:** Mean abs. diff. = 0.20 bps; maximum \(|\Delta|\) = 7.13 bps (November 2023)
* **Months affected:** 945 of 1,200 months change, but only 52 months (4.3%) move by more than 1 bp
* The differences in `mkt_vw_exc` match those in `mkt_vw`, since the risk-free rate cancels out

### USA daily `mkt_vw`

The aggregate effect is negligible: mean abs. diff. ≈ 0.0003 bps and maximum abs. diff. ≈ 0.025 bps.

### Equal-weighted returns

USA monthly `mkt_ew` shifts more, by −1.34 bps/month on average. This is consistent with the delisting adjustments being concentrated among smaller firms.

### World ME-weighted monthly `mkt_vw`

Using countries × `me_lag1`, the geometric annualized mean declines by approximately 0.5 bps/year.

Overall, the row-level differences are real, but the value-weighted U.S. market series changes by only approximately 2 bps/year on a geometric basis. This is small enough that most users would not notice it in factor analysis, while still moving in the appropriate direction by removing double-counted delisting returns.

src/jkp/data/aux_functions.py
miahdy: Here are the counts from the current WRDS pull (`crsp_msf_v2` / `crsp_dsf_v2` × `crsp.stkdelists`; 29,833 sedelist rows, of which 26,464 have non-null `delret` and 3,369 have null `delret`).

**Monthly (`mthdelflg`), joined on `permno` + MMYY(`mthcaldt`) = MMYY(`delistingdt`):**

| `mthdelflg` | `delret` non-null | `delret` null | no sedelist match |
|---|---:|---:|---:|
| N | 0 | 0 | 5,210,773 |
| A | 18,170 | 0 | 0 |
| P | 7,538 | 0 | 0 |
| M | 0 | 3,369 | 0 |
| G | 554 | 0 | 0 |
| V | 202 | 0 | 0 |

So among non-null `delret` matches: **A 68.7%**, **P 28.5%**, **G 2.1%**, **V 0.8%**. `M` lines up exactly with the null-`delret` sedelist rows.

**Daily (`dlydelflg`):** the join key matters.

- On **`delistingdt`** (what `prepare_crsp_sf` uses), every sedelist match lands on **`N`** (26,464 non-null / 3,369 null `delret`); all 29,833 **`Y`** rows have no match.
- On **`deldlydt`** (CRSP’s daily storage date for the delisting return; typically 1–3 calendar days after `delistingdt`), every sedelist match lands on **`Y`** with the same 26,464 / 3,369 split; no **`N`** rows match that date.

So non-null `delret` is associated with `A`/`P`/`G`/`V` monthly, and with `Y` daily only when joining on `deldlydt`, not on `delistingdt`.

Issue #273, opened 2026-09-03
[Bug]: Documented file paths go stale unnoticed (broken release notes link in README, wrong static-data paths in CLAUDE.md)


FernandoRDLL:
### Problem

Documentation in this repo references files by path, and nothing verifies those paths still resolve. When a file moves, the reference silently goes stale — there is no test, no lint rule, and no CI step that would catch it. Two references are currently broken, both pointing at locations the files were moved out of:

| Reference | Points at | Actually lives at |
| --- | --- | --- |
| `README.md:72` | `documentation/release_notes.html` | `documentation/sas_to_python/release_notes.html` |
| `CLAUDE.md:69` | `data/cluster_labels.csv`, `data/country_classification.xlsx`, `data/factor_details.xlsx` | `src/jkp/data/resources/` (loaded via `_resource_path()`, `src/jkp/data/paths.py:39-70`) |

The README one is the more visible of the two: it is the pointer new users are given for "a description of the output files and a comparison between the output of the SAS/R codebase and the new Python codebase", immediately after their first successful pipeline run. On GitHub it renders as plain backticked text rather than a link, so it fails quietly — a reader tries the path, finds nothing, and has no way to know the document exists at all.

The `CLAUDE.md` one is less visible to users but arguably worse in effect: it tells Claude Code that the static reference data is at `data/`, when the pipeline actually loads it from package resources. That is exactly the kind of stale fact that gets acted on rather than checked.

### Why this is worth a systemic fix rather than two edits

These are not independent typos. Both were correct when written and broke when the files were relocated — `release_notes.html` into `documentation/sas_to_python/`, the static data into `src/jkp/data/resources/` as part of packaging. Any future move of a documented file will break its references the same way, and the repo has no mechanism that would notice. Fixing only the two current instances leaves the next one to be found by whoever happens to trip over it.

### Reproduction

```sh
# both print nothing
ls documentation/release_notes.html
ls data/cluster_labels.csv

# the files are here instead
ls documentation/sas_to_python/release_notes.html
ls src/jkp/data/resources/cluster_labels.csv
```

### Expected behaviour

1. Documented paths resolve.
2. A path that stops resolving fails CI rather than waiting to be noticed.

### Proposed fix

Two parts, separable:

- **Correct the two stale references.** Mechanical, no behaviour change.
[...]

PR #275, opened 2026-09-03
docs: correct stale file paths in README and CLAUDE.md

FernandoRDLL:
## Simple explanation

Two places in the docs point at files that have moved, so following either one leads nowhere. This corrects both paths. Nothing about the pipeline changes.

The README one is what a new user is handed right after their first successful run — the pointer to the release notes describing the output files and the SAS/R-to-Python comparison. It renders as plain text rather than a link, so it fails quietly: you try the path, find nothing, and have no way to tell the document exists.

The `CLAUDE.md` one told Claude Code that the static reference data sits in `data/`. It does not — it ships inside the package. That is the kind of stale fact that gets acted on rather than checked, so the line is rewritten to say where the files actually are and how they are loaded.

## Technical details

| File | Was | Now |
| --- | --- | --- |
| `README.md:74` | `documentation/release_notes.html` | `documentation/sas_to_python/release_notes.html` |
| `CLAUDE.md:69` | `data/cluster_labels.csv`, `data/country_classification.xlsx`, `data/factor_details.xlsx` | `src/jkp/data/resources/…`, described as packaged with the code and loaded via `_resource_path()` |

Neither was a typo. Both were correct when written and broke when the files were relocated — the release notes into `documentation/sas_to_python/`, the static data into `src/jkp/data/resources/` as part of packaging. The `CLAUDE.md` line also had its claim corrected, not just its path: "checked into the repo and used by the pipeline" implied the pipeline reads these from the data directory, when `_resource_path()` (`src/jkp/data/paths.py:39-70`) resolves them from package resources. That distinction matters — it is why pointing a run at an arbitrary output directory does not strand the reference data.

This is the first half of #273. The second half — a CI guard so the next relocation fails a check instead of waiting to be tripped over — is deliberately left out so it can be reviewed on its own merits. Without it these two lines are simply correct again, with nothing preventing a third.

## Test plan

No code changed, so there is nothing for pytest or ruff to exercise; CI runs clean by construction.

Verified the paths directly:

```sh
$ ls documentation/sas_to_python/release_notes.html
documentation/sas_to_python/release_notes.html

$ ls src/jkp/data/resources/cluster_labels.csv
src/jkp/data/resources/cluster_labels.csv
```

I also swept every backticked and Markdown-linked repo-relative path in `README.md`, `CONTRIBUTING.md`, `CLAUDE.md`, and `tests/README.md`. These two were the only ones that did not resolve, so this PR closes out the current instances in full.

Fixes the reference-correction portion of #273.

https://claude.ai/code/session_01S8YH76rEtDH4USyoyUqSdA

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -66,7 +66,7 @@ The pipeline has two entry points that run sequentially:
 
 Raw WRDS data → `data/raw/` → intermediate processing in `data/interim/` → final outputs in `data/processed/` (subdirectories: `characteristics/`, `portfolios/`, `return_data/`, `accounting_data/`, `other_output/`).
 
-Static reference data (`data/cluster_labels.csv`, `data/country_classification.xlsx`, `data/factor_details.xlsx`) is checked into the repo and used by the pipeline.
+Static reference data (`src/jkp/data/resources/cluster_labels.csv`, `src/jkp/data/resources/country_classification.xlsx`, `src/jkp/data/resources/factor_details.xlsx`) is packaged with the code and loaded via `_resource_path()` in `paths.py`, not read from the output directory.
 
 ## Code Conventions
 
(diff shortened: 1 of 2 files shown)


capellini (approved):
Approved.

Verified against `main`: all three resource files exist under `src/jkp/data/resources/`, and each loads through `_resource_path()` in `paths.py`, including `factor_details.xlsx` via `get_factor_details_path()` at `paths.py:70`. `documentation/sas_to_python/release_notes.html` is present, and neither old path exists anywhere in the tree, so both references were genuinely broken rather than merely moved.

Also confirmed this is the complete set. A grep across all tracked files turns up exactly these two lines and nothing else still pointing at the old locations.

